PR #18, opened 2019-05-02
Add code to expose Axe.Windows package version

RobGallo:
#### Describe the change

- used the informational version because it represents the semver version, whereas the file version cannot.
- Moved the code that exposes the version out of Utility and into PackageInfo to make it more obvious how to get the version

### DIFF
--- a/src/Axe.Windows.Core/Misc/PackageInfo.cs
+++ b/src/Axe.Windows.Core/Misc/PackageInfo.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+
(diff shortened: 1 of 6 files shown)


src/Axe.Windows.Core/Misc/PackageInfo.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            var attribute = ThisAssembly.Value.GetCustomAttribute<AssemblyInformationalVersionAttribute>();
+
+            return attribute?.InformationalVersion;
+        }
+
+        /// <summary>
+        /// Version string with suffix (e.g., "-prerelease") if the suffix exists
+        /// </summary>
+        public static string InformationalVersion => LazyInformationalVersion.Value;
```
DaveTryon: >InformationalVersion [](start = 29, length = 20)

Is this what the app will eventually call to obtain the AxeWindows version (for the about dialog)?

RobGallo:
Yes.

PR #19, opened 2019-05-03
Remove "Axe.Windows." from directory and file names

RobGallo:
- the text is totally superfluous
- the text is an accessibility issue
	- Screen reader users are forced to hear superfluous, uninformative text before the individuating information
- having the same prefix for each directory prevents efficient keyboard navigation by disabling first-letter navigation

#### Describe the change
(Please provide an overview of the change in this PR)


RobGallo:
@DaveTryon, 

I have made the first two changes you suggested above. I think you are correct on all counts in your third point.

PR #20, opened 2019-05-07
Rename the nuget package to Microsoft.Axe.Windows

karanbirsingh:
This PR updates the .nuspec file so that the NuGet package id includes 'Microsoft'. This allows us to be consistent with guidelines for publishing to NuGet.org via Microsoft.

### DIFF
--- a/src/CI/CI.csproj
+++ b/src/CI/CI.csproj
@@ -39,7 +39,7 @@
   </PropertyGroup>
   <ItemGroup>
     <None Include="App.config" />
-    <None Include="Axe.Windows.nuspec">
+    <None Include="Microsoft.Axe.Windows.nuspec">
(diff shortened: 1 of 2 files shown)


src/CI/CI.csproj
```diff
@@ -39,7 +39,7 @@
   </PropertyGroup>
   <ItemGroup>
     <None Include="App.config" />
-    <None Include="Axe.Windows.nuspec">
+    <None Include="Microsoft.Axe.Windows.nuspec">
```
RobGallo: Do we _need_ to rename the file?

karanbirsingh: Discussed offline, we will leave as-is - trade-off between unhelpful prefixes for navigation and name of file matching name of package

PR #24, opened 2019-05-09
Enable code coverage as a test

DaveTryon:
#### Describe the change
Enable code coverage in the PR build


DaveTryon:
Ortiga and I both started this, I'll let him finish it

PR #25, opened 2019-05-09
Enable code coverage in build result

AhmedAbdoOrtiga:
#### Describe the change
Enable code coverage to see how much we cover and act based on that.

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -53,6 +53,7 @@ jobs:
         !**\obj\**
       vsTestVersion: 15.0
       codeCoverageEnabled: true
+      publishRunAttachments: true
       platform: '$(BuildPlatform)'
       configuration: release
       rerunFailedTests: true
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -52,7 +52,7 @@ jobs:
         **\*test*.dll
         !**\obj\**
       vsTestVersion: 15.0
-      codeCoverageEnabled: false
+      codeCoverageEnabled: true
       platform: '$(BuildPlatform)'
       configuration: release
       rerunFailedTests: true


build/prbuild.yml
```diff
@@ -52,7 +52,8 @@ jobs:
         **\*test*.dll
         !**\obj\**
       vsTestVersion: 15.0
-      codeCoverageEnabled: false
+      codeCoverageEnabled: true
+      publishRunAttachments: true
```
DaveTryon: >true [](start = 29, length = 4)

According to [https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/vstest?view=azure-devops](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/vstest?view=azure-devops), the default value for publishRunAttachments is already true, so I wouldn't expect this to have any impact

DaveTryon:
>       codeCoverageEnabled: false

Shouldn't we enable the unit tests in debug, as well?

---
Refers to: build/prbuild.yml:102 in 9bafcce. [](commit_id = 9bafcce3be4b6a573eb5d47fdea8f28ac0bf9cf4, deletion_comment = False)

AhmedAbdoOrtiga:
> > ```
> >   codeCoverageEnabled: false
> > ```
> 
> Shouldn't we enable the unit tests in debug, as well?
> 
> Refers to: build/prbuild.yml:102 in [9bafcce](https://github.com/microsoft/axe-windows/commit/9bafcce3be4b6a573eb5d47fdea8f28ac0bf9cf4). [](commit_id = [9bafcce](https://github.com/microsoft/axe-windows/commit/9bafcce3be4b6a573eb5d47fdea8f28ac0bf9cf4), deletion_comment = False)

don't think we need it, is there any difference in the unit test cases between release and debug?

build/prbuild.yml
AhmedAbdoOrtiga: the weird thing, it still don't publish it!

DaveTryon:
I'm not aware of any _intentional_ differences, but my PR with both debug and release showed some slight difference (at least visually):



---
In reply to: [491139029](https://github.com/microsoft/axe-windows/pull/25#issuecomment-491139029) [](ancestors = 491139029)

PR #33, opened 2019-05-21
Remove Fakes dependency from DesktopTests project

DaveTryon:
#### Describe the change
Remove Fakes dependency from DesktopTests project. There was only 1 Shim being used, and by initializing a Stream with the test input, I could completely remove the need for the Shim. DesktopTests is now independent of Fakes

### DIFF
--- a/src/DesktopTests/Settings/SnapshotMetaInfoTests.cs
+++ b/src/DesktopTests/Settings/SnapshotMetaInfoTests.cs
@@ -1,10 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using System.IO;
 using Axe.Windows.Desktop.Settings;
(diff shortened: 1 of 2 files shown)


src/DesktopTests/Settings/SnapshotMetaInfoTests.cs
```diff
@@ -100,16 +98,12 @@ public void Deserialize_SecondPixel_Null()
 
         private static Dictionary<SnapshotMetaPropertyName, object> GetDeserializedProperties(Dictionary<SnapshotMetaPropertyName, object> inputProperties)
         {
-            using (ShimsContext.Create())
+            SnapshotMetaInfo info = new SnapshotMetaInfo(A11yFileMode.Contrast,"Test", 0, 0, inputProperties);
+            var text = JsonConvert.SerializeObject(info);
```
DaveTryon: >text [](start = 16, length = 4)

The Shim was only being used to return this from the StreamReader. By using it to initialize the Stream, we get the same result without a Shim

PR #34, opened 2019-05-22
Remove CompletenessMode and associated code

DaveTryon:
#### Describe the change
The CompletenessMode concept was added for scenarios that never materialized, and it includes code that is complicating the effort to remove Fakes. Removing it in the hope that it will make it easier to remove Fakes.


src/Actions/Actions/PrivacyExtensions.cs
```diff
@@ -1,103 +1,19 @@
...
-            return clone;
-        }
-
         /// <summary>
         /// Given the root to an element tree, synthesize a "yellow box" bitmap to describe it
         /// </summary>
         /// <param name="rootElement">The root element for the tree</param>
         /// <returns>A bitmap representing the element tee (normalized to having the app window at location (0,0)</returns>
-        internal static Bitmap SynthesizeBitmapFromElements(this A11yElement rootElement)
+        internal static Bitmap SynthesizeBitmapFromElements(this IA11yElement rootElement)
```
DaveTryon: >SynthesizeBitmapFromElements [](start = 31, length = 28)

This is used by the Load Action. One could build an argument that it ought to be in AIWin, but that's a much larger change (out of scope for now)

src/Actions/Actions/SaveAction.cs
```diff
@@ -42,40 +42,29 @@ public static class SaveAction
...
+        private static void SaveSnapshotFromElement(int? focusedElementId, A11yFileMode mode, Dictionary<SnapshotMetaPropertyName, object> otherProperties, Contexts.ElementContext ec, Package package, A11yElement root)
         {
             var json = JsonConvert.SerializeObject(root, Formatting.Indented);
             using (MemoryStream mStrm = new MemoryStream(Encoding.UTF8.GetBytes(json)))
             {
                 AddStream(package, mStrm, elementFileName);
             }
 
-            if (completenessMode == CompletenessMode.Full && ec.DataContext.Screenshot != null)
+            if (ec.DataContext.Screenshot != null)
```
DaveTryon: >ec.DataContext.Screenshot != null [](start = 16, length = 33)

This will still be true on WCOS

PR #35, opened 2019-05-22
Remove fakes from core tests

DaveTryon:
#### Describe the change
Remove Fakes dependencies from Core.Tests project. The key to making this work was to introduce the ICoreA11yElement interface and to insert it between A11yElement and IA11yElement. 


src/Core/Fingerprint/ScanResultFingerprint.cs
```diff
@@ -94,17 +94,21 @@ internal static void AddRuleSpecificContributions(IList<FingerprintContribution>
         /// <param name="element">Element under consideration</param>
         /// <param name="levelBaseName">The "base" to use when specifying contribution keys</param>
         /// <param name="ignoreNameProperty">If true, exclude the "Name" property from contributions</param>
-        private static void AddRuleAgnosticContributions(IList<FingerprintContribution> contributions, A11yElement element, int level,
+        private static void AddRuleAgnosticContributions(IList<FingerprintContribution> contributions, IA11yElement element, int level,
```
RobGallo: Couldn't the function take IA11yCoreElement as a parameter? We know it's going to be available, don't we?

DaveTryon: Unfortunately, no. The loop in lines 46-48 of this file walk the ancestry tree via IA11yElement.Parent, which returns an IA11yElement.Parent. Based strictly on the interface, there's no guarantee that the object also supports the ICoreA11yElement interface. We could potentially expose AcceratorKey and AccessKey through the IA11yElement interface, but that goes against the ideas that we discussed this morning.

---
In reply to: [286700299](https://github.com/microsoft/axe-windows/pull/35#discussion_r286700299) [](ancestors = 286700299)

PR #37, opened 2019-05-23
Remove remaining a11y element fakes

DaveTryon:
#### Describe the change
Remove the remaining instances of ShimA11yElement. At this point, A11yElement no longer requires shims


src/Actions/Misc/ExtensionMethods.cs
```diff
@@ -40,7 +40,7 @@ public static bool NeedNewDataContext(this ElementDataContext dc, DataContextMod
         /// </summary>
         /// <param name="position">Pixel location</param>
         /// <returns>element overlapping point with smallest area, or null if no elements overlap the given point</returns>
-        public static A11yElement GetSmallestElementFromPoint(Dictionary<int, A11yElement> allElements, System.Drawing.Point position)
+        internal static ICoreA11yElement GetSmallestElementFromPoint(Dictionary<int, ICoreA11yElement> allElements, System.Drawing.Point position)
```
DaveTryon: >internal [](start = 8, length = 8)

ICoreA11yElement is internal at the moment, so this function had to be internal as well.

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
```diff
@@ -17,6 +18,17 @@ namespace Axe.Windows.ActionsTests.Actions
     [TestClass]
     public class ScreenShotActionUnitTests
     {
+        // This sets the Bounding Rectangle. It breaks encapsulation in
+        // many reproachable ways
```
DaveTryon: >many reproachable ways [](start = 11, length = 22)

@RobGallo, any thoughts on cleaner options here? BoundingRect has no setter

DaveTryon:
> using Microsoft.VisualStudio.TestTools.UnitTesting;

I'll sort these blocks

---
Refers to: src/ActionsTests/Misc/ExtensionMethodsTests.cs:3 in 8bb529d. [](commit_id = 8bb529d8724885ee7e95afafceebd872e9857956, deletion_comment = False)

src/ActionsTests/Misc/ExtensionMethodsTests.cs
```diff
@@ -5,9 +5,11 @@
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.QualityTools.Testing.Fakes;
+using Moq;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Bases.Fakes;
 using Axe.Windows.Core.Misc.Fakes;
```
DaveTryon: Bases.Fakes and Misc.Fakes are no longer needed

src/ActionsTests/Misc/ExtensionMethodsTests.cs
```diff
@@ -1,13 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
```
DaveTryon: >Copyright [](start = 3, length = 9)

This file was pulled in accidentally. I can pull it out if there's a desire for it

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
RobGallo: This would be slightly cleaner if it were an extension method. But I would only implement the extension method in the unit test project, which probably means it isn't worth it.

src/ActionsTests/ActionsTests.csproj
```diff
@@ -54,6 +54,9 @@
     <Reference Include="Axe.Windows.Desktop.Fakes" Condition="$(FAKES_SUPPORTED) == 1">
       <HintPath>..\Fakes.Prebuild\FakesAssemblies\Axe.Windows.Desktop.Fakes.dll</HintPath>
```
RobGallo: Should you have removed the fakes references and the instances of the environment variable?

DaveTryon: No. ActionsTests still uses other fakes (this PR just removes the A11yElement fakes)

---
In reply to: [289189333](https://github.com/microsoft/axe-windows/pull/37#discussion_r289189333) [](ancestors = 289189333)

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
DaveTryon: That was more or less my logic. It's only used in this class so I scoped it as such

---
In reply to: [289187766](https://github.com/microsoft/axe-windows/pull/37#discussion_r289187766) [](ancestors = 289187766)

PR #38, opened 2019-05-24
Prune unused win32 code

DaveTryon:
#### Describe the change
Prune code from Win32.cs that isn't being used in Axe-Windows. Some of this code was used exclusively in AIWin, and some was apparently never used at all in AIWin


jalkire:
Is there any analogous work that can be done in AI-Win to remove the win32 code only axe-windows used?

DaveTryon:
Absolutely, although that would need to be in a PR in the AIWin repo. I started with Axe-Windows as part of staging for removing more Fakes

---
In reply to: [495797156](https://github.com/microsoft/axe-windows/pull/38#issuecomment-495797156) [](ancestors = 495797156)

DaveTryon:
Some of the code still in Win32 is used exclusively by the HotKeys code, which will (eventually) live in AIWin. the code to remove HotKeys will hopefully include a second cleanup pass through the Win32 assembly

PR #40, opened 2019-05-30
CustomAssemblyResolver cleanup, use Newtonsoft 9.0.1 throughout

DaveTryon:
#### Describe the change
  When the PowerShell cmdlets were first added, a custom assembly resolver was required to resolve the different versions of Newtonsoft.Json.dll, System.Net.Http.dll, and System.Net.Http.Formatting.dll. These assemblies, along with several Microsoft.IdentityModel* assemblies, were used by the Core project. Refactors since then have removed all but Newtonsoft.Json.dll, and if we can use the same version of Netwonsoft.Json.dll throughout, the custom assembly resolver can be removed.

  This change does the following:
1. Use Newtonsoft 9.0.1 throughout
2. Remove the CustomAssemblyResolver class (and clean up any code associated with it)
3. Remove all app.config references to System.Net.Http, System.Net.Http.Formatting, and Microsoft.IdentityModel*
4. Remove one unit test that exercised functionality specific to CustomAssemblyResolver

All unit tests pass, and the check-automation.ps1 script still succeeds. Generated A11yTest file opens successfully in AIWin


jalkire:
Is there a particular reason we decided to downgrade the higher assemblies to 9.0.1 rather than update the 9.0.1 uses to something more recent?

DaveTryon:
The 9.0.1 version is used by Sarif, and we're trying to get to a single version to avoid needing to perform custom assembly resolution in PowerShell. App-default assembly resolution will work normally in non-PowerShell scenarios.

---
In reply to: [497433866](https://github.com/microsoft/axe-windows/pull/40#issuecomment-497433866) [](ancestors = 497433866)

PR #46, opened 2019-06-05
Add new interface model for Axe.Windows.Automation

RobGallo:
#### Describe the change

The model has not been implemented yet, except for the Config class and the ScannerFactory class.

Changes since the last design review:
- Change ScanResult.OutputFiles from an IEnumerable to a Tuple: (string A11yTest, string Sarif) so that output files are grouped together, but still distinguishable by type
- Removed Scanner.Builder and created ScannerFactory since creating new objects which implement a specific interface is the job of a factory. It also hides the implementation of Scanner entirely.
- For the returned data classes, changed private access to internal access because we don't need to manage construction of these classes with a builder.

### DIFF
--- a/src/Automation/AxeWindowsAutomationException.cs
+++ b/src/Automation/AxeWindowsAutomationException.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Automation
+{
+    /// <summary>
+    /// The exception type thrown for all unhandled errors in Axe.Windows.Automation.
+    /// If an exception was thrown from code not owned by Axe.Windows.Automation,
+    /// that exception will be wrapped in <see cref="AxeWindowsAutomationException"/>
+    /// In the <see cref="Exception.InnerException"/> property.
+    /// </summary>
+    public class AxeWindowsAutomationException : Exception
+    {
+        /// <summary>
+        /// Constructor taking a text description of the exception
+        /// </summary>
+        /// <param name="message">A text description of the exception</param>
+        public AxeWindowsAutomationException(string message)
+            : base(message)
+        { }
+    } // class
+} // namespace
--- a/src/Automation/Enums/OutputFileFormat.cs
+++ b/src/Automation/Enums/OutputFileFormat.cs
@@ -5,7 +5,7 @@ using System;
 namespace Axe.Windows.Automation
 {
     /// <summary>
-    /// Bit flags to specify which output file formats an automation session should write
+    /// Bit flags to specify which output file formats a <see cref="IScanner"/> should write
     /// </summary>
     [Flags]
     public enum OutputFileFormat
--- a/src/Automation/Enums/OutputFileFormat.cs
+++ b/src/Automation/Enums/OutputFileFormat.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Automation
+{
+    /// <summary>
+    /// Bit flags to specify which output file formats an automation session should write
+    /// </summary>
+    [Flags]
+    public enum OutputFileFormat
+    {
+        /// <summary>
+        /// Create no output files
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Create output files which can be opened using <see href="https://accessibilityinsights.io/docs/en/windows/overview">Accessibility Insights for Windows</see>.
+        /// </summary>
+        A11yTest = 1,
+    } // enum
+} // namespace
--- a/src/Automation/Interfaces/IScanner.cs
+++ b/src/Automation/Interfaces/IScanner.cs
@@ -17,7 +17,7 @@ namespace Axe.Windows.Automation
         /// output files may be written to the specified directory.
         /// (Note: no output files will be written if no errors were found.)
         /// An exception may be thrown if the value of <see cref="Config.ProcessId"/> is invalid
-        /// or if the directory provided in <see cref="Config.OutputDirectory"/> does not exist.
+        /// or if the directory provided in <see cref="Config.OutputDirectory"/> cannot be created or accessed.
         /// All exceptions are wrapped in <see cref="AxeWindowsAutomationException"/>.
         /// If the exception was not thrown by AxeWindows automation, the <see cref="Exception.InnerException"/> property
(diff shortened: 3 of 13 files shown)


src/Automation/AxeWindowsAutomationException.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Automation
+{
+    /// <summary>
+    /// The exception type thrown for all unhandled errors in Axe.Windows.Automation.
+    /// If an exception was thrown from code not owned by Axe.Windows.Automation,
+    /// that exception will be wrapped in <see cref="AxeWindowsAutomationException"/>
+    /// In the <see cref="Exception.InnerException"/> property.
```
DaveTryon: >In the <see cref="Exception.InnerException"/> [](start = 8, length = 45)

This seems to have some unintended duplication from the previous line

src/Automation/AxeWindowsAutomationException.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class AxeWindowsAutomationException : Exception
```
DaveTryon: >Exception [](start = 49, length = 9)

If we're exposing this as a public class, we should probably include the full range of ctors and serialization methods that Visual Studio will want to add

src/Automation/Interfaces/IScanner.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Run AxeWindows automated tests
+        /// </summary>
+        /// <remarks>
+        /// If a value was provided in <see cref="Config.OutputDirectory"/>,
+        /// output files may be written to the specified directory.
+        /// (Note: no output files will be written if no errors were found.)
+        /// An exception may be thrown if the value of <see cref="Config.ProcessId"/> is invalid
+        /// or if the directory provided in <see cref="Config.OutputDirectory"/> does not exist.
```
DaveTryon: >if the directory provided in <see cref="Config.OutputDirectory"/> does not exist [](start = 15, length = 80)

Is this the final design? You seemed to agree when I suggested that auto-creation of an output folder is a common thing in test tools

src/Automation/Enums/OutputFileFormat.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Automation
+{
+    /// <summary>
+    /// Bit flags to specify which output file formats an automation session should write
```
DaveTryon: >automation session [](start = 58, length = 18)

We no longer expose the concept of automation session. Should this be reworded?

src/Automation/ReadMe.md
```diff
@@ -0,0 +1,65 @@
+﻿When looking at the workflow of this project, please start at
+`AutomationFactory.CreateAutomationSessionFactory`.
+
+## Compatibility practices
+
+The set of public interfaces defined in this project constitutes the
+requirements to run an AxeWindows automated scan of a given application
+and take action on the errors detected by the scan. We expect  the set
```
DaveTryon: >   [](start = 61, length = 2)

extra space

src/Automation/ReadMe.md
```diff
@@ -0,0 +1,65 @@
...
+update to new versions of the Axe.Windows package, taking advantage of rule
+improvements or updated element properties for example, without having to modify
+their code. To that end, the following sections describe types of
+changes to the interfaces.
+
+### Considerations
+
+Because this library is a strong-named .Net assembly, we can be sure 
+the code that depends on it will always link to the expected 
+version of the dll at runtime. The benefit is that we don't need
```
DaveTryon: This statement is true if the top-level app (i.e., the app consuming axe-windows) uses strong naming. If, however, the top-level application does not use strong naming, then the strong naming in axe-windows will be ignored

src/Automation/AxeWindowsAutomationException.cs
RobGallo: I'm afraid I don't see what's duplicated.

src/Automation/AxeWindowsAutomationException.cs
RobGallo: I don't disagree. But we can do that when visual studio complains about it.

src/Automation/Interfaces/IScanner.cs
RobGallo: I do agree. I just didn't remember there was a comment specifically calling this out.

src/Automation/ReadMe.md
RobGallo: The real question is if the difference should impact the current design. Thoughts? If the consumer can dynamically link to an older or newer version of AxeWindows, is that something we need to plan for? How concerned to we need to be about this possibility?

src/Automation/AxeWindowsAutomationException.cs
DaveTryon: Ah--the capitalization threw me. Since "In" was capitalized, it sounded like an incomplete thought. Given that InnerException is the standard way of wrapping an Exception, I wonder if it might be less confusing to simply end with 

"that exception will be wrapped in an <see cref="AxeWindowsAutomationException">

src/Automation/Interfaces/IScanner.cs
DaveTryon: In email, IIRC

---
In reply to: [291298458](https://github.com/microsoft/axe-windows/pull/46#discussion_r291298458) [](ancestors = 291298458)

src/Automation/ReadMe.md
DaveTryon: There *is* a scenario where a caller could get an unexpected version of axe-windows, but it's sufficiently convoluted that I don't think we need to optimize for it. I'm fine with the current compatibility strategy, but I guess I'm suggesting that we omit the reference to strong naming because our compatibility story is independent of strong naming.

---
In reply to: [291302963](https://github.com/microsoft/axe-windows/pull/46#discussion_r291302963) [](ancestors = 291302963)

src/Automation/Scanner.cs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Automation
+{
+    /// <summary>
+    /// Implementation of IScanner
```
DaveTryon: >IScanner [](start = 26, length = 8)

Maybe include a cref here?

PR #52, opened 2019-06-11
Add ScanResultsAssembler class

jalkire:
#### Describe the change
Added `ScanResultsAssembler` class, which provides methods used to assemble a `ScanResults` object. This class is intended to be used to create the return object for `Scanner.Scan()` once tests have completed.

### DIFF
--- a/src/Automation/ScanResultsAssembler.cs
+++ b/src/Automation/ScanResultsAssembler.cs
@@ -9,7 +9,7 @@ using System.Linq;
 namespace Axe.Windows.Automation
 {
     /// <summary>
-    /// Provides methods used to assemble a ScanResults object
+    /// Provides methods used to assemble a <see cref="ScanResults"/> object
     /// </summary>
     public static class ScanResultsAssembler
     {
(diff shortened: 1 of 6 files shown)


src/Automation/ScanResultsAssembler.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        /// <summary>
+        /// Assembles failed scans from the provided element
+        /// </summary>
+        /// <param name="element">Root element from which scan results will be assembled</param>
+        /// <returns>A ScanResults object containing the relevant errors and error count</returns>
+        public static ScanResults AssembleScanResultsFromElement(A11yElement element)
+        {
+            var errors = new List<ScanResult>();
+            int count = AssembleScanResults(errors, element, null);
```
RobGallo: This variable isn't really necessary because you can get the count from the returned list with as much clarity.

src/Automation/ScanResultsAssembler.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <summary>
+        /// Assembles errors recursively starting at the given element
+        /// </summary>
+        /// <param name="errors">Where the results will be accumulated</param>
+        /// <param name="element">The root element to check</param>
+        /// <param name="parent">The root element to check</param>
+        /// <returns>Count of errors found</returns>
+        internal static int AssembleScanResults(List<ScanResult> errors, A11yElement element, ElementInfo parent)
+        {
+            if (element.ScanResults == null) throw new ArgumentException("element.ScanResults must not be null");
```
RobGallo: When referring to a symbol name in a string, it's a good idea to use nameof(). That will ensure the string is changed if the symbol is renamed.

src/Automation/ScanResultsAssembler.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /// <summary>
+        /// Assembles errors recursively starting at the given element
+        /// </summary>
+        /// <param name="errors">Where the results will be accumulated</param>
+        /// <param name="element">The root element to check</param>
+        /// <param name="parent">The root element to check</param>
+        /// <returns>Count of errors found</returns>
+        internal static int AssembleScanResults(List<ScanResult> errors, A11yElement element, ElementInfo parent)
```
RobGallo: Shouldn't we validate all the parameters?

RobGallo: This method currently only returns results for errors. Might we change the name to reflect that?

src/Automation/ScanResultsAssembler.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            if (element.ScanResults == null) throw new ArgumentException("element.ScanResults must not be null");
+
+            int failedCount = 0;
```
RobGallo: As mentioned above, I think we can get rid of the concept of returning the count. That will simplify the code in this method.

src/Automation/ScanResultsAssembler.cs
```diff
@@ -0,0 +1,89 @@
...
+            ElementInfo elementInfo = MakeElementInfoFromElement(element, parent);
+            IEnumerable<RuleResult> results = GetFailedRuleResultsFromElement(element);
```
RobGallo: I wouldn't object if you moved the function call into the for loop.

src/Automation/ScanResultsAssembler.cs
jalkire: Good point

src/Automation/ScanResultsAssembler.cs
jalkire: nameof used

src/Automation/ScanResultsAssembler.cs
jalkire: Parameters now validated more appropriately

jalkire: Name updated

src/Automation/ScanResultsAssembler.cs
jalkire: Agreed

src/Automation/ScanResultsAssembler.cs
jalkire: GetFailedRuleResultsFromElement now called from loop

src/Automation/ScanResultsAssembler.cs
```diff
@@ -0,0 +1,86 @@
...
+    /// Provides methods used to assemble a <see cref="ScanResults"/> object
+    /// </summary>
+    public static class ScanResultsAssembler
+    {
+        /// <summary>
+        /// Assembles failed scans from the provided element
+        /// </summary>
+        /// <param name="element">Root element from which scan results will be assembled</param>
+        /// <returns>A ScanResults object containing the relevant errors and error count</returns>
+        public static ScanResults AssembleScanResultsFromElement(A11yElement element)
```
DaveTryon: >A11yElement [](start = 65, length = 11)

Should this be ICoreA11yElement instead of A11yElement?

jalkire: `ICoreA11yElement` is internal to Axe.Windows.Core.Bases. And doesn't expose Properties, Patterns, or ScanResults. Is there something I'm missing about how I could use it here?

PR #53, opened 2019-06-12
Remove existing string-based options  from AxeWindows

RobGallo:
#### Describe the change

The strings-based parameter system is no longer necessary because we no longer support a PowerShell cmdlet directly. This has been replaced with the use of the Config object.

Changes:
- Remove CommandParametersExtensions because the old parameter mechanism is no longer used
- Create OutputFileHelper class to Create directories and generate file names
- Remove LocationHelper and replace with OutputFileHelper because
    - OutputFileHelper generates file names whereas LocationHelper did not
    - LocationHelper required strings specifying files extension whereas OutputFileHelper does not
    - OutputFileHelper uses abstractions for system calls, making it possible to unit test without Fakes
- add abstractions and concretions for interfaces that wrap System calls (useful in unit testing and dependency inversion)
- Remove StartCommand and StopCommand because they are unnecessary in a scheme which does not save configuration options
- Remove CommandParameters because the string dictionary system of command parameters is no longer necessary
- Remove CommandConstStrings because they are no longer used
- Update AutomationSession constructor to take a Config object instead of CommandParameters
- Update Snapshot.Execute to take a Config object instead of CommandParameters
- Replace the use of LocationHelper with OutputFileHelper
- Remove the code related to the option for retaining output files when no errors are found because the option no longer exists
- Comment out code which generated sarif files because AxeWindows does not support sarif file output at this time. Leaving the code in because we do plan to support sarif 2.1
- Rename TargetElementLocator.LocateElement to LocateRootElement to be more complete and accurate
- Update LocateRootElement to take a process id rather than CommandParameters which is no longer used
- Delete unit test which are no longer used
- Remove unit tests for snapshot temporarily  so a review of the code can be done while the unit tests are still uhnunder under construction


src/Automation/Automation.csproj
```diff
@@ -99,7 +102,7 @@
     <None Include="ReadMe.md" />
   </ItemGroup>
   <ItemGroup>
-    <DropSignedFile Include="$(OutDir)\Axe.Windows.Automation.dll" />
+    <DropSignedFile Include="bin\Debug\\Axe.Windows.Automation.dll" />
```
jalkire: Was this change to use the Debug file intentional?

src/Automation/AutomationSession.cs
```diff
@@ -19,14 +19,14 @@ internal class AutomationSession
         /// <summary>
         /// ctor - Initializes the app in automation mode
         /// </summary>
-        /// <param name="parameters">The parameters from the command line</param>
-        private AutomationSession(CommandParameters parameters)
+        /// <param name="config">A set of configuration options</param>
+        private AutomationSession(Config config)
```
jalkire: It appears to me that you removed all non-test uses of AutomationSession. Any reason to keep it around?

jalkire: On second thought, I'm guessing we'll want to use this when `Scanner.Scan()` is implemented

src/AutomationTests/TargetElementLocatorUnitTests.cs
```diff
@@ -16,8 +16,7 @@ public void LocateElement_NoTargetSpecifiedInParameters_ThrowsAutomationExceptio
         {
             try
             {
-                CommandParameters parameters = new CommandParameters(new Dictionary<string, string>(), string.Empty);
-                TargetElementLocator.LocateElement(parameters);
+                TargetElementLocator.LocateRootElement(42);
```
jalkire: This change doesn't seem to match what the test is testing for: "NoTargetSpecifiedInParameters". It seems equivalent to the other test, "SpecifiedPIDNotExist". I think there is no need for this test anymore, since it is not longer possible to not specify a PID.

src/Automation/OutputFileHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+            _dateTime = systemFactory.CreateSystemDateTime();
+            if (_dateTime == null) throw new InvalidOperationException($"Expected {nameof(_dateTime)} not to be null");
+
+            var environment = systemFactory.CreateSystemEnvironment();
+            if (environment == null) throw new InvalidOperationException($"Expected {nameof(environment)} not to be null");
+
+            var directory = systemFactory.CreateSystemIODirectory();
+            if (directory == null) throw new InvalidOperationException($"Expected {nameof(directory)} not to be null");
+
+            if (!String.IsNullOrWhiteSpace(outputDirectory))
```
DaveTryon: >S [](start = 17, length = 1)

Other places in this file use string with a lower case s. Should we use lower case here for consistency?

src/Automation/Abstractions/ISystemDateTime.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Abstractions
+{
+    internal interface ISystemDateTime
```
DaveTryon: >internal [](start = 4, length = 8)

I'd love to see some /// comments to capture the intent of these classes that are being used to inject seams

src/Automation/Interfaces/IOutputFileHelper.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+namespace Axe.Windows.Automation
+{
+    internal interface IOutputFileHelper
```
DaveTryon: >internal [](start = 4, length = 8)

Doesn't this interface need to be public, since it's part of the public API in SnapshotCommand?

src/Automation/SnapshotCommand.cs
```diff
@@ -37,19 +38,20 @@ private class ScanResultAccumulator
         /// <summary>
         /// Execute the Start command. Used by both .NET and by PowerShell entry points
```
DaveTryon: >Used by both .NET and by PowerShell entry points [](start = 39, length = 48)

Should this comment be removed?

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -177,11 +117,13 @@ public void Execute_UnableToSelectCandidateElement_ThrowsAutomationException_Err
...
+                InitializeShims(selectAction:selectAction, shimTargetElementLocator: true);
+
+                var config = Config.Builder.ForProcessId(-1).Build();
 
                 try
                 {
-                    SnapshotCommandResult result = SnapshotCommand.Execute(new Dictionary<string, string>());
+                    SnapshotCommandResult result = SnapshotCommand.Execute(config, OutputFileHelperStub);
                 }
                 catch (AxeWindowsAutomationException ex)
```
DaveTryon: >catch [](start = 16, length = 5)

Many of these tests look like they should fail if no Exception is thrown, but there's no enforcement

src/Automation/Automation.csproj
DaveTryon: Seems odd to me

---
In reply to: [293140526](https://github.com/microsoft/axe-windows/pull/53#discussion_r293140526) [](ancestors = 293140526)

DaveTryon:
>         public void NewInstance_ClearInstance_NotPowerShell_AssemblyResolverIsNull_DoesNotThrow()

Seems like this unit test ought to go away. I probably missed it when I removed the assembly resolver

---
Refers to: src/AutomationTests/AutomationSessionUnitTests.cs:67 in 7d7732c. [](commit_id = 7d7732c3244a051e09f373d9c89abe12bab544d0, deletion_comment = False)

src/Automation/OutputFileHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            try
+            {
+                // the following throws various exceptions if the given path is invalid
+                Path.GetFullPath(path);
+
+                if (!Path.IsPathRooted(path))
+                    throw new Exception(DisplayStrings.ErrorIsNotFullPath);
+            }
+            catch (System.Exception ex)
```
DaveTryon: >System. [](start = 19, length = 7)

Not needed?

DaveTryon: "System." is not needed

```suggestion
            catch (Exception ex)
```

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -203,12 +145,14 @@ public void Execute_UnableToSetTestModeDataContext_ThrowsAutomationException_Err
...
                     shimUiFramework: true, setTestModeSucceeds: false);
 
+                var config = Config.Builder.ForProcessId(-1).Build();
+
                 try
                 {
-                    SnapshotCommandResult result = SnapshotCommand.Execute(new Dictionary<string, string>());
+                    SnapshotCommandResult result = SnapshotCommand.Execute(config, OutputFileHelperStub);
                 }
                 catch (System.Exception ex)
```
DaveTryon: "System." is not needed here

```suggestion
                catch (Exception ex)
```

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -259,13 +205,15 @@ public void Execute_UpperBoundIsReached_ThrowsAutomationException_ErrorAutomatio
...
                     setTestModeSucceeds: true);
 
+                var config = Config.Builder.ForProcessId(-1).Build();
+
                 try
                 {
-                    SnapshotCommandResult result = SnapshotCommand.Execute(new Dictionary<string, string>());
+                    SnapshotCommandResult result = SnapshotCommand.Execute(config, OutputFileHelperStub);
                 }
                 catch (System.Exception ex)
```
DaveTryon: "System." is not needed here

```suggestion
                catch (Exception ex)
```

src/Automation/OutputFileHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        public const string DefaultOutputDirectoryName = "AxeWindowsOutputFiles";
+        public const string DefaultFileNameBase = "AxeWindows";
+
+        public OutputFileHelper(string outputDirectory, ISystemFactory systemFactory)
+        {
+            if (systemFactory == null) throw new ArgumentNullException(nameof(systemFactory));
+
+            _dateTime = systemFactory.CreateSystemDateTime();
+            if (_dateTime == null) throw new InvalidOperationException($"Expected {nameof(_dateTime)} not to be null");
```
DaveTryon: >_dateTime == null [](start = 16, length = 17)

Are there scenarios where the production code will return null for _dateTime, environment, or directory? If not, it seems like these checks are not helpful in production code

src/Automation/SnapshotCommand.cs
```diff
@@ -37,19 +38,20 @@ private class ScanResultAccumulator
...
-        public static SnapshotCommandResult Execute(Dictionary<string, string> primaryConfig)
+        public static SnapshotCommandResult Execute(Config config, IOutputFileHelper outputFileHelper)
         {
             return ExecutionWrapper.ExecuteCommand<SnapshotCommandResult>(() =>
             {
-                CommandParameters parameters = new CommandParameters(primaryConfig,
-                    AutomationSession.Instance().SessionParameters);
-                LocationHelper locationHelper = new LocationHelper(parameters);
-                ElementContext ec = TargetElementLocator.LocateElement(parameters);
+            if (outputFileHelper == null) throw new ArgumentNullException(nameof(outputFileHelper));
```
DaveTryon: Indent is incorrect

```suggestion
                if (outputFileHelper == null) throw new ArgumentNullException(nameof(outputFileHelper));
```

src/Automation/SnapshotCommand.cs
```diff
@@ -70,28 +72,28 @@ public static SnapshotCommandResult Execute(Dictionary<string, string> primaryCo
...
+                            if (accumulator.MayHaveErrors)
                             {
-                                ScreenShotAction.CaptureScreenShot(ec2.Id);
-                                SaveAction.SaveSnapshotZip(locationHelper.GetOutputFilePath(), ec2.Id, ec2.Element.UniqueId, A11yFileMode.Test);
-                                if (locationHelper.IsSarifExtension())
+                                if (config.OutputFileFormat.HasFlag(OutputFileFormat.A11yTest))
                                 {
-                                    // Generate SARIF file.
-                                    SaveAction.SaveSarifFile(locationHelper.GetSarifFilePath(), ec2.Id, !locationHelper.IsAllOption());
+                                ScreenShotAction.CaptureScreenShot(ec2.Id);
```
DaveTryon: Indent is incorrect

```suggestion
                                    ScreenShotAction.CaptureScreenShot(ec2.Id);
```

src/Automation/SnapshotCommand.cs
```diff
@@ -70,28 +72,28 @@ public static SnapshotCommandResult Execute(Dictionary<string, string> primaryCo
...
+                                SaveAction.SaveSnapshotZip(a11yTestOutputFile, ec2.Id, ec2.Element.UniqueId, A11yFileMode.Test);
```
DaveTryon: Indent is incorrect

```suggestion
                                    SaveAction.SaveSnapshotZip(a11yTestOutputFile, ec2.Id, ec2.Element.UniqueId, A11yFileMode.Test);
```

src/Automation/OutputFileHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            return Path.Combine(_outputDirectory
+                , GetNewFileName()
+                + ".a11ytest");
+        }
+
+        private string GetNewFileName()
+        {
+            var now = _dateTime.Now;
+            if (now == null) throw new InvalidOperationException();
```
DaveTryon: >now == null [](start = 16, length = 11)

Can this happen in production code? If not, do we need the check?

src/AutomationTests/OutputFileHelperUnitTests.cs
```diff
@@ -0,0 +1,223 @@
...
+
+        [TestMethod]
+        [Timeout(1000)]
+        public void OutputFileHelperCtor_CreatesNonexistentDirectory()
+        {
+            System.Diagnostics.Debug.WriteLine("here");
+            try
+            {
+                var mockSystemFactory = new Mock<ISystemFactory>(MockBehavior.Strict);
+            var mockDateTime = new Mock<ISystemDateTime>(MockBehavior.Strict);
```
DaveTryon: Lines 158-177 should be indented one more tab

src/AutomationTests/OutputFileHelperUnitTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            var mockDirectory = new Mock<ISystemIODirectory>(MockBehavior.Strict);
+            mockSystemFactory.Setup(x => x.CreateSystemDateTime()).Returns(mockDateTime.Object);
+            mockSystemFactory.Setup(x => x.CreateSystemEnvironment()).Returns(mockEnvironment.Object);
+            mockSystemFactory.Setup(x => x.CreateSystemIODirectory()).Returns(mockDirectory.Object);
+
+            string directory = @"c:\NonexistentDirectory";
+            mockDirectory.Setup(x => x.Exists(directory)).Returns(false);
+            mockDirectory.Setup(x => x.CreateDirectory(directory)).Returns<System.IO.DirectoryInfo>(null);
+
+
```
DaveTryon: Remove this line? It's nothing but whitespaces
```suggestion
```

src/Automation/OutputFileHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            var nowString = $"{now:yy-MM-dd_HH-mm-ss}";
```
DaveTryon: >ss [](start = 50, length = 2)

Is second granularity sufficient? I can imagine a case where a unit test might trigger 2 scans within a second...

src/Automation/Abstractions/ISystemDateTime.cs
RobGallo: I'm thinking perhaps such classes belong in their own project whose intent is clear what they're for. I didn't have time to do that in this rev.

src/Automation/OutputFileHelper.cs
RobGallo: Right you are.

src/Automation/OutputFileHelper.cs
RobGallo: It isn't. But I'm planning to revisit this before the feature is done.

src/Automation/AutomationSession.cs
RobGallo: I've moved the disposal into the SnapshotCommand.Execute

src/AutomationTests/TargetElementLocatorUnitTests.cs
RobGallo: It appears I may have removed the duplicate test already. Please recheck when you look again.

jalkire: The current file now has the (almost) same test twice

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -25,38 +25,7 @@ public class SnapshotCommandUnitTests
...
-            if (matchExactString)
-            {
-                Assert.AreEqual(expectedString, result.SummaryMessage);
-            }
-            else
-            {
-                Assert.IsTrue(result.SummaryMessage.Contains(expectedString), "\"" + result.SummaryMessage+ "\" doesn't contain \"" + expectedString + "\"");
-            }
-        }
+        private IOutputFileHelper OutputFileHelperStub = new Axe.Windows.Automation.Fakes.StubIOutputFileHelper();
```
jalkire: This use of Fakes needs to be inside of #if FAKES_SUPPORTED

PR #54, opened 2019-06-13
Make ScanResultsAssembler non-static and add interface

jalkire:
#### Describe the change
To improve future unit testing scenarios, this PR makes ScanResultsAssembler non-static and introduces the IScanResultsAssembler interface. ScanResultsAssembler is still not in use, so there is no impact on other classes.

### DIFF
--- a/src/Automation/Interfaces/IScanResultsAssembler.cs
+++ b/src/Automation/Interfaces/IScanResultsAssembler.cs
@@ -2,12 +2,12 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using Axe.Windows.Core.Bases;
 
-namespace Axe.Windows.Automation.Interfaces
+namespace Axe.Windows.Automation
 {
     /// <summary>
     /// Provides methods used to assemble a <see cref="ScanResults"/> object
     /// </summary>
-    public interface IScanResultsAssembler
+    internal interface IScanResultsAssembler
     {
         /// <summary>
         /// Assembles failed scans from the provided element
--- a/src/Automation/Interfaces/IScanResultsAssembler.cs
+++ b/src/Automation/Interfaces/IScanResultsAssembler.cs
@@ -1,0 +1,17 @@
+﻿using Axe.Windows.Core.Bases;
+
+namespace Axe.Windows.Automation.Interfaces
+{
+    /// <summary>
+    /// Provides methods used to assemble a <see cref="ScanResults"/> object
+    /// </summary>
+    public interface IScanResultsAssembler
+    {
+        /// <summary>
+        /// Assembles failed scans from the provided element
+        /// </summary>
+        /// <param name="element">Root element from which scan results will be assembled</param>
+        /// <returns>A ScanResults object containing the relevant errors and error count</returns>
+        ScanResults AssembleScanResultsFromElement(A11yElement element);
+    }
+}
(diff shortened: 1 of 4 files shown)


src/Automation/Interfaces/IScanResultsAssembler.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+
+namespace Axe.Windows.Automation.Interfaces
+{
+    /// <summary>
+    /// Provides methods used to assemble a <see cref="ScanResults"/> object
+    /// </summary>
+    public interface IScanResultsAssembler
```
RobGallo: This should be internal instead of public. No one outside the assembly needs to know about it.

src/Automation/ScanResultsAssembler.cs
```diff
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Automation.Interfaces;
```
RobGallo: Even though the interface files are in a subdirectory called "interfaces", I believe I've been avoiding actually putting them into a "Interfaces" namespace. So it might be good to follow that pattern.

src/Automation/ScanResultsAssembler.cs
```diff
@@ -11,14 +12,14 @@ namespace Axe.Windows.Automation
     /// <summary>
     /// Provides methods used to assemble a <see cref="ScanResults"/> object
     /// </summary>
-    public static class ScanResultsAssembler
+    public class ScanResultsAssembler : IScanResultsAssembler
```
RobGallo: This class should be internal as well because it is not intended to be used by clients.

src/Automation/ScanResultsAssembler.cs
```diff
@@ -37,7 +38,7 @@ public static ScanResults AssembleScanResultsFromElement(A11yElement element)
         /// <param name="errors">Where the <see cref="ScanResult"/> objects created from errors will be added</param>
         /// <param name="element">Root element from which errors will be assembled</param>
         /// <param name="parent">The parent of element</param>
-        internal static void AssembleErrorsFromElement(List<ScanResult> errors, A11yElement element, ElementInfo parent)
+        internal void AssembleErrorsFromElement(List<ScanResult> errors, A11yElement element, ElementInfo parent)
```
RobGallo: I think you can actually leave the private static functions static because they don't rely on any member data.

src/Automation/Interfaces/IScanResultsAssembler.cs
jalkire: Now internal

src/Automation/ScanResultsAssembler.cs
jalkire: Changed namespace

src/Automation/ScanResultsAssembler.cs
jalkire: Now internal

src/Automation/ScanResultsAssembler.cs
jalkire: Restored non-public functions to be static

PR #55, opened 2019-06-13
Integrate ScanResultsAssembler into Automation

RobGallo:
#### Describe the change

- Implement Scanner.Scan based on updated SnapshotCommand
- Remove abandoned command results classes
- Remove snapshot unit tests which are no longer relevant to the new ScanResults format


src/Automation/Scanner.cs
```diff
@@ -24,7 +31,10 @@ internal Scanner(Config config)
         /// <returns></returns>
         public ScanResults Scan()
         {
-            throw new NotImplementedException();
+            lock (LockObject)
```
DaveTryon: >LockObject [](start = 18, length = 10)

SnapshotCommand.Execute calls ExecutionWrapper, which already provides a lock. Can we pick just one of the two?

src/Automation/Scanner.cs
```diff
@@ -9,13 +9,20 @@ namespace Axe.Windows.Automation
     /// </summary>
     class Scanner : IScanner
     {
+        private static Object LockObject = new Object();
```
DaveTryon: readonly, as well?
```suggestion
        private static readonly Object LockObject = new Object();
```

src/Automation/SnapshotCommand.cs
```diff
@@ -88,29 +73,11 @@ public static SnapshotCommandResult Execute(Config config, IOutputFileHelper out
...
-                                string summaryMessage;
 
-                                if (accumulator.MayHaveErrors)
-                                {
-                                    summaryMessage = string.Format(CultureInfo.InvariantCulture, DisplayStrings.SnapshotDetailViolationsFormat, accumulator.Failed, accumulator.Inconclusive, a11yTestOutputFile);
-                                }
-                                else
-                                {
-                                    summaryMessage = DisplayStrings.SnapshotDetailNoViolationsDataDiscarded;
+                                    results.OutputFile = (a11yTestOutputFile, null);
```
DaveTryon: >(a11yTestOutputFile, null) [](start = 57, length = 26)

This line is pre-set for when we add the Sarif 2.1 output, but it will require touching if/when a third output file is added (Sarif 3.0, maybe). It may be overengineering, but I'd love to see all of the OutputFile creation be pulled into a tiny factory class, so that if/when new formats are added, the only change in inside that factory class. This like would be something like:

results.OutputFile = OutputFileFactory.FromA11yTestOutputFile(a11yTestOutputFile);

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -379,102 +306,24 @@ public void Execute_OnlyFailuresInPOI_SavesFile_ReturnsComplete_FailuresOnly()
...
-                InitializeShims(shimTargetElementLocator: true,
-                    selectAction: selectAction, elementBoundExceeded: false, shimUiFramework: true,
-                    setTestModeSucceeds: true, shimScreenCapture: true, shimSnapshot: true, shimSarif: true);
-
-                var config = Config.Builder.ForProcessId(-1).Build();
-
-                SnapshotCommandResult result = SnapshotCommand.Execute(config, OutputFileHelperStub);
+                    results = SnapshotCommand.Execute(config, OutputFileHelperStub, resultsAssembler);
+                }
+                catch (System.Exception ex)
```
DaveTryon: "System." is not needed

```suggestion
                catch (Exception ex)
```

src/Automation/SnapshotCommand.cs
```diff
@@ -88,29 +73,11 @@ public static SnapshotCommandResult Execute(Config config, IOutputFileHelper out
                                 if (locationHelper.IsSarifExtension())
```
DaveTryon: Should have one more indent
```suggestion
                                    if (locationHelper.IsSarifExtension())
```

src/Automation/SnapshotCommand.cs
```diff
@@ -88,29 +73,11 @@ public static SnapshotCommandResult Execute(Config config, IOutputFileHelper out
...
                                     // SaveAction.SaveSarifFile(outputFileHelper.GetNewSarifFilePath(), ec2.Id, !locationHelper.IsAllOption());
```
DaveTryon: Should have one more indent
```suggestion
                                        // SaveAction.SaveSarifFile(outputFileHelper.GetNewSarifFilePath(), ec2.Id, !locationHelper.IsAllOption());
```

src/Automation/Scanner.cs
RobGallo: Yes. I will take this out in an upcoming commit.

src/Automation/SnapshotCommand.cs
RobGallo: I can definitely get behind the idea. If we don't have time to make such a change during this feature, we can do it when we add the sarif format.

src/AutomationTests/SnapshotCommandUnitTests.cs
RobGallo: You're correct. This should not be here. I will get rid of it when I rework the snapshot unit tests to remove Fakes. That change is very soon in coming.

src/Automation/SnapshotCommand.cs
RobGallo: I can definitely get behind this idea. If we don't have time to make the change during this feature. We should do that when we add sarif.

src/AutomationTests/SnapshotCommandUnitTests.cs
RobGallo: Your right. This was only for debugging. I will take it out when I rework the snapshot command unit tests to remove Fakes.

PR #56, opened 2019-06-14
Correct the code to rethrow an AxeWindowsAutomationException

DaveTryon:
#### Describe the change
The code to rethrow a caught AxeWindowsAutomationException was using the wrong syntax, causing the Exception to reach the caller without its stack trace. Reference: https://blogs.msdn.microsoft.com/perfworld/2009/06/15/how-can-i-throw-an-exception-without-losing-the-original-stack-trace-information-in-net/

Fixed the code, added a unit test to demonstrate the bad behavior and to pin the correct behavior.

### DIFF
--- a/src/Automation/ExecutionWrapper.cs
+++ b/src/Automation/ExecutionWrapper.cs
@@ -22,9 +22,9 @@ namespace Axe.Windows.Automation
                 {
                     return command();
                 }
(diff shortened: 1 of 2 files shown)


DaveTryon:
Force rebuild via close/open

PR #57, opened 2019-06-16
Refactor to remove Fakes

RobGallo:
#### Describe the change

Note: No Fakes were actually removed as part of this change, unless you count the unnecessary test files deleted, or the constantly in-flux SnapshotCommandUnitTests file which was excluded from the project because those tests are now under construction.

Changes:
- Create ITargetElementLocator interface for mocking and, you know, SOLID principles
- Create IScanTools interface to contain the set of interfaces required by the snapshot command
- Create classes to implement the new interfaces
- Update factories to create objects implementing the new interfaces
- Remove unit test files which are no longer used
- temporarily exclude snapshot unit tests while the SnapshotCommand is under construction


src/Automation/Interfaces/ITargetElementLocator.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+
+namespace Axe.Windows.Automation
+{
+    internal interface ITargetElementLocator
+    {
+        A11yElement LocateRootElement(int processId);
```
DaveTryon: >A11yElement [](start = 8, length = 11)

Can this be ICoreA11yElement?

src/Automation/SnapshotCommand.cs
```diff
@@ -14,30 +14,33 @@
...
         /// <returns>A SnapshotCommandResult that describes the result of the command</returns>
-        public static ScanResults Execute(Config config, IOutputFileHelper outputFileHelper, IScanResultsAssembler resultsAssembler)
+        public static ScanResults Execute(Config config, IScanTools scanTools)
         {
             return ExecutionWrapper.ExecuteCommand<ScanResults>(() =>
             {
-                if (outputFileHelper == null) throw new ArgumentNullException(nameof(outputFileHelper));
-                if (resultsAssembler == null) throw new ArgumentNullException(nameof(resultsAssembler));
+                if (scanTools == null) throw new ArgumentNullException(nameof(scanTools));
+                if (scanTools.OutputFileHelper == null) throw new ArgumentNullException(nameof(scanTools.OutputFileHelper));
```
DaveTryon: >OutputFileHelper [](start = 30, length = 16)

These checks already exist in the ScanTools class. If we posit that IScanTools *must* provide a non-null value here, then we can       omit the extra checks.

src/Automation/SnapshotCommand.cs
```diff
@@ -86,5 +74,25 @@ public static ScanResults Execute(Config config, IOutputFileHelper outputFileHel
...
+                ScreenShotAction.CaptureScreenShot(elementContext.Id);
+
+                a11yTestOutputFile = outputFileHelper.GetNewA11yTestFilePath();
+                SaveAction.SaveSnapshotZip(a11yTestOutputFile, elementContext.Id, elementContext.Element.UniqueId, A11yFileMode.Test);
+            }
+
+#if NOT_CURRENTLY_SUPPORTED
+                                if (locationHelper.IsSarifExtension())
+                                    // SaveAction.SaveSarifFile(outputFileHelper.GetNewSarifFilePath(), ec2.Id, !locationHelper.IsAllOption());
+#endif
```
DaveTryon: Can we just remove lines 90-93 for now? They're using the old classes and old assumptions, and could easily be replaced with a comment along the lines of:

// Future: If needed, generate Sarif output via SaveAction.SaveSarifFIle

src/Automation/ScannerFactory.cs
```diff
@@ -16,9 +16,9 @@ public static class ScannerFactory
         /// <returns></returns>
         public static IScanner CreateScanner(Config config)
         {
-            return new Scanner(config
-                , Factory.CreateOutputFileHelper(config.OutputDirectory)
-                , new ScanResultsAssembler());
+            // Factory.CreateOutputFileHelper(config.OutputDirectory)
```
DaveTryon: Should line 19 be removed, rather than commented out?

src/Automation/Scanner.cs
```diff
@@ -30,7 +27,7 @@ internal Scanner(Config config, IOutputFileHelper outputFileHelper, ScanResultsA
         /// <returns></returns>
         public ScanResults Scan()
         {
-            return SnapshotCommand.Execute(_config, _outputFileHelper, _scanResultsAssembler);
+                return SnapshotCommand.Execute(_config, _scanTools);
```
DaveTryon: Has an extra indent
```suggestion
            return SnapshotCommand.Execute(_config, _scanTools);
```

src/AutomationTests/TargetElementLocatorUnitTests.cs
```diff
@@ -1,45 +0,0 @@
...
-        [Timeout (1000)]
-        public void LocateElement_NoTargetSpecifiedInParameters_ThrowsAutomationException_ErrorAutomation007()
-        {
-            try
-            {
-                TargetElementLocator.LocateRootElement(-1);
-            }
-            catch (AxeWindowsAutomationException ex)
-            {
-                Assert.IsTrue(ex.Message.Contains("Automation017:"));
```
DaveTryon: >Automation017 [](start = 51, length = 13)

Is there (or should there be) a replacement unit test to confirm the Automation017 message?

src/Automation/Interfaces/IScanTools.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Axe.Windows.Automation
+{
+    internal interface IScanTools
```
DaveTryon: >internal [](start = 4, length = 8)

/// comments to capture interface intent are always appropriate

src/Automation/ScanTools.cs
```diff
@@ -0,0 +1,23 @@
...
+        public IOutputFileHelper OutputFileHelper { get; }
+        public IScanResultsAssembler ResultsAssembler { get; }
+        public ITargetElementLocator TargetElementLocator { get; }
+
+        public ScanTools(IOutputFileHelper outputFileHelper, IScanResultsAssembler resultsAssembler, ITargetElementLocator targetElementLocator)
+        {
+            if (outputFileHelper == null) throw new ArgumentNullException(nameof(outputFileHelper));
+            if (resultsAssembler == null) throw new ArgumentNullException(nameof(resultsAssembler));
+            if (targetElementLocator == null) throw new ArgumentNullException(nameof(targetElementLocator));
+            this.OutputFileHelper = outputFileHelper;
```
DaveTryon: >this. [](start = 12, length = 5)

Can we remove the unnecessary "this." prefixes? This is a really old coding pattern going back into the Visual Studio code base (and into the Exchange code base before that). This pattern is discouraged by Roslyn checks and I'd love to see us stop using it.

src/Automation/Factory.cs
```diff
@@ -8,6 +8,14 @@ namespace Axe.Windows.Automation
 {
     static class Factory
     {
+        static public IScanTools CreateScanTools()
+        {
+            return new ScanTools(
+                CreateOutputFileHelper(null),
```
DaveTryon: >null [](start = 39, length = 4)

How does the Config's OutputDirectory property override the null that is getting set by default? It feels like maybe the CreateScanTools method should accept an IConfig parameter

DaveTryon:
>   <PropertyGroup Condition="$(FAKES_SUPPORTED) == 1">

Do these changes also allow us to remove the 8 instances of FAKES_SUPPORTED that exist in the csproj file?

---
Refers to: src/AutomationTests/AutomationTests.csproj:44 in a1bd505. [](commit_id = a1bd50518a81cdecaf32dcd90ce76c67a870d4a9, deletion_comment = False)

src/Automation/Factory.cs
RobGallo: In changes that are already made and coming down the pipe, the Config is set to a parameter passed into the factory method.

src/Automation/Interfaces/ITargetElementLocator.cs
RobGallo: I would like that as well, but I don't have time to make that change if _anything_ goes awry. I aslo think it might be more appropriate when we try to remove A11yElement out of AI-Win altogether.

src/Automation/ScanTools.cs
RobGallo: You'll notice I don't do that very often. I did it in this case because the property names are exactly the same as the type names. But I have removed them because it sounds like they're no longer welcome in the modern world.

src/Automation/ScannerFactory.cs
RobGallo: It has been taken out in an upcoming PR.

src/Automation/SnapshotCommand.cs
RobGallo: This class only deals with the IScanTools abstraction and therefore cannot make that assumption.

DaveTryon: My point is that we can make it a requirement within the interface that the implementation has valid values. There's a proposal to provide non-nullable reference types in the future, maybe we ought to use it when it's available

---
In reply to: [294516127](https://github.com/microsoft/axe-windows/pull/57#discussion_r294516127) [](ancestors = 294516127)

src/Automation/SnapshotCommand.cs
RobGallo: I'm keeping it because having that code will make it easier once we support sarif 2.1.

src/AutomationTests/TargetElementLocatorUnitTests.cs
RobGallo: All the unit tests are changing.

PR #48, opened 2019-06-06
Remove PowerShell cmdlet code

RobGallo:
#### Describe the change

We are no longer building a PowerShell cmdlet because PowerShell users can directly consume the .Net assemblies.

> Note: After the PR has been created, certain checks will be kicked off. All of these checks must pass before a merge.

PR #58, opened 2019-06-17
Introducing automation e2e tests

jalkire:
#### Describe the change
- Add an E2E test to AutomationTests
- [Crossing the bridge](https://github.com/microsoft/axe-windows/pull/48#issuecomment-499989796) to update the check-automation script to use the new scan interface

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -1,0 +1,79 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Automation;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading;
+
+namespace Axe.Windows.AutomationTests
+{
+    [TestClass]
+    public class AutomationIntegrationTests
+    {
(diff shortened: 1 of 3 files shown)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -0,0 +1,79 @@
...
+                .Build();
+
+            var scanner = ScannerFactory.CreateScanner(config);
+
+            var output = scanner.Scan();
+
+            var regexForExpectedFile = $"{OutputDir.Replace("\\", "\\\\")}.*\\.a11ytest";
+
+            // Validate that we got the expected error count
+            Assert.AreEqual(12, output.ErrorCount);
```
DaveTryon: >12 [](start = 28, length = 2)

Having a hardcoded value here has the potential to break whenever rules change.

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -0,0 +1,79 @@
...
+            Assert.AreEqual(12, output.Errors.Count());
```
DaveTryon: Can we remove the duplicated constant?

```suggestion
            Assert.AreEqual(output.ErrorCount, output.Errors.Count());
```

src/AutomationTests/AutomationIntegrationTests.cs
jalkire: Updated to just make sure that there are > 0 errors. If there are ever 0, we should probably update the test app.

src/AutomationTests/AutomationIntegrationTests.cs
jalkire: Yes

PR #61, opened 2019-06-18
Add InternalScanner class

RobGallo:
#### Describe the change

This class extracts out the logic from SnapshotCommand which talks  to parts of AxeWindows that are outside the Automation package. This now means the logic of SnapshotCommand is now testable without Fakes.

### DIFF
--- a/src/Automation/Interfaces/IInternalScanner.cs
+++ b/src/Automation/Interfaces/IInternalScanner.cs
@@ -5,7 +5,7 @@ using System;
 
 namespace Axe.Windows.Automation
 {
-    delegate ResultsT InternalScannerCallback<ResultsT>(A11yElement element, Guid elementId);
+    delegate T InternalScannerCallback<T>(A11yElement element, Guid elementId);
 
     /// <summary>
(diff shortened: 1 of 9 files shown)


src/Automation/Interfaces/IInternalScanner.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using System;
+
+namespace Axe.Windows.Automation
+{
+    delegate ResultsT InternalScannerCallback<ResultsT>(A11yElement element, Guid elementId);
```
jalkire: Naming conventions for type parameters in C# are either a single T or T as a prefix, not a suffix. [Read more here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/generics/generic-type-parameters). Personally, I found the descriptive name here more confusing than I would have found just T.

Issue #42, opened 2019-05-31
[BUG] Adjacent identical hyperlinks should raise a warning instead of an error


karanbirsingh:
**Describe the bug**
The rule `Focusable sibling elements must not have the same Name and LocalizedControlType` fails when there are two adjacent Hyperlink elements. Hyperlinks should be excluded from this rule and moved into a warning based on the rationale given [here ](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-link.html)(`Best practice is that links with the same destination would have the same descriptions, but links with different purposes and destinations would have different descriptions`).

karanbirsingh:
Download, rename to feedback.a11ytest, and open in Accessibility Insights for Windows to see repro:
[feedback.zip](https://github.com/microsoft/axe-windows/files/3242403/feedback.zip)

DaveTryon:
Closing this out

PR #66, opened 2019-06-18
Exclude hyperlinks from unique name rule for siblings

karanbirsingh:
We had a false positive described in issue #42 . This PR removes the hyperlink from the unique name check for siblings. Instead a new rule is created for sibling hyperlinks and a warning is returned if the names are identical.

### DIFF
--- a/src/Rules/Library/HyperlinkNameShouldBeUnique.cs
+++ b/src/Rules/Library/HyperlinkNameShouldBeUnique.cs
@@ -1,4 +1,6 @@
-﻿using Axe.Windows.Core.Bases;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
 using Axe.Windows.Rules.PropertyConditions;
 using Axe.Windows.Rules.Resources;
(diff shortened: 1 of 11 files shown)


src/Rules/Library/HyperlinkNameShouldBeUnique.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Axe.Windows.Core.Bases;
```
RobGallo: Missing copyright.

karanbirsingh: Thanks Rob, added the copyright now

src/Rules/Library/HyperlinkNameShouldBeUnique.cs
```diff
@@ -0,0 +1,50 @@
...
+        public override EvaluationCode Evaluate(IA11yElement e)
+        {
+            if (e == null) throw new ArgumentException(nameof(e));
+            if (e.Parent == null) throw new ArgumentNullException(nameof(e.Parent));
+
+            var siblings = SiblingCount(EligibleHyperlink
+                & Name.Is(e.Name));
+            var count = siblings.GetValue(e);
+
+            return count == 1 ? EvaluationCode.Pass : EvaluationCode.Warning;
```
karanbirsingh: @RobGallo I was unable to compare siblings with 1 directly. I get the error `Cannot implicitly convert type 'Axe.Windows.Rules.Condition' to 'bool'`. We can chat when you're in about what I'm missing.

PR #72, opened 2019-06-19
Update package copyright to include contributors

RobGallo:
Open source!

#### Describe the change
(Please provide an overview of the change in this PR)

### DIFF
--- a/src/CI/Axe.Windows.nuspec
+++ b/src/CI/Axe.Windows.nuspec
@@ -8,12 +8,12 @@
     <description>The package provides the assemblies necessary to automate accessibility testing for Windows applications using the included axe-windows rule set.</description>
     <projectUrl>https://github.com/Microsoft/axe-windows</projectUrl>
     <iconUrl>https://github.com/Microsoft/axe-windows/raw/master/brand/brand-blue-128px.png</iconUrl>
-    <authors>Microsoft</authors>
+    <authors>Microsoft and contributors</authors>
     <owners></owners>
     <license type="expression">MIT</license>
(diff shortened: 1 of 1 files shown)


src/CI/Axe.Windows.nuspec
```diff
@@ -8,12 +8,12 @@
...
     <projectUrl>https://github.com/Microsoft/axe-windows</projectUrl>
     <iconUrl>https://github.com/Microsoft/axe-windows/raw/master/brand/brand-blue-128px.png</iconUrl>
-    <authors>Microsoft</authors>
+    <authors>Microsoft and contributors</authors>
     <owners></owners>
     <license type="expression">MIT</license>
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <language>en-US</language>
-    <copyright>© Microsoft Corporation. All rights reserved.</copyright>
+    <copyright>© Microsoft Corporation and contributors. All rights reserved.</copyright>
```
flyingUnderTheRadar: Is this a directive from Legal / OSS review? We have a copyright all over this project and all other repos that might need to changes if that is the case

DaveTryon: If it's supposed to apply to all files, then we probably need to update the PowerShell script to reflect that change (and it's going to touch nearly every file in the repo)

---
In reply to: [295427124](https://github.com/microsoft/axe-windows/pull/72#discussion_r295427124) [](ancestors = 295427124)

RobGallo:
This does not apply to the code, only to the package. And yes, this came from legal. Or at least I'm given to understand that it came from legal.

PR #79, opened 2019-06-25
Refactor the System abstractions to take the same form as System

RobGallo:
#### Describe the change

Making the System abstractions take the same form as System. I'm interested to know if this makes the system abstractions easier to understand or to work with.

### DIFF
--- a/src/Automation/Concretions/System.cs
+++ b/src/Automation/Concretions/System.cs
@@ -23,7 +23,7 @@ namespace Axe.Windows.Concretions
 
         public ISystemDateTime DateTime => _dateTime.Value;
 
-    public ISystemEnvironment Environment => _environment.Value;
+        public ISystemEnvironment Environment => _environment.Value;
 
         public ISystemIO IO => _io.Value;
     } // class
--- a/src/Automation/Concretions/System.cs
+++ b/src/Automation/Concretions/System.cs
@@ -1,0 +1,30 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Abstractions;
+using System;
+using System.Collections.Generic;
+
+namespace Axe.Windows.Concretions
+{
+    class System : ISystem
+    {
+        private readonly Lazy<ISystemDateTime> _dateTime;
+        private readonly Lazy<ISystemEnvironment> _environment;
+        private readonly Lazy<ISystemIO> _io;
+
+        public System(ISystemFactory factory)
+        {
+            if (factory == null) throw new ArgumentNullException(nameof(factory));
+
+            _dateTime = new Lazy<ISystemDateTime>(factory.CreateSystemDateTime);
+            _environment = new Lazy<ISystemEnvironment>(factory.CreateSystemEnvironment);
+            _io = new Lazy<ISystemIO>(factory.CreateSystemIO);
+        }
+
+        public ISystemDateTime DateTime => _dateTime.Value;
+
+    public ISystemEnvironment Environment => _environment.Value;
+
+        public ISystemIO IO => _io.Value;
+    } // class
+} // namespace
--- a/src/AutomationTests/OutputFileHelperUnitTests.cs
+++ b/src/AutomationTests/OutputFileHelperUnitTests.cs
@@ -15,90 +15,98 @@ namespace Axe.Windows.AutomationTests
         [TestMethod]
         [Timeout(1000)]
         [ExpectedException(typeof(ArgumentNullException))]
-        public void OutputFileHelperCtor_NullSystemFactory_ThrowsException()
+        public void OutputFileHelperCtor_NullSystem_ThrowsException()
         {
-            new OutputFileHelper(outputDirectory: null, systemFactory: null);
+            new OutputFileHelper(outputDirectory: null, system: null);
         }
 
         [TestMethod]
         [Timeout(1000)]
         public void OutputFileHelperCtor_NullSystemDateTime_ThrowsException()
         {
-            var mockSystemFactory = new Mock<ISystemFactory>(MockBehavior.Strict);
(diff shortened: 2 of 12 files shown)


src/Automation/Concretions/System.cs
```diff
@@ -0,0 +1,30 @@
...
+            if (factory == null) throw new ArgumentNullException(nameof(factory));
+
+            _dateTime = new Lazy<ISystemDateTime>(factory.CreateSystemDateTime);
+            _environment = new Lazy<ISystemEnvironment>(factory.CreateSystemEnvironment);
+            _io = new Lazy<ISystemIO>(factory.CreateSystemIO);
+        }
+
+        public ISystemDateTime DateTime => _dateTime.Value;
+
+    public ISystemEnvironment Environment => _environment.Value;
```
DaveTryon: Indent is off
```suggestion
        public ISystemEnvironment Environment => _environment.Value;
```

src/AutomationTests/SystemUnitTests.cs
```diff
@@ -0,0 +1,59 @@
...
+            var timeSpan = now2.Subtract(now1);
+
+            Assert.IsTrue(timeSpan.TotalSeconds <= 1);
+        }
+
+        [TestMethod]
+        [Timeout(1000)]
+        public void CurrentDirectory_Matches()
+        {
+            Assert.AreEqual(System.Environment.CurrentDirectory, _system.Environment.CurrentDirectory);
```
DaveTryon: >CurrentDirectory [](start = 47, length = 16)

This test feels like it's testing both System and SystemIO classes. Should they be split?

src/AutomationTests/OutputFileHelperUnitTests.cs
```diff
@@ -15,90 +15,98 @@ public class OutputFileHelperUnitTests
...
+            mockIO.Setup(x => x.Directory).Returns<ISystemIODirectory>(null);
 
-            Action action = () => new OutputFileHelper(outputDirectory: null, systemFactory: mockSystemFactory.Object);
+            Action action = () => new OutputFileHelper(outputDirectory: null, system: mockSystem.Object);
             var ex = Assert.ThrowsException<InvalidOperationException>(action);
             Assert.IsTrue(ex.Message.Contains("directory"));
 
-            mockSystemFactory.VerifyAll();
+            mockSystem.VerifyAll();
+            mockIO.VerifyAll();
```
DaveTryon: >mockIO [](start = 12, length = 6)

Should there be a mockEnvironment.VerifyAll here?

src/AutomationTests/OutputFileHelperUnitTests.cs
```diff
@@ -15,90 +15,98 @@ public class OutputFileHelperUnitTests
...
+            mockSystem.VerifyAll();
         }
 
         [TestMethod]
         [Timeout(1000)]
         public void OutputFileHelperCtor_NullSystemEnvironment_ThrowsException()
         {
-            var mockSystemFactory = new Mock<ISystemFactory>(MockBehavior.Strict);
+            var mockSystem = new Mock<ISystem>(MockBehavior.Strict);
             var mockDateTime = new Mock<ISystemDateTime>(MockBehavior.Strict);
```
DaveTryon: >mockDateTime [](start = 16, length = 12)

A pattern that I like to use in unit tests: Add this to the class:

private static readonly ISystemDateTime InertSystemDateTime = new Mock<ISystemDateTime>(MockBehavior.String).Object;

Then in the unit tests that should never call this mock, pass it in:

mockSystem.Setup(x => x.DateTime).Returns(InertSystemDateTime);

That makes it explicit that you're using a mock that should never be invoked and shortens the unit test by 1 line for each inert mock that you use.

I noticed this because I was trying to determine if there should be a mockDateTime.VerifyAll at the bottom of the test. With the Inert* pattern, _every_ mock created in the test has a VerifyAll, so it's trivial to catch cases where a VerifyAll should exist.

src/AutomationTests/SystemUnitTests.cs
```diff
@@ -0,0 +1,59 @@
...
+    using Path = System.IO.Path;
+
+    [TestClass]
+    public class SystemUnitTests
+    {
+        private readonly ISystem _system = SystemFactory.CreateSystem();
+
+        [TestMethod]
+        [Timeout(1000)]
+        public void DateTime_Matches()
```
DaveTryon: >DateTime_Matches [](start = 20, length = 16)

These unit tests are basically testing System methods. I'd suggest that what you really care about is that the class is mapping correctly to the underlying methods, so you can test these directly:

ISystem system = SystemFactory.CreateSystem();
Assert.AreEqual(System.DateTime, system.DateTime);
Assert.AreEqual(System.Environment, _system.Environment);
Assert.AreEqual(System.IO.Directory, _system.IO.Directory);

RobGallo: I thought about that. But I think what I'm most interested in is that the system I've set up returns the actual system values. And besides, the way I've set it up also tests that the interfaces are set up as wel.

DaveTryon: The class you're testing is to map the interfaces to system classes. Your validation is relying on the behavior of those underlying classes to prove the mapping is correct. You probably wouldn't do this if the underlying class were some other class that you provided, since it violates the principle of only testing one class at a time. If you prove that the underlying class _is_ the system class, you don't need to prove that the underlying class _acts_ like the system class

---
In reply to: [298803943](https://github.com/microsoft/axe-windows/pull/79#discussion_r298803943) [](ancestors = 298803943)

DaveTryon: Rob and I discussed this, and there's no way to do what I'm describing

---
In reply to: [299598150](https://github.com/microsoft/axe-windows/pull/79#discussion_r299598150) [](ancestors = 299598150,298803943)

PR #80, opened 2019-06-26
Attempt to sign just the NuGet package, not the binaries that it contains

DaveTryon:
#### Describe the change
Attempt to sign just the NuGet package, not the binaries that it contains. This change intentionally leaves it easy to re-enable the digital signing, if needed (a fuller scrub can be performed in the future, as needed). The best way to test these would probably be to create a branch in the Microsoft repo (not in a fork), queue a signed build against that branch, and examine the files that get build. The NuGet package should be signed, but the DLLs should not.

### DIFF
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -6,8 +6,8 @@ variables:
   DropFolder: 'signed'
   MAICreateNuget: 'true'
   PublicRelease: 'true'
-  SignAppForRelease: 'true'
-  MicroBuild_NuPkgSigningEnabled: 'true'
+  SignAppForRelease: 'false'
+  MicroBuild_NuPkgSigningEnabled: 'false'
   TeamName: 'Axe Windows'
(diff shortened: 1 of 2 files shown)


build/signedbuild.yml
```diff
@@ -295,6 +291,8 @@ jobs:
   - ComplianceDebug
   condition: and(succeeded(), succeeded())
   pool: VSEng-MicroBuildVS2017
+  variables:
+    MicroBuild_NuPkgSigningEnabled: 'true'
```
RobGallo: Why add signing here?

DaveTryon: oops--wrong phase

---
In reply to: [297877301](https://github.com/microsoft/axe-windows/pull/80#discussion_r297877301) [](ancestors = 297877301)

DaveTryon:
This was superseded by a different PR

PR #86, opened 2019-07-01
Add package dependencies to NuSpec file

RobGallo:
#### Describe the change

This means we will allow NuGet to manage getting dependency packages, instead of including dependency package files in our package.

Also, removed inclusion of system files from our package. These files should already exist on target machines as part of the .Net runtime.

### DIFF
--- a/src/CI/Axe.Windows.nuspec
+++ b/src/CI/Axe.Windows.nuspec
@@ -15,9 +15,13 @@
     <language>en-US</language>
     <copyright>© Microsoft and contributors</copyright>
     <tags>axe windows accessibility automation</tags>
+    <dependencies>
+        <dependency id="Newtonsoft.Json" version="9.0.1" />
+        <dependency id="Sarif.SDK" version="2.0.0" />
+    </dependencies>
   </metadata>
   <files>
-    <file src="bin\$configuration$\*.dll" target="lib\net471" />
+    <file src="bin\$configuration$\*.dll" exclude="bin\$configuration$\newtonsoft.json.dll;bin\$configuration$\system.*.dll;bin\$configuration$\sarif.dll" target="lib\net471" />
     <file src="bin\$configuration$\Axe.Windows.Automation.xml" target="lib\net471" />
   </files> 
 </package>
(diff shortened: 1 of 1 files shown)


src/CI/Axe.Windows.nuspec
```diff
@@ -15,9 +15,13 @@
...
     <copyright>© Microsoft and contributors</copyright>
     <tags>axe windows accessibility automation</tags>
+    <dependencies>
+        <dependency id="Newtonsoft.Json" version="9.0.1" />
+        <dependency id="Sarif.SDK" version="2.0.0" />
+    </dependencies>
   </metadata>
   <files>
-    <file src="bin\$configuration$\*.dll" target="lib\net471" />
+    <file src="bin\$configuration$\*.dll" exclude="bin\$configuration$\newtonsoft.json.dll;bin\$configuration$\system.*.dll;bin\$configuration$\sarif.dll" target="lib\net471" />
```
jalkire: Might we want to start explicitly including DLLs? 
With this approach, adding a new package would require the package adder to modify this file to exclude that package's DLLs. 
With the explicit approach, adding a new project to axe-windows would require the adder to modify this file to include the project DLL. 
Personally, the latter option feels more natural to me.

DaveTryon: I like the approach of specifying what's included. Axe.Windows*.dll and Axe.Windows.Automation.xml will cover it, I think

---
In reply to: [299118869](https://github.com/microsoft/axe-windows/pull/86#discussion_r299118869) [](ancestors = 299118869)

src/CI/Axe.Windows.nuspec
```diff
@@ -15,9 +15,13 @@
     <language>en-US</language>
     <copyright>© Microsoft and contributors</copyright>
     <tags>axe windows accessibility automation</tags>
+    <dependencies>
+        <dependency id="Newtonsoft.Json" version="9.0.1" />
+        <dependency id="Sarif.SDK" version="2.0.0" />
```
DaveTryon: >Sarif [](start = 24, length = 5)

Should this also include System.Collections.Immutable ?

PR #90, opened 2019-07-01
Updated documentation for automation

waabid:
#### Describe the change

This updates the documentation for how to consume and use AxeWindows for automation purposes.

### DIFF
--- a/docs/Automation.md
+++ b/docs/Automation.md
@@ -19,9 +19,9 @@ a solution where you can make those calls from separate processes.
 
 Consumers should look to follow the processs below:
 
-1. Create `Config` using `ConfigBuilder`
-2. Create `Scanner` using `Config` created in step 1.
-3. Run `Scan` method on `Scanner` object.
+1. Create a `Config` object using the `ConfigBuilder` class
+2. Create a `Scanner` object using the `Config` object created in step 1.
+3. Run the `Scan` method on the `Scanner` object.
 4. Analyze the returned `ScanResults` object for your respective needs.
 
 The details of these objects/methods are below.
@@ -94,6 +94,24 @@ The **Errors** property contains **ScanResult** objects which are the result of 
 Rule | `RuleInfo` | Information about the rule (description, how to fix information, etc.) that was evaluated on the element.
 Element | `ElementInfo` | The element which was tested against the rule.
 
+`RuleInfo` contains the following properties:
+
+**Name** | **Type** | **Description**
+---|---|---
+ID | `RuleId` | Contains a unique identifier for the rule from the RuleId enumeration.
+Description | `string` | Contains a short description of the rule.
+HowToFix | `string` | Detailed information on how to resolve a violation reported by the rule.
+Standard | `A11yCriteriaId` | An enum which identifies the standards documentation from which the rule was derived.
+PropertyID | `int` | In cases where the rule tests one specific UI Automation property, this contains the UI Automation property ID in question. This property is used to link elements with rule violations to relevant documentation.
+Condition | 'string' | A description of the conditions under which a rule will be evaluated.
+
+`ElementInfo` contains the following properties:
+
+**Name** | **Type** | **Description**
+---|---|---
+Properties | `Dictionary<string, string>` | A string to string dictionary where the key is a UIAutomation property name and the value is the corresponding UI Automation property value.
+Patterns | `IEnumerable<string>` | A list of names of supported patterns.
+
 ### Using the assembly
 You can get the files via a NuGet package Configure NuGet to retrieve the
 **Axe.Windows** package from
@@ -101,8 +119,7 @@ You can get the files via a NuGet package Configure NuGet to retrieve the
 then use the classes in the Axe.Windows.Automation namespace (see
 example below):
 
--   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by
-    Accessibility Insights).
+-   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by Axe-Windows).
 -   If you’re using NuGet, add the appropriate feed to your project.
 -   Add **using Axe.Windows.Automation;** to your code.
 -   Follow the steps in **How To Use**.
@@ -122,19 +139,17 @@ example below):
             /// </summary>
             static void Main(string[] args)
             {
-                string TestAppPath = Path.GetFullPath("../myApplication.exe");
-                string OutputDir = Path.GetFullPath("./TestOutput");
-                Process TestProcess = Process.Start(TestAppPath);
-                var config = Config.Builder.ForProcessId(TestProcess.Id)
-                    .WithOutputDirectory(OutputDir)
+                string testAppPath = Path.GetFullPath("..\myApplication.exe");
+                string outputDir = Path.GetFullPath(".\TestOutput");
+                Process testProcess = Process.Start(testAppPath);
+                var config = Config.Builder.ForProcessId(testProcess.Id)
+                    .WithOutputDirectory(outputDir)
                     .WithOutputFileFormat(OutputFileFormat.A11yTest)
                     .Build();
 
                 var scanner = ScannerFactory.CreateScanner(config);
 
                 var output = scanner.Scan();
-
-                var regexForExpectedFile = $"{OutputDir.Replace("\\", "\\\\")}.*\\.a11ytest";
 
                 Assert.IsTrue(output.ErrorCount == 0);
             }
--- a/docs/Automation.md
+++ b/docs/Automation.md
@@ -15,7 +15,7 @@ AxeWindows functionality to automation systems.
         myConfigBuilder.WithOutputFileFormat(OutputFileFormat.A11yTest);
 
         // Optional: configure to output the file to a specific directory (otherwise, current directory will be used)
-        myConfigBuilder.WithOutputFileFormat(OutputFileFormat.A11yTest);
+        myConfigBuilder.WithOutputDirectory(".\test-directory");
 
         // Ready to use config
         var myConfig = myConfigBuilder.build();
@@ -47,7 +47,7 @@ A [complete code example](#example) can be found below.
 
 #### Config.Builder
 
-##### ForProcessId (static)
+##### *static* ForProcessId
 Create the builder for the config for the specified process.
 
 ##### Parameters
@@ -66,7 +66,7 @@ The **ForProcessId** method returns an instance of **Config.Builder**.
 
 Specify the type(s) of output files you wish AxeWindows to create.
 
-The [`OutputFileFormat` enum](../src/Automation/enums/OutputFileFormat.cs), currently has the following possible values:
+The [`OutputFileFormat` enum](../src/Automation/enums/OutputFileFormat.cs) currently has the following possible values:
 
 **Name** | **Value** | **Description**
 ---|---|---
@@ -79,7 +79,7 @@ The **WithOutputFileFormat** method accepts the following parameters:
 
 **Name** | **Type** | **Description**
 ---|---|---
-format | `OutputFileFormat` | The type(s) of output files you wish AxeWindows to create. No output files will be created if this is left unspecified. The default value is `None`.
+format | `OutputFileFormat` | The type(s) of output files you wish AxeWindows to create. No output files will be created if this is left unspecified or 0 errors are found. The default value is `None`.
 
 ##### Return object
 
@@ -165,7 +165,7 @@ Description | `string` | Contains a short description of the rule.
 HowToFix | `string` | Detailed information on how to resolve a violation reported by the rule.
 Standard | `A11yCriteriaId` | An enum which identifies the standards documentation from which the rule was derived.
 PropertyID | `int` | In cases where the rule tests one specific UI Automation property, this contains the UI Automation property ID in question. This property is used to link elements with rule violations to relevant documentation.
-Condition | 'string' | A description of the conditions under which a rule will be evaluated.
+Condition | `string` | A description of the conditions under which a rule will be evaluated.
 
 `ElementInfo` contains the following properties:
 
@@ -175,7 +175,7 @@ Properties | `Dictionary<string, string>` | A string to string dictionary where 
 Patterns | `IEnumerable<string>` | A list of names of supported patterns.
 
 ### Using the assembly
-You can get the files via a NuGet package Configure NuGet to retrieve the
+You can get the files via a NuGet package; configure NuGet to retrieve the
 **Axe.Windows** package from
 <https://api.nuget.org/v3/index.json>,
 then use the classes in the Axe.Windows.Automation namespace (see
@@ -184,7 +184,7 @@ example below):
 -   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by Axe-Windows).
 -   If you’re using NuGet, add the appropriate feed to your project.
 -   Add **using Axe.Windows.Automation;** to your code.
--   Follow the steps in **How To Use**.
+-   Follow the steps in [How To Run An Automated Scan](#how-to-run-an-automated-scan).
 
 #### Example
 ```
@@ -211,13 +211,15 @@ example below):
 
                 var scanner = ScannerFactory.CreateScanner(config);
 
-                try {
+                try
+                {
                     var output = scanner.Scan();
                     Assert.IsTrue(output.ErrorCount == 0);
                 }
-                catch(AxeWindowsAutomationException e) {
+                catch(AxeWindowsAutomationException e)
+                {
                     var errorMessage = e.InnerException != null ? e.InnerException.Message : e.Message;
-                    Assert.IsTrue(false, errorMessage);
+                    Assert.Fail(errorMessage);
                 }
             }
         }
--- a/docs/Automation.md
+++ b/docs/Automation.md
@@ -3,12 +3,7 @@
 ### Overview
 To enable various automation scenarios, we have created an assembly
 (`Axe.Windows.Automation.dll`) that exposes a subset of core
-AxeWindows functionality to automation systems. This can be used in
-a variety of ways, including as part of a standalone test system, 
-PowerShell script, or CI/CD solution. One or more scans can be
-performed during a test run and the outputs will be saved to disk. No UI is
-provided in automation mode, and the code is intended to be compatible with
-Windows 7 or newer.
+AxeWindows functionality to automation systems.
 
 ### General Characteristics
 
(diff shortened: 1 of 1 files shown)


docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
+#### How To Use (.NET)
 
-**Name** | **Type** | **Description**
----|---|---
-primaryConfig | `Dictionary<string, string>` | The set of Tier 2 parameters to apply during this session. See [Defined Constants](#defined-constants) for specific parameters.
-configFile | `string` | The full path to a file that contains JSON-serialized Tier 3 parameters. Can be `string.Empty`. Must point to a valid JSON-serialized file if non-empty.
+Consumers should look to follow the processs below:
 
-###### PowerShell
+1. Create `Config` using `ConfigBuilder`
```
DaveTryon: Let's flesh out the grammar:
```suggestion
1. Create a `Config` object using the `ConfigBuilder` class
```

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
+2. Create `Scanner` using `Config` created in step 1.
```
DaveTryon: Let's flesh out the grammar:
```suggestion
2. Create a `Scanner` object using the `Config` object created in step 1.
```

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
+3. Run `Scan` method on `Scanner` object.
```
DaveTryon: Let's flesh out the grammar:
```suggestion
3. Run the `Scan` method on the `Scanner` object.
```

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
-session is open (i.e., between the Start and Stop commands):
+##### Return object
 
-##### Inputs
+The **ForProcessId** method returns a **Config.Builder** object, which has the following methods:
 
-###### .NET
+**Name** | **Parameters** | **Output** | **Description**
+---|---|---|---
+WithOutputFileFormat | `(OutputFileFormat format)` | `Config.Builder` | Specify the type(s) of output files you wish AxeWindows to create. No output files will be created if this is left unspecified.
```
DaveTryon: >OutputFileFormat [](start = 25, length = 16)

Suggestion: Document this enum and make this be a link to the enum. The enum will only have 1 value for now, but we'll soon be adding a second value

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
-CommandConstStrings.NoViolationPolicy | “NoViolationPolicy” | Specifies how output files should be handled if no violations are found. <ul><li>**Discard** will discard the results file.<li>**Retain** will retain the results file.</ul>*Default value = **Retain***. Valid only as a Tier 3 setting.
-CommandConstStrings.Discard  | “Discard” | Value for NoViolationPolicy.
-CommandConstStrings.Retain | “Retain” | Value for NoViolationPolicy.
-CommandConstStrings.TeamName | “TeamName” | Value for Team name in telemetry.
+Rule | `RuleInfo` | Information about the rule (description, how to fix information, etc.) that was evaluated on the element.
```
DaveTryon: >RuleInfo [](start = 8, length = 8)

We should document what this is

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
+Element | `ElementInfo` | The element which was tested against the rule.
```
DaveTryon: >ElementInfo [](start = 11, length = 11)

We should document what this is

docs/Automation.md
```diff
@@ -188,78 +122,22 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
-                        { CommandConstStrings.OutputFile, autoFileId++.ToString() },
-                    };
-                    Console.WriteLine(SnapshotCommand.Execute(snapshotParameters).ToString());
-                }
-                Console.WriteLine(StopCommand.Execute().ToString());
-            }
-        }
-    }
-```
+                string TestAppPath = Path.GetFullPath("../myApplication.exe");
```
DaveTryon: >/ [](start = 57, length = 1)

We generally use backslash (.NET framework cross-platform support is very limited)

DaveTryon: >T [](start = 23, length = 1)

Stylistically, local variables generally start with a lower case letter

docs/Automation.md
```diff
@@ -188,78 +122,22 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
-    Start-Sleep 2
-    $appProcId=get-process notepad |select -expand id
-    Invoke-Snapshot -OutputFile notepadSnapshot -TargetProcessId $appProcId
-    Stop-Process -Id $appProcId
-    Stop-AxeWindows
-```
+                var output = scanner.Scan();
 
-**Note**: This script will work only if the instance of notepad.exe started is the only instance in the entire system. If there is more than one instance, the PowerShell layer will throw an Exception because it tries to convert something like "123 456" to a number, which will fail.
+                var regexForExpectedFile = $"{OutputDir.Replace("\\", "\\\\")}.*\\.a11ytest";
```
DaveTryon: >regexForExpectedFile [](start = 20, length = 20)

This seems to be unused ?

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
 You can get the files via a NuGet package Configure NuGet to retrieve the
 **Axe.Windows** package from
 <https://api.nuget.org/v3/index.json>,
 then use the classes in the Axe.Windows.Automation namespace (see
-examples below):
+example below):
 
-#### From .NET code
 -   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by
     Accessibility Insights).
```
DaveTryon: Axe-Windows

docs/Automation.md
```diff
@@ -20,164 +15,103 @@ and is not expected to change at any future time. If you have a scenario that
...
 ### Using the assembly
 You can get the files via a NuGet package Configure NuGet to retrieve the
 **Axe.Windows** package from
 <https://api.nuget.org/v3/index.json>,
 then use the classes in the Axe.Windows.Automation namespace (see
-examples below):
+example below):
 
-#### From .NET code
 -   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by
```
DaveTryon: >4.7.1 [](start = 47, length = 5)

Long-term, we might want to consider allowing earlier versions (the 4.7.1 is to get a fix that AIWin needed)

docs/Automation.md
```diff
@@ -188,78 +122,22 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
             /// </summary>
             static void Main(string[] args)
             {
```
DaveTryon: >{ [](start = 12, length = 1)

Should we add a try/catch handler to show the AxeWindowsAutomationException type?

waabid: I'd like to briefly sync on this offline just to make sure I have this right.

docs/Automation.md
waabid: In the comments for this, it points to AIWin, should we also link to it?

RobGallo: Yes

docs/Automation.md
RobGallo: Perhaps, but let's leave it until there's a compelling reason to downgrade.

docs/Automation.md
waabid: Added

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
 
-**Name** | **Type** | **Description**
----|---|---
-primaryConfig | `Dictionary<string, string>` | The set of Tier 2 parameters to apply during this session. See [Defined Constants](#defined-constants) for specific parameters.
-configFile | `string` | The full path to a file that contains JSON-serialized Tier 3 parameters. Can be `string.Empty`. Must point to a valid JSON-serialized file if non-empty.
+        // Optional: configure to create an A11yTest file
+        myConfigBuilder.WithOutputFileFormat(OutputFileFormat.A11yTest);
+
+        // Optional: configure to output the file to a specific directory (otherwise, current directory will be used)
+        myConfigBuilder.WithOutputFileFormat(OutputFileFormat.A11yTest);
```
karanbirsingh: Line 18 and 15 both reference `WithOutputFileFormat`, but it seems the second one should instead use `WithOutputDirectory`

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
-The **Start-AxeWindows** cmdlet accepts the following optional
-parameters:
+
+A [complete code example](#example) can be found below.
+
+### Class Details
+
+#### Config.Builder
+
+##### ForProcessId (static)
```
karanbirsingh: Not a blocker - I thought `static` was an incomplete parameter definition at first because of the parentheses

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
+##### WithOutputFileFormat
 
-#### Snapshot Command
-The Snapshot command scans a single process for accessibility issues and saves
-the output file for later analysis. It can only be called while an automation
-session is open (i.e., between the Start and Stop commands):
+Specify the type(s) of output files you wish AxeWindows to create.
 
-##### Inputs
+The [`OutputFileFormat` enum](../src/Automation/enums/OutputFileFormat.cs), currently has the following possible values:
```
karanbirsingh: I wonder if we need this comma - can we remove?

docs/Automation.md
waabid: Do you think maybe

`Static: ForProcessId` works better here?

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
+`RuleInfo` contains the following properties:
+
+**Name** | **Type** | **Description**
+---|---|---
+ID | `RuleId` | Contains a unique identifier for the rule from the RuleId enumeration.
+Description | `string` | Contains a short description of the rule.
+HowToFix | `string` | Detailed information on how to resolve a violation reported by the rule.
+Standard | `A11yCriteriaId` | An enum which identifies the standards documentation from which the rule was derived.
+PropertyID | `int` | In cases where the rule tests one specific UI Automation property, this contains the UI Automation property ID in question. This property is used to link elements with rule violations to relevant documentation.
+Condition | 'string' | A description of the conditions under which a rule will be evaluated.
```
karanbirsingh: `string` here is using single quotes instead of backticks so the text does not look monotyped

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
-CommandConstStrings.NoViolationPolicy | “NoViolationPolicy” | Specifies how output files should be handled if no violations are found. <ul><li>**Discard** will discard the results file.<li>**Retain** will retain the results file.</ul>*Default value = **Retain***. Valid only as a Tier 3 setting.
-CommandConstStrings.Discard  | “Discard” | Value for NoViolationPolicy.
-CommandConstStrings.Retain | “Retain” | Value for NoViolationPolicy.
-CommandConstStrings.TeamName | “TeamName” | Value for Team name in telemetry.
+Properties | `Dictionary<string, string>` | A string to string dictionary where the key is a UIAutomation property name and the value is the corresponding UI Automation property value.
+Patterns | `IEnumerable<string>` | A list of names of supported patterns.
 
 ### Using the assembly
 You can get the files via a NuGet package Configure NuGet to retrieve the
```
karanbirsingh: A line/sentence break would be helpful ("You can get the files via a NuGet package Configure NuGet"). This wasn't from your PR but it would be nice to fix while we're here

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
-#### From .NET code
--   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by
-    Accessibility Insights).
+-   Prerequisite: Your project *must* use .NET 4.7.1 (this is required by Axe-Windows).
 -   If you’re using NuGet, add the appropriate feed to your project.
 -   Add **using Axe.Windows.Automation;** to your code.
--   Access the Start command via the **StartCommand.Execute** method.
--   Access the Snapshot command via the **SnapshotCommand.Execute** method.
--   Access the Stop command via the **StopCommand.Execute** method.
+-   Follow the steps in **How To Use**.
```
karanbirsingh: If these how-to-use steps are elsewhere in our docs can we link to them like you linked to the code sample?

docs/Automation.md
```diff
@@ -188,78 +201,39 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
+
+                var scanner = ScannerFactory.CreateScanner(config);
+
+                try {
+                    var output = scanner.Scan();
+                    Assert.IsTrue(output.ErrorCount == 0);
+                }
+                catch(AxeWindowsAutomationException e) {
+                    var errorMessage = e.InnerException != null ? e.InnerException.Message : e.Message;
+                    Assert.IsTrue(false, errorMessage);
```
karanbirsingh: `Assert.Fail` is more canonical here
```suggestion
                    Assert.Fail(errorMessage);
```

docs/Automation.md
```diff
@@ -188,78 +201,39 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
+                string outputDir = Path.GetFullPath(".\TestOutput");
+                Process testProcess = Process.Start(testAppPath);
+                var config = Config.Builder.ForProcessId(testProcess.Id)
+                    .WithOutputFileFormat(OutputFileFormat.A11yTest)
+                    .WithOutputDirectory(outputDir)
+                    .Build();
+
+                var scanner = ScannerFactory.CreateScanner(config);
+
+                try {
```
karanbirsingh: We should put opening brackets on new lines to be consistent with the rest of the code sample

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
 
-The **StartCommandObject** has the following properties:
+The **ForProcessId** method returns an instance of **Config.Builder**.
 
-**Name** | **Type** | **Description**
----|---|---
-Completed | `bool` | True if the command ran to completion. If true, then the Succeeded property indicates if the call succeeded.
-SummaryMessage | `string` | A human-readable message to summarize the result. This message can change at any time, so do not rely on parsing it in tools.
-Succeeded | `bool` | True if the Start command completed successfully.
+##### WithOutputFileFormat
```
karanbirsingh: If it's not possible that's alright, but it would be nice for the methods to be larger/bolder than their attributes (parameter, return, etc)

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
 
-The **SnapshotCommand.Execute** method accepts the following parameters:
+##### Parameters
+
+The **WithOutputFileFormat** method accepts the following parameters:
 
 **Name** | **Type** | **Description**
 ---|---|---
-primaryConfig | `Dictionary<string, string>` | The set of Tier 1 parameters to apply during this call. See [Defined Constants](#defined-constants) for specific parameters.
+format | `OutputFileFormat` | The type(s) of output files you wish AxeWindows to create. No output files will be created if this is left unspecified. The default value is `None`.
```
karanbirsingh: I think Axe.Windows does not serialize an a11ytest file unless there were errors. We should reflect this in the docs as well

docs/Automation.md
karanbirsingh: I think so, some other options include italics for the word static, or full method signatures such as `static Builder ForProcessId(int processId)`. I slightly prefer full signatures but am not too opinionated about it

docs/Automation.md
karanbirsingh: I should have commented in both places, this is also at the top in step 3 of 'how to run...'

docs/Automation.md
```diff
@@ -1,183 +1,196 @@
...
-The **StartCommand.Execute** method accepts the following parameters:
+        // Create config to specifically target a process
+        var myConfigBuilder = Config.Builder.ForProcessId(1234);
 
-**Name** | **Type** | **Description**
----|---|---
-primaryConfig | `Dictionary<string, string>` | The set of Tier 2 parameters to apply during this session. See [Defined Constants](#defined-constants) for specific parameters.
-configFile | `string` | The full path to a file that contains JSON-serialized Tier 3 parameters. Can be `string.Empty`. Must point to a valid JSON-serialized file if non-empty.
+        // Optional: configure to create an A11yTest file
+        myConfigBuilder.WithOutputFileFormat(OutputFileFormat.A11yTest);
```
karanbirsingh: The docs are accurate, although I was surprised this code works without chaining the calls or saving the return value - I incorrectly assumed the `With` methods return a new builder. It might just be me :)

docs/Automation.md
```diff
@@ -188,78 +203,41 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
             /// </summary>
             static void Main(string[] args)
             {
-                var parameters = new Dictionary<string, string>();
-                string secondaryConfigFile = string.Empty;
-                char[] delimiters = {'='};
-                foreach (string arg in args)
+                string testAppPath = Path.GetFullPath("..\myApplication.exe");
```
DaveTryon: C# will treat this as an invalid escape sequence
```suggestion
                string testAppPath = Path.GetFullPath(@"..\myApplication.exe");
```

docs/Automation.md
```diff
@@ -188,78 +203,41 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
+                string outputDir = Path.GetFullPath(".\TestOutput");
```
DaveTryon: C# will treat this as an invalid escape sequence
```suggestion
                string outputDir = Path.GetFullPath(@".\TestOutput");
```

docs/Automation.md
```diff
@@ -1,183 +1,198 @@
...
+The `ForProcessId` method returns an instance of `Config.Builder`.
 
-**Name** | **Type** | **Description**
+##### `WithOutputFileFormat`
+
+Specify the type(s) of output files you wish AxeWindows to create.
+
+The [`OutputFileFormat` enum](../src/Automation/enums/OutputFileFormat.cs) currently has the following possible values:
+
+**Name** | **Value** | **Description**
```
DaveTryon: >**Value** [](start = 11, length = 9)

Not a fan of including the Value column. The user code should always be in terms of the enum, and anyone who wants to know the value can quickly determine what it is.

docs/Automation.md
```diff
@@ -1,183 +1,198 @@
...
-The **Start-AxeWindows** cmdlet accepts the following optional
-parameters:
+The `ForProcessId` method returns an instance of `Config.Builder`.
 
-**Name** | **Type** | **Description**
+##### `WithOutputFileFormat`
+
+Specify the type(s) of output files you wish AxeWindows to create.
+
+The [`OutputFileFormat` enum](../src/Automation/enums/OutputFileFormat.cs) currently has the following possible values:
```
DaveTryon: C# enums define values but are non-exclusive. The following is valid C# code: OutputFileFormat x = (OutputFileFormat) 123456; We can better express this with a slight reword:
```suggestion
The [`OutputFileFormat` enum](../src/Automation/enums/OutputFileFormat.cs) currently defines the following values:
```

docs/Automation.md
```diff
@@ -1,183 +1,198 @@
...
+The `Scan` method returns a `ScanResults` object and has the following properties:
 
 **Name** | **Type** | **Description**
 ---|---|---
-Completed | `bool` | True if the command ran to completion. If true, then the Succeeded property indicates if the call succeeded.
-SummaryMessage | `string` | A human-readable message to summarize the result. This message can change at any time, so do not rely on parsing it in tools.
-Succeeded | `bool` | True if the Start command completed successfully.
+ErrorCount | `int` | A count of all errors across all elements scanned.
+Errors | `IEnumerable<ScanResult>` | A collection of errors found during the scan.
+OutputFile | `(string A11yTest, string Sarif)` | A Tuple with paths to any output files written as a result of a scan. Tuple members are A11yTest and Sarif (**not implemented yet**).
```
DaveTryon: >Tuple [](start = 51, length = 5)

We should discuss this. The Tuple was changed in [https://github.com/microsoft/axe-windows/pull/91](https://github.com/microsoft/axe-windows/pull/91). My recommendation is to document OutputFile as a struct with fields called A11yTest and Sarif. The Tuple then becomes an implementation detail independent of the consuming code.

docs/Automation.md
```diff
@@ -1,183 +1,198 @@
...
+---|---|---
+Rule | `RuleInfo` | Information about the rule (description, how to fix information, etc.) that was evaluated on the element.
+Element | `ElementInfo` | The element which was tested against the rule.
 
-**Name** | **Value** | **Description**
+`RuleInfo` contains the following properties:
+
+**Name** | **Type** | **Description**
+---|---|---
+ID | `RuleId` | Contains a unique identifier for the rule from the RuleId enumeration.
```
DaveTryon: >enumeration [](start = 74, length = 11)

This provides no information about the RuleId enumeration. Maybe a pointer to the class in GitHub?

docs/Automation.md
```diff
@@ -1,183 +1,198 @@
...
+Description | `string` | Contains a short description of the rule.
+HowToFix | `string` | Detailed information on how to resolve a violation reported by the rule.
+Standard | `A11yCriteriaId` | An enum which identifies the standards documentation from which the rule was derived.
```
DaveTryon: >enum [](start = 33, length = 4)

This provides no information about the A11yCriteriaId enumeration. Maybe a pointer to the class in GitHub?

docs/Automation.md
```diff
@@ -1,183 +1,198 @@
...
-CommandConstStrings.NoViolationPolicy | “NoViolationPolicy” | Specifies how output files should be handled if no violations are found. <ul><li>**Discard** will discard the results file.<li>**Retain** will retain the results file.</ul>*Default value = **Retain***. Valid only as a Tier 3 setting.
-CommandConstStrings.Discard  | “Discard” | Value for NoViolationPolicy.
-CommandConstStrings.Retain | “Retain” | Value for NoViolationPolicy.
-CommandConstStrings.TeamName | “TeamName” | Value for Team name in telemetry.
+Properties | `Dictionary<string, string>` | A string to string dictionary where the key is a UIAutomation property name and the value is the corresponding UI Automation property value.
```
DaveTryon: We've been trying to use "UI Automation" instead of "UIAutomation"
```suggestion
Properties | `Dictionary<string, string>` | A string to string dictionary where the key is a UI Automation property name and the value is the corresponding UI Automation property value.
```

docs/Automation.md
```diff
@@ -188,78 +210,41 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
-                            continue;
-                        }
-                    }
-                    Console.WriteLine("Ignoring malformed input: {0}", arg);
-                };
-                Console.WriteLine(StartCommand.Execute(parameters secondaryConfigFile).ToString());
-                int autoFileId = 0;
-                while (true)
+                    var output = scanner.Scan();
+                    Assert.IsTrue(output.ErrorCount == 0);
```
karanbirsingh: `Assert.AreEqual(0, output.ErrorCount)` might be a bit nicer. That way the log will tell us what the error count was if it fails
```suggestion
                    Assert.AreEqual(0, output.ErrorCount);
```

docs/Automation.md
```diff
@@ -1,183 +1,205 @@
...
+        }
+        catch(AxeWindowsAutomationException e)
+        {
+            // Get the message from an exception, if one is thrown.
+            var errorMessage = e.InnerException != null ? e.InnerException.Message : e.Message;
+        }
+
+4. Check the results.
+
+        Console.WriteLine("Number of errors found in scan: " + scanResults.errorCount);
```
karanbirsingh: The property here is likely `ErrorCount` instead of `errorCount`

docs/Automation.md
```diff
@@ -1,183 +1,205 @@
...
+Errors | `IEnumerable<ScanResult>` | A collection of errors found during the scan.
+OutputFile | `OutputFile` | Represents the output file(s), if any, associated with a ScanResults object.
 
-###### PowerShell
+The `OutputFile` property is a struct with the following properties:
 
-The **Stop-AxeWindows** cmdlet accepts no parameters
+**Name** | **Type** | **Description**
+---|---|---
+A11yTest | `string` | The A11yTest file that was generated (or null if no file was generated).
```
karanbirsingh: Not necessary, but "path to the A11yTest file" might be more clear

docs/Automation.md
```diff
@@ -1,183 +1,205 @@
...
+
+        var scanResults;
+        try
+        {
+            scanResults = scanner.Scan();
+        }
+        catch(AxeWindowsAutomationException e)
+        {
+            // Get the message from an exception, if one is thrown.
+            var errorMessage = e.InnerException != null ? e.InnerException.Message : e.Message;
```
DaveTryon: There's at least 1 place (Line 23 of TargetElementLocator.cs) where this might return a less-than-helpful message. Would it make sense to just ToString() the whole Exception?
```suggestion
            Console.WriteLine(e.ToString());
```

docs/Automation.md
```diff
@@ -188,78 +210,41 @@ Sample `C#` code—this is interactive, but yours doesn’t need to be:
...
-                        Console.WriteLine("Not a valid int: " + input);
-                        continue;
-                    }
-                    Dictionary<string, string> snapshotParameters = new Dictionary<string, string>
-                    {
-                        { CommandConstStrings.TargetProcessId, input },
-                        { CommandConstStrings.OutputFile, autoFileId++.ToString() },
-                    };
-                    Console.WriteLine(SnapshotCommand.Execute(snapshotParameters).ToString());
+                    var errorMessage = e.InnerException != null ? e.InnerException.Message : e.Message;
```
DaveTryon: There's at least 1 place (Line 23 of TargetElementLocator.cs) where this might return a less-than-helpful message. Would it make sense to just ToString() the whole Exception? That automatically walks the Exception chain, displaying messages, stack traces, etc.
```suggestion
                    var errorMessage = e.ToString();
```

PR #91, opened 2019-07-02
Don't use named Tuple for ScanResults.OutputFile

DaveTryon:
#### Describe the change
I came across an environment that provides only a partial implementation of .NET, and that implementation doesn't include the named Tuple struct. To enable the scan, I changed ScanResults.OutputFile from a Tuple to a struct? It's a struct (instead of a class) for compatibility with the named Tuple, which is a value type, not a reference type. We can certainly make it a class, but it's a potentially breaking change from the 0.2.0 interface, where ScanResults.OutputFile could never be null

### DIFF
--- a/src/Automation/Data/ScanResults.cs
+++ b/src/Automation/Data/ScanResults.cs
@@ -1,7 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using Axe.Windows.Automation.Data;
-using System;
 using System.Collections.Generic;
 
 namespace Axe.Windows.Automation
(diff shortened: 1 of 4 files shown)


src/Automation/Data/ScanResults.cs
```diff
@@ -11,16 +12,16 @@ namespace Axe.Windows.Automation
     public class ScanResults
     {
         /// <summary>
-        /// A Tuple with paths to any output files written as a result of a scan.
-        /// Tuple members are A11yTest and Sarif (not implemented yet)
+        /// An OutputFile object containing paths to any output files written as
```
RobGallo: Could you include the <See> tag for symbols like OutputFile?

PR #96, opened 2019-07-09
Fix heading title case

RobGallo:
#### Describe the change

Microsoft documentation says to have headings use sentence case


docs/RulesOverview.md
```diff
@@ -1,4 +1,7 @@
-## Rules Overview
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
```
peterdur: Not changed by this PR, but noticed that at line 23 in this file, Inheritence should be spelled Inheritance

PR #97, opened 2019-07-10
Add release notes page to docs

RobGallo:
#### Describe the change

Add a release notes page which covers changes related to AxeWindows automation. Changes related to other parts of Axe.Windows have not been included.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,0 +1,9 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+
+## Release notes
+
+### 0.3.0
+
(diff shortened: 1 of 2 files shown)


docs/ReleaseNotes.md
```diff
@@ -0,0 +1,9 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+
+## Release notes
```
DaveTryon: Based on the readme, this is supposed to be "details about the latest release", but the initial formatting feels more like it will include the history of multiple releases. If we're trying to have per-version release notes, perhaps using the nuSpec file's releaseNotes (see [https://docs.microsoft.com/en-us/nuget/reference/nuspec#releasenotes](https://docs.microsoft.com/en-us/nuget/reference/nuspec#releasenotes)) to actually put this into the package is a cleaner option.

RobGallo:
This is not how we plan to do release notes.

Issue #27, opened 2019-05-13
[BUG] EditSupportsOnlyValuePattern should return an error


karanbirsingh:
[Migrated from ai-win](https://github.com/microsoft/accessibility-insights-windows/issues/287), see full detail there

> The UIA spec (https://docs.microsoft.com/en-us/windows/desktop/winauto/uiauto-supporteditcontroltype) requires every edit control to support at least the Text control pattern. The Value pattern is supplemental. This rule should return an error.

> As I understand it, the rule fires when an edit control supports the value pattern but does not support the text pattern. The documentation says ITextProvider is required, while IValueProvider depends. The only thing wrong with the rule is how the outcome is reported. I believe failure instances should be shown as errors instead of uncertain.

Sirisha-Kommushetty:
@RobGallo - Please comment and assign back to me

Sirisha-Kommushetty:
Please update the release notes for Axe.Windows to callout the rule changes for reference.

DaveTryon:
@karanbirsingh, can you please close this out? Thanks!

PR #103, opened 2019-07-11
Fix issue #27 - EditSupportsOnlyValuePattern should return an error

RobGallo:
#### Describe the change

- Update ControlShouldSupportTextPattern to always return an error for edit fields which do not support the text pattern based on the change to the Windows UIA documentation here: https://github.com/MicrosoftDocs/win32-pr/pull/123
- Remove rule EditSupportsOnlyValuePattern because it is no longer informative after the above change


src/Core/Enums/RuleId.cs
```diff
@@ -111,7 +113,7 @@ public enum RuleId
         ControlShouldNotSupportTablePattern,
         ControlShouldNotSupportTogglePattern,
         ControlShouldNotSupportValuePattern,
-        ControlShouldNotSupportWindowPattern,
+ControlShouldNotSupportWindowPattern,
```
DaveTryon: Indent changed (presumably unintentionally)
```suggestion
        ControlShouldNotSupportWindowPattern,
```

src/Core/Enums/RuleId.cs
```diff
@@ -129,7 +131,10 @@ public enum RuleId
 
         // Axe.Windows.Rules
         EditSupportsIncorrectRangeValuePattern,
+
+        [Obsolete("This rule has been depricated", true)]
```
DaveTryon: Typo
```suggestion
        [Obsolete("This rule has been deprecated", true)]
```

DaveTryon:
Do we want this for axe.windows 0.3.0?

PR #108, opened 2019-07-15
Create SystemAbstractions project

RobGallo:
#### Describe the change

- Enable other projects to use the system abstractions formerly limited to Axe.Windows.Automation
	- Allow System calls to be mocked so Fakes can be replaced
- Use .Net Standard to ensure broadest future compatibility

### DIFF
--- a/build/NetStandardRelease.targets
+++ b/build/NetStandardRelease.targets
@@ -1,7 +1,7 @@
 ﻿<Project>
 
   <PropertyGroup>
-  <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
 </PropertyGroup>
 
   <ItemGroup>
--- a/build/NetStandardRelease.targets
+++ b/build/NetStandardRelease.targets
@@ -1,0 +1,26 @@
+﻿<Project>
+
+  <PropertyGroup>
+  <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+</PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="$(TEMP)\AxeWindowsVersionInfo.cs" Link="AxeWindowsVersionInfo.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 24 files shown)


src/SystemAbstractions/Abstractions/ISystem.cs
```diff
@@ -2,9 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System;
 
-namespace Axe.Windows.Abstractions
+namespace SystemAbstractions
```
DaveTryon: >SystemAbstractions [](start = 10, length = 18)

I'd argue that this namespace belongs under Axe.Windows, so Axe.Windows.SystemAbstractions. That's also the name of the assembly, so it's consistent with the pattern that we follow elsewhere.

build/NetStandardRelease.targets
```diff
@@ -0,0 +1,26 @@
+﻿<Project>
+
+  <PropertyGroup>
+  <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```
DaveTryon: indents
```suggestion
    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```

build/NetStandardRelease.targets
```diff
@@ -0,0 +1,26 @@
...
+</PropertyGroup>
```
DaveTryon: indents
```suggestion
  </PropertyGroup>
```

build/NetStandardTest.targets
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+<PropertyGroup>
```
DaveTryon: indents
```suggestion
  <PropertyGroup>
```

build/NetStandardTest.targets
```diff
@@ -0,0 +1,10 @@
...
+    <NoWarn>1701;1702;CA1707</NoWarn>
+  </PropertyGroup>
+
+<Import Project=".\NetStandardAll.targets" />
```
DaveTryon: indents
```suggestion
  <Import Project=".\NetStandardAll.targets" />
```

PR #130, opened 2019-08-06
Update all projects to include Roslyn analyzers 2.9.4

RobGallo:
#### Describe the change

Types of warnings:
- Improper use of ArgumentException
- Improper use of ArgumentNullException
- not checking for null parameters to public functions
- Quoted strings, which should be placed in resource files, passed to exceptions
- Missing the NeutralResourcesLanguage attribute for the assembly
- Use Any() instead of Count() (Linq extensions) because it is more efficient
- Custom exceptions not serializable
- Custom exceptions not supporting expected constructors
- Not (obviously) disposing of objects inheriting IDisposable
- Function can be made static because it modifies no members


src/Core/Misc/ExtensionMethods.cs
```diff
@@ -59,12 +64,11 @@ public static bool IsControlOrContent(this A11yElement e)
...
         {
-            if (e.Properties != null && e.Properties.ContainsKey(PropertyType.UIA_ClassNamePropertyId))
-            {
-                return e.Properties[PropertyType.UIA_ClassNamePropertyId].TextValue;
-            }
-            else
-                return null;
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (e.Properties == null) return null;
+            if (!e.Properties.ContainsKey(PropertyType.UIA_ClassNamePropertyId)) return null;
```
DaveTryon: Great place for a TryGetValue:

if (e.Properties.TryGetvalue(PropertyType.UIA_ClassNamePropertyId, out A11yProperty property))
    return property.TextValue;

return null;

src/Core/Misc/Preconditions.cs
```diff
@@ -18,7 +19,7 @@ public static void ArgumentIsNotNull(this object value, string valueName)
         {
             if (value == null)
             {
-                throw new ArgumentNullException(valueName, "Value can not be null");
+                throw new ArgumentNullException(valueName, ErrorMessages.VallueCannotBeNull);
```
DaveTryon: Typo in name
```suggestion
                throw new ArgumentNullException(valueName, ErrorMessages.ValueCannotBeNull);
```

src/Core/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,141 @@
...
+  <data name="ParameterMustNotBeNegative" xml:space="preserve">
+    <value>This parameter must not be negative</value>
+  </data>
+  <data name="PropertyAlreadySet" xml:space="preserve">
+    <value>The given property was already set.</value>
+  </data>
+  <data name="UnableToLocateTargetElementInFile" xml:space="preserve">
+    <value>Unable to locate the target element in the file.</value>
+  </data>
+  <data name="VallueCannotBeNull" xml:space="preserve">
```
DaveTryon: Typo in name
```suggestion
  <data name="ValueCannotBeNull" xml:space="preserve">
```

src/Desktop/ColorContrastAnalyzer/ColorPair.cs
```diff
@@ -65,6 +65,8 @@ public Boolean AreVisuallySimilarColors()
          */
         public Boolean IsVisiblySimilarTo(ColorPair otherPair)
         {
+            if (otherPair == null) throw new ArgumentNullException(nameof(otherPair));
```
DaveTryon: Would returning false make sense here (instead of throwing)?

src/Desktop/UIAutomation/DesktopElement.cs
```diff
@@ -101,6 +103,8 @@ public static bool IsFromCurrentProcess(IUIAutomationElement uia)
         /// <returns></returns>
         private static dynamic GetPropertyValue(IUIAutomationElement element, int id)
         {
+            if ( element == null) throw new ArgumentNullException(nameof( element));
```
DaveTryon: Extra space got introduced in the nameof operation
```suggestion
            if ( element == null) throw new ArgumentNullException(nameof(element));
```

src/Core/Misc/ExtensionMethods.cs
RobGallo: I don't disagree. However, I have _tried_ to draw the line with these changes to items found by the analyzers.

src/Desktop/UIAutomation/EventHandlers/EventListenerFactory.cs
```diff
@@ -540,10 +577,12 @@ public void UnregisterAutomationEventListener(int eventId,bool wait = false)
         /// </summary>
         public void UnregisterAllAutomationEventListners()
         {
+            #pragma warning disable CA2000 // Call IDisposable.Dispose()
```
DaveTryon: Inconsistent indents
```suggestion
#pragma warning disable CA2000 // Call IDisposable.Dispose()
```

src/Desktop/UIAutomation/EventHandlers/PropertyChangedEventListener.cs
```diff
@@ -32,7 +32,10 @@ public override void Init()
 
         public void HandlePropertyChangedEvent(IUIAutomationElement sender, int propertyId, object newValue)
         {
+#pragma warning disable CA2000 // Call IDisposable.Dispose()
             var m = EventMessage.GetInstance(this.EventId, sender);
+#pragma warning restore CA2000
+
```
DaveTryon: Remove the blank line on line 38?

src/Desktop/UIAutomation/EventHandlers/TextEditTextChangedEventListener.cs
```diff
@@ -45,13 +46,16 @@ public void HandleTextEditTextChangedEvent(IUIAutomationElement sender, TextEdit
...
                     new KeyValuePair<string, dynamic>("TextEditChangeType", type.ToString()),
                 };
-                for (int i = 0; i < array.Length; i++)
+
+                if (array != null)
                 {
-                    m.Properties.Add(new KeyValuePair<string, dynamic>(Invariant($"[{i}]"), array.GetValue(i)));
-                };
+                    for (int i = 0; i < array.Length; i++)
+                        m.Properties.Add(new KeyValuePair<string, dynamic>(Invariant($"[{i}]"), array.GetValue(i)));
```
DaveTryon: Personally, I'd make line 53 a block with curly braces. Failing to do so makes the next line jump 2 indents, which is awkward

src/Desktop/Utility/ExtensionMethods.cs
```diff
@@ -72,6 +76,8 @@ public static ProcessModule GetProcessModule(this A11yElement e)
         /// <returns></returns>
         public static List<DesktopElement> ToListOfDesktopElements( this IUIAutomationElementArray array)
```
DaveTryon: Not strictly in this PR, but there's an extra space in this line
```suggestion
        public static List<DesktopElement> ToListOfDesktopElements(this IUIAutomationElementArray array)
```

src/Desktop/UIAutomation/EventHandlers/TextEditTextChangedEventListener.cs
RobGallo: The best way to fix this problem is probably to make the code with the `for` loop its own function. I hesitate to add the braces because multiple sets of braces become confusing for us screen reader users (those who don't use Braille displays anyway).

src/RuleSelection/RuleRunner.cs
```diff
@@ -60,7 +60,6 @@ private static void Run(ScanResults results, A11yElement e)
         private static ScanResult ConvertRunResultToScanResult(RunResult runResult)
         {
             if (runResult == null) throw new ArgumentNullException(nameof(runResult));
-            if (runResult.RuleInfo == null) throw new ArgumentException(nameof(runResult.RuleInfo));
```
DaveTryon: I understand that CreateResult will throw is runResult.RuleInfo is null, but if someone later changed this method to not call CreateResult, then we're going to have NullReferenceExceptions in the code. It may be "belt and suspenders", but I'd lobby for explicitly throwing *in this method* if runResult.RuleInfo is null on input

src/Rules/Library/NameExcludesSpecialCharacters.cs
```diff
@@ -22,8 +22,8 @@ public NameExcludesSpecialCharacters()
 
         public override EvaluationCode Evaluate(IA11yElement e)
         {
-            if (e == null) throw new ArgumentException(nameof(e));
-            if (e.Name == null) throw new ArgumentException(nameof(e.Name));
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (e == null) throw new ArgumentNullException(nameof(e));
```
DaveTryon: Lines 25 and 26 are identical. I suspect that line 26 is supposed to be:
```suggestion
            if (e.name == null) throw new ArgumentException(ErrorMessages.ElementNameNullOrWhiteSpace, nameof(e));
```

src/Rules/Library/NameIsReasonableLength.cs
```diff
@@ -26,8 +26,8 @@ public NameIsReasonableLength()
 
         public override EvaluationCode Evaluate(IA11yElement e)
         {
-            if (e == null) throw new ArgumentException(nameof(e));
-            if (e.Name == null) throw new ArgumentException(nameof(e.Name));
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (e == null) throw new ArgumentNullException(nameof(e));
```
DaveTryon: Lines 29 and 30 are identical. I suspect that line 30 is supposed to be:
```suggestion
            if (e.name == null) throw new ArgumentException(ErrorMessages.ElementNameNullOrWhiteSpace, nameof(e));
```

src/Rules/Library/SiblingUniqueAndNotFocusable.cs
```diff
@@ -39,8 +39,8 @@ public SiblingUniqueAndNotFocusable()
 
         public override EvaluationCode Evaluate(IA11yElement e)
         {
-            if (e == null) throw new ArgumentException(nameof(e));
-            if (e.Parent == null) throw new ArgumentNullException(nameof(e.Parent));
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (e == null) throw new ArgumentNullException(nameof(e));
```
DaveTryon: Lines 42 and 43 are identical. I suspect line 43 is supposed to be
```suggestion
            if (e.Parent == null) throw new ArgumentException(ErrorMessages.ElementParentNull, nameof(e));
```

DaveTryon:
Too late now, but future changes like this might be cleaner if the changes for each project were in a separate PR. That also allows parallelization--one PR can be in review while another PR is being created. I'd lean toward having one branch per PR (as opposed to "stacking" them into a single branch), since there boundaries between the projects are so well-defined.

src/Rules/PropertyConditions/Relationships.cs
```diff
@@ -158,16 +158,16 @@ public static Condition AllAncestors(Condition c)
 
         private static bool HasSameTypeAsReferenceElement(IA11yElement e)
         {
-            if (e == null) throw new ArgumentException(nameof(e));
-            if (!Condition.Context.IsValueCreated) throw new Exception("Expected Condition.Context to be valid");
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (!Condition.Context.IsValueCreated) throw new InvalidCastException(ErrorMessages.ExpectedValidConditionContext);
```
DaveTryon: InvalidCastException seems like an odd choice here... but I don't necessarily know what would be a better option

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,153 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="ElementBoundingRectangleNull" xml:space="preserve">
+    <value>The element's BoundingRectangle property should not be null</value>
```
DaveTryon: "must" instead of "should" ?
```suggestion
    <value>The element's BoundingRectangle property must not be null</value>
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,153 @@
...
+  </data>
+  <data name="ElementLocalizedControlTypeNullOrWhiteSpace" xml:space="preserve">
+    <value>The element's LocalizedControlType property should not be null or white space</value>
```
DaveTryon: "must" instead of "should" ?
```suggestion
    <value>The element's LocalizedControlType property must not be null or white space</value>
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,153 @@
...
+  </data>
+  <data name="ElementNameNullOrWhiteSpace" xml:space="preserve">
+    <value>The element's Name property should not be null or white space</value>
```
DaveTryon: "must" instead of "should" ?
```suggestion
    <value>The element's Name property must not be null or white space</value>
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,153 @@
...
+  </data>
+  <data name="ElementParentNull" xml:space="preserve">
+    <value>The element's Parent property should not be null</value>
```
DaveTryon: "must" instead of "should" ?
```suggestion
    <value>The element's Parent property must not be null</value>
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,153 @@
...
+    <value>Expected T to be an enumeration type</value>
+  </data>
+  <data name="ExpectedValidAncestor" xml:space="preserve">
+    <value>Expected to find a valid ancestor element</value>
+  </data>
+  <data name="ExpectedValidConditionContext" xml:space="preserve">
+    <value>Expected Condition.Context to be valid</value>
+  </data>
+  <data name="IntParameterEqualsZero" xml:space="preserve">
+    <value>The given int parameter should not be zero</value>
```
DaveTryon: "must" instead of "should" ?
```suggestion
    <value>The given int parameter must not be zero</value>
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -0,0 +1,153 @@
...
+  </data>
+  <data name="IntParameterLessThanZero" xml:space="preserve">
+    <value>Int parameter should not be less than zero</value>
```
DaveTryon: "must" instead of "should" ?
```suggestion
    <value>Int parameter must not be less than zero</value>
```

src/Rules/Rule.cs
```diff
@@ -20,7 +20,10 @@ abstract class Rule : IRule
         protected Rule()
         {
             // keep these two calls in the following order or the RuleInfo.Condition string won't get populated
+#pragma warning disable CA2214
             this.Condition = CreateCondition();
```
DaveTryon: Based on the docs for CA2214, it seems like the current code isn't guaranteed to work in the future. We may want to consider using dependency injection in the future, and having the ctor take a Condition object (built in the derived class) as a parameter.

src/Rules/Rules.cs
```diff
@@ -12,7 +12,9 @@ namespace Axe.Windows.Rules
     /// <summary>
     /// Use this class to run rules and get information about any rule in the Rules assembly.
     /// </summary>
+    #pragma warning disable CA1724
```
DaveTryon: Inconsistent indents
```suggestion
#pragma warning disable CA1724
```

src/Telemetry/ExcludedException.cs
```diff
@@ -20,7 +20,7 @@ public class ExcludedException : Exception
         public Type ExcludedType => InnerException.GetType();
 
         public ExcludedException(Exception innerException)
-            : base (innerException.Message, innerException)
+            : base (innerException?.Message, innerException)
```
DaveTryon: ExcludedException requires that innerException be non-null. If we use null propagation here, we should add an explicit check in the ctor:
```
           if (innerException == null) throw new ArgumentNullException(nameof(innerException));
```

src/UnitTestSharedLibrary/Utility.cs
```diff
@@ -58,7 +58,7 @@ public static A11yElement LoadA11yElementsFromJSON(string path)
         /// <returns></returns>
         public static void PopulateChildrenTests(A11yElement ke)
         {
-            ke.ScanResults.Items.ForEach(item => {
+            ke?.ScanResults?.Items?.ForEach(item => {
```
DaveTryon: The old code would throw NullReferenceException if ke or ke.ScanResults or ke.ScanResults.Items were null. Should be retain the throwing behavior for bad inputs, but map them to a flavor of ArgumentException?

src/Core/Misc/ExtensionMethods.cs
DaveTryon: Makes sense.

---
In reply to: [311284107](https://github.com/microsoft/axe-windows/pull/130#discussion_r311284107) [](ancestors = 311284107)

src/Desktop/ColorContrastAnalyzer/ColorPair.cs
DaveTryon: Old code would have thrown, so keeping the throwing seems reasonable

---
In reply to: [311285509](https://github.com/microsoft/axe-windows/pull/130#discussion_r311285509) [](ancestors = 311285509)

src/Rules/Library/NameExcludesSpecialCharacters.cs
```diff
@@ -22,8 +22,8 @@ public NameExcludesSpecialCharacters()
 
         public override EvaluationCode Evaluate(IA11yElement e)
         {
-            if (e == null) throw new ArgumentException(nameof(e));
-            if (e.Name == null) throw new ArgumentException(nameof(e.Name));
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (e.name == null) throw new ArgumentException(ErrorMessages.ElementNameNullOrWhiteSpace, nameof(e));
```
DaveTryon: Looks like you grabbed my first version, which had a typo:
```suggestion
            if (e.Name == null) throw new ArgumentException(ErrorMessages.ElementNameNullOrWhiteSpace, nameof(e));
```

src/RuleSelection/RuleRunner.cs
RobGallo: Fixed

src/RuleSelection/RuleSelection.csproj
```diff
@@ -15,7 +15,7 @@
     <OutputType>Library</OutputType>
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Axe.Windows.RuleSelection</RootNamespace>
-    <AssemblyName>Axe.Windows.RuleSelection</AssemblyName>
+    <AssemblyName>es</AssemblyName>
```
DaveTryon: >es [](start = 18, length = 2)

Typo?

src/Rules/PropertyConditions/Relationships.cs
RobGallo: It was an intellisense/screen reader anomaly. Fixed.

src/Telemetry/ExcludedException.cs
RobGallo: Fixed

src/Rules/Rule.cs
RobGallo: Agreed. That is a big change though. And I was trying to keep this PR limited in scope, quite obviously ineffectually.

src/RuleSelection/RuleSelection.csproj
RobGallo: Fixed

src/UnitTestSharedLibrary/Utility.cs
RobGallo: Fixed

DaveTryon: Do we want protection if ke.ScanItems == null or ke.ScanItems.Results == null?

---
In reply to: [311581411](https://github.com/microsoft/axe-windows/pull/130#discussion_r311581411) [](ancestors = 311581411)

Issue #140, opened 2019-08-22
[BUG] False positives created by change to ControlShouldSupportTextPattern rule


RobGallo:
The ControlSupportsTextPattern rule was changed to require all edit controls to support the text pattern. This is in-line with a recent change to the UIA documentation. However, there are legacy situations which the rule now reports as errors and which are unlikely to be resolved. These are perceived as false positives by users.

The first reproduceable instance given to us was the find edit field in Notepad. This edit field is built with the Win32 framework. Axe.Windows should not be reporting this error for Win32 because it is unlikely to be fixed and Win32 edit fields don’t seem to present a barrier to ATs, probably because they can be accessed using Windows messages.

The second reproduceable instance given to us was Windows Explorer, where there are many edit fields which generate this error on every page. These don't actually appear to be editable content. But the framework, DirectUI, has designated them as such. We believe it is safe to exempt edit fields which are not focusable in the DirectUI framework from the ControlShouldSupportTextPattern rule because these do not appear to be truly editable.

ControlShouldSupportTextPattern should be updated with the exemptions described above.

RobGallo:
Fixed in #147

PR #147, opened 2019-09-09
Fix [BUG] False positives created by change to ControlShouldSupportTextPattern ruleIssue #140

RobGallo:
#### Describe the change

Please see the bug description for details on the fix.

### DIFF
--- a/src/Rules/Library/ControlShouldSupportTextPattern.cs
+++ b/src/Rules/Library/ControlShouldSupportTextPattern.cs
@@ -5,6 +5,7 @@ using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
 using Axe.Windows.Rules.PropertyConditions;
 using Axe.Windows.Rules.Resources;
+using static Axe.Windows.Rules.PropertyConditions.BoolProperties;
 using static Axe.Windows.Rules.PropertyConditions.ControlType;
 
 namespace Axe.Windows.Rules.Library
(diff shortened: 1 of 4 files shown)


src/Rules/Library/ControlShouldSupportTextPattern.cs
```diff
@@ -28,7 +29,13 @@ public override EvaluationCode Evaluate(IA11yElement e)
 
         protected override Condition CreateCondition()
         {
-            return Document | Edit;
+            var win32Edit = Edit & StringProperties.Framework.Is(Framework.Win32);
+            var nonfocusableDirectUIEdit = Edit & ~IsKeyboardFocusable & StringProperties.Framework.Is(Framework.DirectUI);
+
+            return Document
+                | (Edit
+                & ~win32Edit
```
DaveTryon: >win32Edit [](start = 19, length = 9)

This executes only once per rule, right? If not, it might be worth making these static to save memory

src/RulesTest/Library/ControlShouldSupportTextPatternUnitTests.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            Assert.IsFalse(Rule.Condition.Matches(e));
+        }
+
+        [TestMethod]
+        public void DirectUIEdit_NotApplicable()
+        {
+            var e = new MockA11yElement();
+            e.ControlTypeId = ControlType.Edit;
+            e.Framework = Framework.Win32;
```
DaveTryon: >Win32 [](start = 36, length = 5)

Based on the test name, should this be DirectUI?

src/Rules/Library/ControlShouldSupportTextPattern.cs
RobGallo: The CreateCondition function is only called once per rule per session.

PR #149, opened 2019-09-17
Add backward file compatibility tests 

DaveTryon:
#### Describe the change
Add tests to ensure that old versions of Axe.Windows can open A11yTest files generated by the current build. There's a standalone test app (CurrentFileVersionCompatibilityTests.exe) that gets run after the scan_integration test completes. This test app uses Axe.Windows 0.3.1 to pin the old behavior. If the results don't match expectations, then the test fails.


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -59,6 +60,41 @@ public void Scan_Integration()
...
             Assert.IsTrue(Regex.IsMatch(output.OutputFile.A11yTest, regexForExpectedFile));
             Assert.IsTrue(File.Exists(output.OutputFile.A11yTest));
+
+            EnsureGeneratedFileIsReadableByOldVersionsOfAxeWindows(output, TestProcess.Id);
+        }
+
+        private void EnsureGeneratedFileIsReadableByOldVersionsOfAxeWindows(ScanResults scanResults, int processId)
+        {
+            string targetFolder = Path.GetFullPath(
+                Path.Combine(Directory.GetCurrentDirectory(), @"..\..\..\CurrentFileVersionCompatibilityTests\bin",
```
jalkire: Other paths in this class are pulled out as readonly at the top of the class. Might be worth considering here.

src/AxeWindows.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.757
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29215.179
```
jalkire: Did we decide to embrace VS2019?

jalkire:
Do we want the reverse of this? ie a test to ensure the current version of Axe.Windows can open an old file. Not that it would need to be a part of this PR

DaveTryon:
Those tests were added earlier. They're in the OldFileVersionCompatibilityTests project. 

---
In reply to: [532432934](https://github.com/microsoft/axe-windows/pull/149#issuecomment-532432934) [](ancestors = 532432934)

src/AxeWindows.sln
DaveTryon: I thought we had, but the docs still say 2017. Reverting...

---
In reply to: [325417068](https://github.com/microsoft/axe-windows/pull/149#discussion_r325417068) [](ancestors = 325417068)

src/AutomationTests/AutomationIntegrationTests.cs
DaveTryon: Done!

PR #154, opened 2019-09-24
Update to Roslyn Analyzers 2.9.5 (final beta)

DaveTryon:
#### Describe the change
Update to Roslyn Analyzers 2.9.5 (final beta). The final 2.9.5 release (based on the 2.9.4 beta) is probably about 2 weeks away, but it should be just an update to the project files with no code changes, as the code changes are contained in this PR. The only code changes were around P/Invoke handling, where we restrict P/Invoke to assemblies that are in the Windows\System32 directory


src/SystemAbstractions/SystemAbstractions.csproj
```diff
@@ -9,7 +9,15 @@
   <Import Project="..\..\build\NetStandardRelease.targets" />
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.9.5-beta1.final">
```
RobGallo: I'm wondering if these additions should be in build/NetStandardTest.targets.

src/Win32/InternalsVisibleTo.cs
```diff
@@ -2,6 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System.Diagnostics.CodeAnalysis;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// Limit P/Invoke to assemblies located in the System32 folder
```
RobGallo: I see now why you wanted to rename this file. The floor is open.

src/Win32/Win32.csproj
```diff
@@ -7,7 +7,15 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.9.5-beta1.final">
```
RobGallo: These are still/already set in build/NetStandardRelease.targets. I'm wondering if the analyzers should be in NetStandardAll.targets

DaveTryon: Per our conversation, removed from NetStandardRelease.targets and leaving them where VS tooling can update them

src/SystemAbstractions/SystemAbstractions.csproj
DaveTryon: Per our conversation, removed from NetStandardRelease.targets and leaving them where VS tooling can update them

src/Win32/InternalsVisibleTo.cs
DaveTryon: Per our conversation, file is again named AssemblyInfo.cs. I checked docs\AssessingInternals.md, and that file already says AssemblyInfo.cs

PR #155, opened 2019-09-24
Add control sampler apps for testing

DaveTryon:
#### Describe the change
We occasionally receive "false positive" reports from issues in frameworks, but we have no explicit testing of them. This adds a couple of test apps (one using WinForms, one using WPF) to provide a "sampler" of framework-provided controls. In the short term, we can use this as a baseline for what may require exemptions. In the longer-term, we should create E2E tests to ensure that rule changes don't introduce additional framework-owned errors. Ultimately, we'd probably want to work with the framework teams to be able to fix the issues, then enforce the rules as fixes are introduced.


tools/WpfControlSampler/src/WpfControlSampler/MainWindow.xaml
```diff
@@ -0,0 +1,50 @@
...
+            <TabItem Header="TabItem" AutomationProperties.Name="First Tab">
+                <Grid Background="#FFE5E5E5"/>
+            </TabItem>
+            <TabItem Header="TabItem" AutomationProperties.Name="Second Tab">
+                <Grid Background="#FFE5E5E5"/>
+            </TabItem>
+        </TabControl>
+        <DatePicker HorizontalAlignment="Left" Margin="248,174,0,0" VerticalAlignment="Top" AutomationProperties.Name="Please pick a date"/>
+        <Calendar HorizontalAlignment="Left" Margin="248,224,0,0" VerticalAlignment="Top" AutomationProperties.Name="Monthly snapshot"/>
+
```
jalkire:
```suggestion
```

PR #160, opened 2019-10-04
Use VS2019 build environment in both PR and signed builds

DaveTryon:
#### Describe the change
Try VS2019 in the build loops. Test only at this point


dbjorge:
LGTM, but waiting for (internal) test signed build before approving/merging

PR #166, opened 2019-10-16
Enable Roslyn Analyzers, gather more data in compliance artifacts

### LINKED ISSUE #152: [ENG] SystemAbstractionsTests are not running in build loop
Please check whether the bug has [already been filed](https://github.com/Microsoft/axe-windows/issues).

**Describe the bug**
Tests from SystemAbstractionsTests.dll are not running in the build loop. 

**To Reproduce**
1. Open the results of a build loop (I tried both PR builds and official builds)
2. Open the test results
3. Search for any of the following unit tests
    - RegistryGetValue_Matches
    - CreateDirectory_CreatesDirectory
    - CurrentDirectory_Matches
    - DateTime_Matches
    - DirectoryExists_ReturnsExpected

**Expected behavior**
Unit tests should be executing

**Actual behavior**
Unit tests do not execute. There is, however, the following message in the logs:
[MSTest][Discovery][d:\a\1\s\src\SystemAbstractionsTests\bin\Release\netcoreapp2.1\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll] Unable to load types from the test source 'd:\a\1\s\src\SystemAbstractionsTests\bin\Release\netcoreapp2.1\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll'. Some or all of the tests in this source may not be discovered.
[MSTest][Discovery][d:\a\1\s\src\SystemAbstractionsTests\bin\Release\netcoreapp2.1\SystemAbstractionsTests.dll] Unable to load types from the test source 'd:\a\1\s\src\SystemAbstractionsTests\bin\Release\netcoreapp2.1\SystemAbstractionsTests.dll'. Some or all of the tests in this source may not be discovered.
[MSTest][Discovery][d:\a\1\s\src\packages\MSTest.TestAdapter.1.4.0\build\netcoreapp1.0\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll] Unable to load types from the test source 'd:\a\1\s\src\packages\MSTest.TestAdapter.1.4.0\build\netcoreapp1.0\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll'. Some or all of the tests in this source may not be discovered.
[MSTest][Discovery][d:\a\1\s\src\packages\MSTest.TestAdapter.1.4.0\build\uap10.0\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll] Unable to load types from the test source 'd:\a\1\s\src\packages\MSTest.TestAdapter.1.4.0\build\uap10.0\Microsoft.VisualStudio.TestPlatform.MSTestAdapter.PlatformServices.dll'. Some or all of the tests in this source may not be discovered.

**Additional context**
These unit tests work as expected (meaning they are discovered and run successfully) when run locally

DaveTryon:
#### Describe the change
Fix #152. The Roslyn build task is unable to obtain any information about what it needs to scan. This change does 3 things:
1. Uses alternate arguments to launch the Roslyn pass. The argument uses a hardcoded path to the default location of VS2019. I don't like this at all, but it seems to be a necessary evil for the short term. Long-term, we may be able to revert to using the default arguments (this will require as-yet-unknown changes in the build definition)
2. Fixes a small error that was causing us to attempt to upload FxCop results instead of Roslyn results
3. Moves the file copy and artifact publishing as late as possible, so that we can capture more files (especially log files) in the published artifacts.

Unless marked otherwise with a comment, changes are all cut/paste within a job

### DIFF
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -133,7 +133,7 @@ jobs:
         userProvideBuildInfo: msBuildInfo
         msbuildVersion: 16.0
         msBuildArchitecture: '$(BuildPlatform)'
-        msBuildCommandline: '"$(MSBuildPath)" "$(ProjectFile)" /nologo /nr:false /fl /flp:"logfile=$(ProjectFile).log;verbosity=diagnostic" /p:platform="$(BuildPlatform)" /p:configuration="debug" /p:VisualStudioVersion="16.0"'
+        msBuildCommandline: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\src\Axe.Windows.sln /p:platform="$(BuildPlatform)" /p:configuration="debug" /p:VisualStudioVersion="16.0"'
         rulesetName: recommended
         internalAnalyzersVersion: Latest
         microsoftAnalyzersVersion: Latest
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -133,7 +133,7 @@ jobs:
         userProvideBuildInfo: msBuildInfo
         msbuildVersion: 16.0
         msBuildArchitecture: '$(BuildPlatform)'
-        msBuildCommandline: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" .\src\Axe.Windows.sln /p:platform="$(BuildPlatform)" /p:configuration="debug" /p:VisualStudioVersion="16.0"'
+        msBuildCommandline: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" "$(Build.SourcesDirectory)\src\Axe.Windows.sln" /p:platform="$(BuildPlatform)" /p:configuration="debug" /p:VisualStudioVersion="16.0"'
         rulesetName: recommended
         internalAnalyzersVersion: Latest
         microsoftAnalyzersVersion: Latest
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -130,7 +130,10 @@ jobs:
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-roslynanalyzers.RoslynAnalyzers@2
     displayName: 'Run Roslyn analyzers'
     inputs:
-        userProvideBuildInfo: auto
+        userProvideBuildInfo: msBuildInfo
+        msbuildVersion: 16.0
+        msBuildArchitecture: '$(BuildPlatform)'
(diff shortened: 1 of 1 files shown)


build/signedbuild.yml
```diff
@@ -108,20 +97,23 @@ jobs:
...
+    continueOnError: true
+
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-roslynanalyzers.RoslynAnalyzers@2
     displayName: 'Run Roslyn analyzers'
     inputs:
-        userProvideBuildInfo: auto
+        userProvideBuildInfo: msBuildInfo
+        msbuildVersion: 16.0
+        msBuildArchitecture: '$(BuildPlatform)'
+        msBuildCommandline: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\msbuild.exe" "$(Build.SourcesDirectory)\src\AxeWindows.sln" /p:platform="$(BuildPlatform)" /p:configuration="debug" /p:VisualStudioVersion="16.0"'
```
DaveTryon: These lines specify the commands for the Roslyn analyzer. I wish I could get rid of the hardcoded path to msbuild.exe, but that doesn't appear to be an option 😢

build/signedbuild.yml
```diff
@@ -154,9 +146,9 @@ jobs:
       uploadAPIScan: false
       uploadBinSkim: false
       uploadFortifySCA: false
+      uploadFxCop: false
```
DaveTryon: This was `upoadRoslyn: false` before, so TSA was trying to upload FxCop instead of Roslyn

build/signedbuild.yml
```diff
@@ -108,20 +97,23 @@ jobs:
       verboseOutput: true
       debugMode: false
 
+  - task: securedevelopmentteam.vss-secure-development-tools.build-task-policheck.PoliCheck@1
+    displayName: 'Run PoliCheck'
+    inputs:
+      targetType: F
```
RobGallo: Target type "F"? Is this correct?

build/signedbuild.yml
```diff
@@ -154,9 +146,9 @@ jobs:
       uploadAPIScan: false
       uploadBinSkim: false
       uploadFortifySCA: false
+      uploadFxCop: false
       uploadModernCop: false
       uploadPREfast: false
-      uploadRoslyn: false
```
RobGallo: Is it true scans that aren't listed are uploaded?

build/signedbuild.yml
```diff
@@ -303,6 +282,30 @@ jobs:
       uploadRoslyn: false
```
RobGallo: Do we also need to add UploadFSCop: false?

build/signedbuild.yml
DaveTryon: Yes.

PR #175, opened 2019-10-21
Update to Moq 4.13.1

DaveTryon:
#### Describe the change
Use VS to update to Moq 4.13.1


src/ActionsTests/ActionsTests.csproj
```diff
@@ -64,7 +64,7 @@
       <HintPath>..\packages\MSTest.TestFramework.2.0.0\lib\net45\Microsoft.VisualStudio.TestPlatform.TestFramework.Extensions.dll</HintPath>
     </Reference>
     <Reference Include="Moq, Version=4.13.0.0, Culture=neutral, PublicKeyToken=69f491c39445e920, processorArchitecture=MSIL">
```
DaveTryon: @RobGallo asked if the 4.13.0 should also change here. Looking back at previous Moq changes, they always leave the revision value as 0 in the reference. Seems a bit odd, but that's how they build their package and we're just using the data from their package

dbjorge: Usually when a project does this it's because they want to avoid the need for users to have their projects do binding redirects between versions that are known-API-compatible with one another

PR #200, opened 2019-10-25
Remove TextAnalyzers from UnitTestSharedLibrary

DaveTryon:
#### Describe the change
We don't need to run TextAnalyzers on unit tests. This was overlooked in #174 


DaveTryon:
This has already been handled

PR #213, opened 2019-11-06
Prune more fakes

DaveTryon:
#### Describe the change
Clean up some spurious references and usages of fakes. At this point, the only build and referenced namespaces are those are used by 3 test cases in ActionsTests.ScreenShotActionUnitTests. Future work will ideally convert those 3 test cases, as well.


src/ActionsTests/Misc/ExtensionMethodsTests.cs
```diff
@@ -21,19 +20,16 @@ public class ExtensionMethodsTests
         [TestMethod()]
         public void GetSmallestElementFromPointTest()
```
DaveTryon: The changes in this file are much clearer if you enable "Hide whitespace changes". We're just removing unnecessary `using ShimsContext.Create())` blocks

DaveTryon:
GitHub doesn't show the fakes since they're binary files. This change removes mscorlib.fakes and Axe.Windows.Desktop.fakes (nothing was referencing fakes from these assemblies). The change also removes all stubs from Axe.Windows.Actions (they weren't used) and removes shims from Axe.Windows.Actions.Misc (they weren't used, either)

PR #214, opened 2019-11-07
Remove Axe.Windows methods that aren't used by Axe.Windows

DaveTryon:
#### Describe the change
We identified some Axe.Windows methods that are used only in AIWin. We're adding these methods to AIWin and removing them from Axe.Windows. This is the complement of https://github.com/microsoft/accessibility-insights-windows/pull/630


RobGallo:
LGTM, although you can remove System.Drawing.Common from the Win32 proj.

DaveTryon:
> LGTM, although you can remove System.Drawing.Common from the Win32 proj.

You're right, but I'm leaving it in for now to avoid thrashing in the nuspec file. We're likely to use System.Drawing.Common in upcoming work, and Win32 is currently the only component to reference it. We should remove it from Win32 once we have references in the longer-term locations

PR #215, opened 2019-11-09
Added file Rules Description

ferBonnin:
#### Describe the change
Added file with rules descriptions. The markdown is auto-generated but added manually a column with the type of failure (error or warning)
This is a table with markdown. The lines do not wrap so we have a horizontal scrollbar.
I couldn't find an easy way to make them wrap with markdown, but we could either have the table as an html table or move the description as the last column.

### DIFF
--- a/docs/RulesDescription.md
+++ b/docs/RulesDescription.md
@@ -1,0 +1,153 @@
+
+## Rules in Axe.Windows
+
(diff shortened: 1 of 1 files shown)


ferBonnin:
Peter will create a new one that has the table with no horizontal scrolling

PR #221, opened 2019-11-12
Add text to explicitly specify which APIs are supported by axe-windows

RobGallo:
#### Describe the change

Adding text to let users know which APIs are considered supported or not supported by Accessibility Insights.

### DIFF
--- a/docs/AutomationReference.md
+++ b/docs/AutomationReference.md
@@ -1,5 +1,7 @@
 ## Axe.Windows - automation reference
 
+Note: The following documentation describes the supported API provided by Axe.Windows. There are other, public interfaces and classes in various assemblies throughout the package; however, please be aware interfaces and classes not described in this document are not supported and may change at any time.
+
 ### Class details
 
 #### Config.Builder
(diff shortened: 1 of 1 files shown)


docs/AutomationReference.md
```diff
@@ -1,5 +1,7 @@
 ## Axe.Windows - automation reference
 
+Note: The following documentation describes the supported API provided by Axe.Windows. There are other, public interfaces and classes in various assemblies throughout the package; however, please be aware interfaces and classes not described in this document are not supported and may change at any time.
```
karanbirsingh: I think this text is good. I do wonder whether we should put it at a 'higher' level in the repo such as the root .README file. That way it will be hard to miss regardless of what part of the docs / code one explores?

RobGallo: I looked at the readme, but I didn't feel like I found a good fit there. Suggestions?

DaveTryon: Add a "that" for sentence flow?
```suggestion
Note: The following documentation describes the supported API provided by Axe.Windows. There are other, public interfaces and classes in various assemblies throughout the package; however, please be aware that interfaces and classes not described in this document are not supported and may change at any time.
```

PR #222, opened 2019-11-18
Remove Fakes used in ScreenShotAction unit tests

RobGallo:
#### Describe the change

Using static property injection as a light-weight way to avoid using Fakes for getting the Datamanager and creating Bitmaps.

The original tests ensured the ElementDataContext.Screenshot property setter was called. But since that would involve a larger change including creating an interface which could be mocked to replace ElementDataContext, that idea was abandoned in favor of simply checking the resulting data.

I will remove all remaining vestiges of Fakes in another PR. I thought it would be good to keep this PR as minimal as practical.

### DIFF
--- a/src/Actions/Actions/ScreenShotAction.cs
+++ b/src/Actions/Actions/ScreenShotAction.cs
@@ -16,6 +16,10 @@ namespace Axe.Windows.Actions
     [InteractionLevel(UxInteractionLevel.NoUxInteraction)]
     public static class ScreenShotAction
     {
+        // unit test hooks
+        internal static Func<DataManager> GetDataManager = () => DataManager.GetDefaultInstance();
+        internal static Func<int, int, Bitmap> CreateBitmap = (width, height) => new Bitmap(width, height);
+
         /// <summary>
         /// Take a screenshot of the given element's parent window, if it has one
         ///     returns null if the bounding rectangle is 0-sized
@@ -25,7 +29,7 @@ namespace Axe.Windows.Actions
         {
             try
             {
-                var ec = DataManager.GetDefaultInstance().GetElementContext(ecId);
+                var ec = GetDataManager().GetElementContext(ecId);
                 var el = ec.Element;
 
                 var win = el.GetParentWindow();
@@ -35,7 +39,8 @@ namespace Axe.Windows.Actions
                 {
                     return; // no capture. 
                 }
-                Bitmap bmp = new Bitmap(rect.Width, rect.Height);
+
+                Bitmap bmp = CreateBitmap(rect.Width, rect.Height);
                 Graphics g = Graphics.FromImage(bmp);
 
                 g.CopyFromScreen(rect.X, rect.Y, 0, 0, rect.Size);
--- a/src/Actions/Contexts/ElementContext.cs
+++ b/src/Actions/Contexts/ElementContext.cs
@@ -71,7 +71,7 @@ namespace Axe.Windows.Actions.Contexts
                 return _dataContext;
             }
 
-            set
+            internal set
             {
                 _dataContext?.Dispose();
                 _dataContext = value;
--- a/src/Actions/Contexts/ElementContext.cs
+++ b/src/Actions/Contexts/ElementContext.cs
@@ -71,7 +71,7 @@ namespace Axe.Windows.Actions.Contexts
                 return _dataContext;
             }
 
-            internal set
+            set
             {
                 _dataContext?.Dispose();
                 _dataContext = value;
--- a/src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
+++ b/src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
@@ -29,42 +29,40 @@ namespace Axe.Windows.ActionsTests.Actions
             element.Properties.Add(typeId, new A11yProperty(typeId, data));
         }
 
+        [TestInitialize]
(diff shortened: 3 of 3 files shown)


src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
```diff
@@ -29,127 +29,103 @@ private void SetBoundingRectangle(A11yElement element, Rectangle b)
...
 
-                // bitmap is set
-                Assert.IsTrue(bitmapsetcalled);
+                Assert.IsNotNull(dc.Screenshot);
+                Assert.AreEqual(element.UniqueId, dc.ScreenshotElementId);
             }
         }
 
-        [TestMethod]
+            [TestMethod]
```
DaveTryon: An extra indent was added. Sugestion:
```suggestion
        [TestMethod]
```

src/Actions/Actions/ScreenShotAction.cs
```diff
@@ -16,6 +16,10 @@ namespace Axe.Windows.Actions
     [InteractionLevel(UxInteractionLevel.NoUxInteraction)]
     public static class ScreenShotAction
     {
+        // unit test hooks
+        internal static Func<DataManager> GetDataManager = () => DataManager.GetDefaultInstance();
```
DaveTryon: Code coverage shows that the call to GetDefaultInstance is never being called here. That would suggest that this line can be removed

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
```diff
@@ -29,127 +29,103 @@ private void SetBoundingRectangle(A11yElement element, Rectangle b)
             element.Properties.Add(typeId, new A11yProperty(typeId, data));
         }
 
+        [TestInitialize]
+        public void TestInitialize()
+        {
+            ScreenShotAction.GetDataManager = () => DataManager.GetDefaultInstance();
```
DaveTryon: Code coverage shows that the call to GetDefaultInstance is never being called here. That would suggest that this line can be removed

src/Actions/Contexts/ElementContext.cs
```diff
@@ -71,7 +71,7 @@ public ElementDataContext DataContext
                 return _dataContext;
             }
 
-            set
+            internal set
```
DaveTryon: I'm a bit concerned that someone might incorrectly try to use this method in production code. Should we consider adding a simple wrapper (something like SetElementContextForUnitTests) that then calls the setter?

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
RobGallo: But don't you agree that the pattern is good practice? I think it's good to reset such values to their defaults in case someone comes along behind us and writes a new unit test, then gets unexpected results that would be hard to track down.

src/Actions/Contexts/ElementContext.cs
RobGallo: I like it. Will do.

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
dbjorge: I agree with that principle, but I think if that's the intended purpose it might be better to do this in either a `[TestCleanup]` method or a `Dispose()` of a using-based helper to temporarily set it to something different (see the `CaptureActionTestHookOverrides` class inside `CaptureActionUnitTests` for an example of this)

DaveTryon: One of the benefits of the fakes was that we never accidentally fell through to any underlying classes. This pattern seems to discard that benefit, so I'm not really a fan. Setting these on a per-test basis feels like the better approach. If we want to make the tests easy to debug, the TestInitialize code would call an Assert.Fail with an error message to tell the code writer that they forgot to set the test override.

src/Actions/Contexts/ElementContext.cs
RobGallo: Turns out this change wasn't necessary for the commit. I've reverted it.

src/ActionsTests/Actions/ScreenShotActionUnitTests.cs
RobGallo: @dbjorge I have updated the code to use [TestCleanup], which is the right attribute to use since it ensures the DataManager will be reset after it's potentially been changed and not before it's potentially used. So thanks for that.

Before I began, I studied your solution in CaptureActionTestHookOverrides. If this were production code, I would have considered using either that technique or adding a method with a callback to ensure the static members were reset to their default values. However, this is not production code and the tests provide a mechanism for exactly this purpose, which I used because it's simpler and conventional in unit tests.

@DaveTryon Let's don't lose sight of what we're trying to do here and the environment (singletons) in which we find ourselves working. None of us likes the static injection. The options are to

1. Continue using Fakes
2. Use a RAII technique as Dan did
3. Use a callback mechanism which is essentially like RAII except calling a method instead of creating an object
4. Using [TestCleanup]
5. Rewrite the code to use interfaces and mocks

Using Fakes is obviously not desireable. As I explained above, I find the RAII techniques to be unnecessary given that unit tests have both a mechanism and a convention to deal with this type of problem. And I don't think we want to invest in a rewrite of the code. Ergo, I chose the solution that I did.

Issue #244, opened 2019-12-07
[BUG] Type mismatches are leading to internal Exceptions


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/axe-windows/issues).

## Describe the bug

When saving to A11yElement.PlatformProperties, the Window Style and Window Extended Style are added as uints, because that's the return type declared in the P/Invoke wrapper of GetWindowLong. When fetching the values via A11yElement.GetPlatformPropertyValue<int>, we have a mismatch between int and uint and throw an Exception, which later gets caught, but we've interrupted what we were trying to retrieve. Looking through the code, some calls to A11yElement.GetPlatformPropertyValue specify int, while others specify uint. After talking with @RobGallo, we'll standardize on int to not impact any publicly exposed interfaces.

DaveTryon:
This was fixed in commit https://github.com/microsoft/axe-windows/commit/c4c2fdd88fb040020f62c119093e13128a5443ff but I forgot to close the issue

PR #245, opened 2019-12-07
Standardize on int for values returned from GetWindowLong

DaveTryon:
#### Describe the change

This fixes issue #244 by forcing int as the type for numeric platform properties. A11yElement.GetPlatformPropertyValue<T> is currently only used for values from GetWindowLong, which we're standardizing as int (instead of always writing as uint, sometimes reading as uint, and sometimes reading as int). I've hidden GetPlatformPropertyValue<T> and exposed GetPlatformPropertyInt in its place--if/when we need to expose other value types, we may wish to add additional wrappers.

This is a breaking change for any code implementing IA11yElement, but that interface isn't publicly documented and we know of nobody trying to implement their own versions.

Tested by successfully running all unit tests, then running all unit tests under the debugger with a breakpoint at the point where we throw on a mismatch. Breakpoint was never hit, so no mismatches remain.

### DIFF
--- a/src/Core/Bases/A11yElement.cs
+++ b/src/Core/Bases/A11yElement.cs
@@ -409,12 +409,22 @@ namespace Axe.Windows.Core.Bases
         }
 
         /// <summary>
+        /// Gets a specified int property value for the current platform, e.g., Windows.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


DaveTryon:
Fixed this by changing the specific case that was broken; doesn't look like we'll be making this change

PR #246, opened 2019-12-07
Correct spelling of "Sibling", which only has 1 'b'. 

DaveTryon:
#### Describe the change

Correct spelling of "Sibling", which only has 1 'b'. This will require some minor changes in AIWin when we integrate it there, but has no impact on automation consumers.

### DIFF
--- a/src/Actions/Trackers/TreeTracker.cs
+++ b/src/Actions/Trackers/TreeTracker.cs
@@ -73,24 +73,24 @@ namespace Axe.Windows.Actions.Trackers
             return treeWalker?.GetLastChildElement(element);
         }
 
-        public void MoveToNextSibbling()
+        public void MoveToNextSibling()
         {
-            MoveTo(this.GetNexSibbling);
+            MoveTo(this.GetNexSibling);
         }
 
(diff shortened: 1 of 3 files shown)


src/Actions/Trackers/TreeTracker.cs
```diff
@@ -73,24 +73,24 @@ private IUIAutomationElement GetLastChild(IUIAutomationTreeWalker treeWalker, IU
...
 
-        public void MoveToNextSibbling()
+        public void MoveToNextSibling()
         {
-            MoveTo(this.GetNexSibbling);
+            MoveTo(this.GetNexSibling);
         }
 
-        private IUIAutomationElement GetNexSibbling(IUIAutomationTreeWalker treeWalker, IUIAutomationElement element)
+        private IUIAutomationElement GetNexSibling(IUIAutomationTreeWalker treeWalker, IUIAutomationElement element)
```
dbjorge: GetNex **t** Sibling?

DaveTryon: Fixed--thanks for catching :)

PR #247, opened 2019-12-07
Automate scans of control sampler apps

DaveTryon:
#### Describe the change

We recently added some "control sampler" apps to help pin our behavior with controls from different frameworks. These are __not__ intended to be exhaustive tests of framework controls. Instead, they are intended to pin the behavior of our rules and to prevent unintended consequences of rule additions or modifications. These apps currently rely on manual testing--this change wraps them in integration tests so that they run automatically.

The existing code includes a 10 second delay when starting the test app, since experience has shown that the pipeline machines are sometimes slow to respond to the newly started app. Rather than add 30 more seconds to the local unit tests, I've optimized the delay to be only 2 seconds when running on a machine outside of the build pipeline (we use the VSTS variable to decide if we're building in a pipeline). The net result is that local test runs now run 4 integration tests in about the same time it used to take to run only 1. We can tune the 3 second delay if it proves insufficient for some dev machines.

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -3,6 +3,7 @@
 using Axe.Windows.Automation;
 using Axe.Windows.UnitTestSharedLibrary;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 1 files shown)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -14,11 +15,26 @@ namespace Axe.Windows.AutomationTests
...
+        readonly string WindowsFormsControlSamplerAppPath = Path.GetFullPath("../../../../tools/WindowsFormsControlSampler/WindowsFormsControlSampler.exe");
+        readonly string WpfControlSamplerAppPath = Path.GetFullPath("../../../../tools/WpfControlSampler/WpfControlSampler.exe");
+
         readonly string OutputDir = Path.GetFullPath("./TestOutput");
         readonly string ValidationAppFolder;
         readonly string ValidationApp;
 
+        // Build agents need more than than local dev machines to have the test app
+        // up and running. Pipelines set BUILD_BUILDID, dev machines don't
+        private TimeSpan testAppDelay = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("BUILD_BUILDID")) ?
```
jalkire: 👏

PR #248, opened 2019-12-10
Allow uint property values to be read as ints; framework can be made generic

### LINKED ISSUE #244: [BUG] Type mismatches are leading to internal Exceptions

DaveTryon:
#### Describe the change

Allow uint property values to be read as ints; framework can be made generic. This fixes #244 by allowing for specific type mappings. This change does not attempt to handle all mappings (for example, a uint can appropriately fit into a long), but only covers the cases that we know occur. 

It also throws an `InvalidCastException` instead of throwing a generic `Exception`, since it's generally discouraged to throw the basic `Exception` type.

### DIFF
--- a/src/Core/Bases/A11yElement.cs
+++ b/src/Core/Bases/A11yElement.cs
@@ -418,9 +418,22 @@ namespace Axe.Windows.Core.Bases
         {
             var property = this.PlatformProperties?.ById(propertyId);
             if (property == null) return default(T);
-            if (!(property.Value is T)) throw new Exception(Invariant($"Expected property.Value, which is type {property.Value.GetType().Name}, to be type {typeof(T).Name}"));
+            Type propertyType = property.Value.GetType();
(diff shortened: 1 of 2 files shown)


DaveTryon:
Fixed this by changing the specific case that was broken; doesn't look like we'll be making this change

Issue #162, opened 2019-10-04
[Feature Request] Add rules to ensure that an element's clickable point and IsOffscreen property agree.


RobGallo:
### new rule

Based on https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.automationelement.automationelementinformation.isoffscreen?view=netframework-4.8:

“If the element has a clickable point that can cause it to be focused, it is considered to be onscreen even when a portion of the element is off screen.”

Based on this, there should be two rules:
1. When an element's clickable point is on screen, its IsOffscreen property should be set to false.
2. When an element's clickable point is off screen, its IsOffscreen property should be set to true.

ferBonnin:
Reviewed this with PMs and Leads, this looks like a good suggestion, but we need to schedule time to do this feature. @RobGallo is this a 2-day thing or a 2-week thing?

RobGallo:
@ferBonnin, it's more of a two day thing.

ferBonnin:
@peterdur do we want to take this and the other rule (Also 2 days) as a feature or in slack time?

RobGallo:
This issue was fixed and released in axe-windows version 0.9.0.

PR #250, opened 2019-12-11
Fix Issue #162 - [Feature Request] Add rules to ensure that an element's clickable point and IsOffscreen property agree. PR (1/2)

RobGallo:
#### Describe the change

- Add a new rule, ClicablePointOnScreen, to test that when the clickable point for an element is within the desktop's bounding rectangle, the element's IsOffScreen property is set to false.
- Add a new property condition ClickablePoint.OnScreen -- tests if an element's clickable point both exists and is within the desktop's bounding rectangle.
- Remove DesktopElement.ExcludedPropertyIds because it is not used by axe windows or ai-win.
- Add code to add the ClickablePoint property to element's which are not inside Edge. According to the existing comments, querying for the ClickablePoint property causes a crash in Edge.
- Add an if condition to CurrentFileVersionCompatibilityTests to ignore element's of type thumb when ensuring bounding rectangles are not empty. This is necessary because the new rule, ClickablePointOnScreen, adds an additional error on a thumb element with an empty bounding rectangle to the a11ytest file.

### DIFF
--- a/src/Rules/Extensions/ExtensionMethods.cs
+++ b/src/Rules/Extensions/ExtensionMethods.cs
@@ -8,7 +8,7 @@ using Axe.Windows.Core.Enums;
 using Axe.Windows.Core.Misc;
 using Axe.Windows.Core.Types;
 using Axe.Windows.Rules.PropertyConditions;
-
+using UIAutomationClient;
 using static System.FormattableString;
 
 namespace Axe.Windows.Rules.Extensions
@@ -141,6 +141,14 @@ namespace Axe.Windows.Rules.Extensions
                 && a.Right >= (b.Right + BoundingRectangle.OverlapMargin)
                 && a.Bottom >= (b.Bottom + BoundingRectangle.OverlapMargin)
                 && a.Area() != (b.Area() + (BoundingRectangle.OverlapMargin * 4));
+        }
+
+        public static bool Contains(this tagRECT rect, Point point)
+        {
+            return point.X >= rect.left
+                && point.X <= rect.right
+                && point.Y >= rect.top
+                && point.Y <= rect.bottom;
         }
     } // class
 } // namespace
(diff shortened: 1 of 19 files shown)


DaveTryon:
Test runs are failing because an additional error is being flagged in WildlifeManager.exe. That may be expected, based on conversations we had offline. You'll want to update `WildlifeManagerKnownErrorCount` in AutomationIntegrationTests.cs

src/Rules/Extensions/ExtensionMethods.cs
```diff
@@ -8,7 +8,7 @@
 using Axe.Windows.Core.Misc;
 using Axe.Windows.Core.Types;
 using Axe.Windows.Rules.PropertyConditions;
-
+using UIAutomationClient;
```
DaveTryon: Is this using statement necessary?

src/Rules/Misc/Helpers.cs
```diff
@@ -0,0 +1,24 @@
...
+using UIAutomationClient;
+
+namespace Axe.Windows.Rules.Misc
+{
+    static class Helpers
+    {
+        private static readonly Lazy<IUIAutomationElement> _desktop = new Lazy<IUIAutomationElement>(GetDesktopElement);
+        public static IUIAutomationElement Desktop => _desktop.Value;
+
+        private static IUIAutomationElement GetDesktopElement()
```
DaveTryon: This seems odd to me. Rules (so far) is not a data collector--it operates purely on provided data. Now it's reaching out to the system to collect its own data.

src/Rules/Extensions/ExtensionMethods.cs
RobGallo: Yes, because the interop type tagRECT is defined in UIAutomation.

PR #252, opened 2019-12-13
Allow integration tests to return inconclusive in build pipelines

DaveTryon:
#### Describe the change

The Integration tests are failing in the signed build pipeline. We don't yet understand _why_ the tests are failing the signed build, especially since they're working in PR builds (which use the same environment). This test does not fix the tests--it merely allows them to return inconclusive in the signed build. This is a stopgap to mitigate the broken pipeline, but will need a more permanent fix (possibly after the New Year).

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -30,10 +30,8 @@ namespace Axe.Windows.AutomationTests
         readonly string ValidationAppFolder;
         readonly string ValidationApp;
 
-        // Build agents need more than than local dev machines to have the test app
-        // up and running. Pipelines set BUILD_BUILDID, dev machines don't
-        private TimeSpan testAppDelay = string.IsNullOrEmpty(Environment.GetEnvironmentVariable("BUILD_BUILDID")) ?
-            TimeSpan.FromSeconds(2) : TimeSpan.FromSeconds(10);
+        private readonly TimeSpan testAppDelay;
+        private readonly bool allowInconclusive;
 
         public AutomationIntegrationTests()
         {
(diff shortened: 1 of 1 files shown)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -46,6 +44,21 @@ public AutomationIntegrationTests()
 #endif
                 ));
             ValidationApp = Path.Combine(ValidationAppFolder, @"CurrentFileVersionCompatibilityTests.exe");
+
+            // Build agents are less predictable that dev machines. Set the flags based
```
RobGallo:
```suggestion
            // Build agents are less predictable than dev machines. Set the flags based
```

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -123,6 +136,22 @@ private ScanResults Scan_Integration_Core(string testAppPath, int expectedErrorC
...
+            try
+            {
+                return scanner.Scan();
+            }
+            catch (AxeWindowsAutomationException e)
+            {
+                if (allowInconclusive && e.Message.Contains("Automation017"))
+                {
+                    Assert.Inconclusive("Unable to complete Integration tests");
+                }
```
RobGallo: Shouldn't there be a return statement here to return before throwing the exception?

src/AutomationTests/AutomationIntegrationTests.cs
DaveTryon: Fixed

src/AutomationTests/AutomationIntegrationTests.cs
DaveTryon: Not needed. Assert.Inconclusive ends the test. According to https://docs.microsoft.com/en-us/dotnet/api/microsoft.visualstudio.testtools.unittesting.assert.inconclusive?view=mstest-net-1.2.0, AssertInconclusive is a wrapper than throws an AssertInconclusiveException, when then gets handled by the framework.

PR #264, opened 2020-01-06
Correct reported values for HeadingLevel propery

### LINKED ISSUE #263: UIA HeadingLevel values are reported incorrectly
**This was transferred from AIWin, but the root cause is in Axe.Windows** 

Please check whether the bug has [already been filed](https://github.com/Microsoft/accessibility-insights-windows/issues).

**Describe the bug**
UIA HeadingLevel properties are not reported in the most helpful way to AIWin users.

**To Reproduce**
1. Go to the Windows Settings app System Sound page. 
2. Point the AIWin tool at the "Output" heading on the page. 
3. Check the reported UIA HeadingLevel property data, "Unknown(80052)".

**Expected behavior**
As listed at https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-heading-level-identifiers, a HeadingLevel property value of 80052 is "Heading level 2". As such, AIWin should say something like "Heading level 2, (80052)".

**Actual behavior**
By saying "Unknown(80052)", it's less likely that the AIWin user will learn the heading level of the UI of interest.

**Screenshots or .GIF**
AIWin_HeadingLevel_Unknown

**Desktop (please complete the following information):**
 - OS: Windows 10 1909, OS Build 18363.535
 - Accessibility Insights for Windows Version: 1.1.1072.1
 - Target Application: Settings app in Windows.

**Additional context**

DaveTryon:
#### Describe the change
Fixes #263 

Most UIA properties are prefixed with the UIA_ prefix. Values for the HeadingLevel property (see 
https://msdn.microsoft.com/en-us/library/windows/desktop/mt829376(v=vs.85).aspx) don't follow this pattern, and the incorrect values were not noticed until very recently.

This change ensures that the values are correctly reported, and adds unit tests to pin the behavior.


src/CoreTests/Types/HeadingLevelTypeTests.cs
```diff
@@ -0,0 +1,73 @@
...
+        [TestMethod()]
+        public void HeadingLevelNone()
+        {
+            Assert.AreEqual("HeadingLevel_None(80050)", HeadingLevelType.GetInstance().GetNameById(HeadingLevelType.HeadingLevelNone));
+        }
+
+        [TestMethod()]
+        public void HeadingLevel1()
+        {
+            Assert.AreEqual("HeadingLevel1(80051)", HeadingLevelType.GetInstance().GetNameById(HeadingLevelType.HeadingLevel1));
```
DaveTryon: @RobGallo, do you think we should use a friendlier name here, say "Heading Level 1" or maybe just "Level 1" ? The values I've chosen here match the documentation values, but aren't necessarily friendly

PR #267, opened 2020-01-07
Convert Rules project to .Net Standard

RobGallo:
#### Describe the change

This is part of the work to convert axe-windows to .Net Standard.

- Convert RulesTests to .NetCore as well


src/RulesTest/RulesTests.csproj
```diff
@@ -1,203 +1,33 @@
...
-    <Compile Include="PropertyConditions\IntPropertyTest.cs" />
-    <Compile Include="MonsterTest.cs" />
-    <Compile Include="PropertyConditions\LandmarksTest.cs" />
-    <Compile Include="PropertyConditions\EnumPropertyUnitTests.cs" />
-    <Compile Include="PropertyConditions\ElementGroupsTests.cs" />
-    <Compile Include="PropertyConditions\ValueConditionUnitTests.cs" />
-    <Compile Include="RulesProviderTests.cs" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0-preview-20200102-01" />
+    <PackageReference Include="Moq" Version="4.13.1" />
+    <PackageReference Include="MSTest.TestAdapter" Version="2.1.0-beta2" />
```
DaveTryon: Would it make sense to use the non-prerelease assemblies? It looks like the prerelease versions got picked up in the project that is our template, but I'd prefer to use released versions, if possible. Post-change cleanup step, maybe?

PR #271, opened 2020-01-07
Explictily Disable signing of .NET Standard assemblies

DaveTryon:
#### Describe the change
Assemblies built with the .NET Core SDK are signed by default in the signed build environment. We explicitly don't want to sign these assemblies, so we need to explicitly add the property to disable signing.

Validation build: https://dev.azure.com/mseng/1ES/_build/results?buildId=11254969&view=results


DaveTryon:
This wasn't the correct fix

PR #275, opened 2020-01-08
Restore Assembly Metadata to projects built with .NET Core SDK

DaveTryon:
#### Describe the change
In porting from .NET Framework to .NET Core, we stopped setting the following assembly metadata fields that we used to set:
AssemblyTitle (becomes the "FileDescription" property)
AssemblyProduct (becomes the "ProductName" property)
AssemblyCopyright (becomes the "LegalCopyright" property)

This PR sets the AssemblyTitle and AssemblyProduct values to their pre-port values, and sets the Copyright to 2020 so it's up-to-date. Future changes could potentially use a common value for AssemblyProduct (instead of using the assembly name) and/or centralize these values in a .props file, but the objective here is to achieve parity (modulo the updated copyright).

### DIFF
--- a/src/Win32/AssemblyInfo.cs
+++ b/src/Win32/AssemblyInfo.cs
@@ -1,8 +1,13 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
+
+[assembly: AssemblyProduct("Axe.Windows.Telemetry")]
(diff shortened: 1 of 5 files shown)


src/Win32/AssemblyInfo.cs
```diff
@@ -1,9 +1,14 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System.Diagnostics.CodeAnalysis;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
+[assembly: AssemblyProduct("Axe.Windows.Telemetry")]
```
RobGallo: I believe this should be Win32.

PR #276, opened 2020-01-09
Let IScanner.Scan caller specify scan ID (intended for CLI usage)

DaveTryon:
#### Describe the change
The automation layer's current scanning code auto-generates a name for the output file. This is fine for code calling the automation layer directly, since the returned data includes the name of the generated file. It's more problematic for the CLI that we're creating, since the CLI can only return an exit code. We could try to create ways to expose the name to the caller, but they all add complexity and increase the work for whoever is writing the script.

To simplify scripting via the CLI, we'll allow the CLI to accept an optional command line parameter that specifies a scan ID, which will then be used as the base file name of the output file. We'll use the current behavior (auto-generating the id/file name) if this command line parameter is omitted. The scanId is managed via the IOutputFileHelper interface, so it's opaque to the SnapshotCommand class.

As part of this change, I've also moved the ExecutionWrapper up the stack to the Scanner class, instead of having it inside the SnapshotCommand wrapper. It was in the wrapper for historical reasons, but makes more logical sense to have it in the Scanner itself.

### DIFF
--- a/src/Automation/Scanner.cs
+++ b/src/Automation/Scanner.cs
@@ -38,8 +38,7 @@ namespace Axe.Windows.Automation
         /// <returns></returns>
         public ScanResults Scan(string scanId)
         {
-            // Defer validation of scanId so that we can wrap it inside the
-            // ExecutionWrapper
+            // Defer validation of scanId so that we can wrap it with ExecutionWrapper
             return ExecuteScan(scanId, () => ValidateScanId(scanId));
         }
 
--- a/src/Automation/Scanner.cs
+++ b/src/Automation/Scanner.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Automation.Resources;
 using System;
 
 namespace Axe.Windows.Automation
@@ -16,6 +17,7 @@ namespace Axe.Windows.Automation
         {
             if (config == null) throw new ArgumentNullException(nameof(config));
             if (scanTools == null) throw new ArgumentNullException(nameof(scanTools));
+            if (scanTools.OutputFileHelper == null) throw new ArgumentException(ErrorMessages.ScanToolsOutputFileHelperNull, nameof(scanTools));
 
             _config = config;
             _scanTools = scanTools;
@@ -27,7 +29,7 @@ namespace Axe.Windows.Automation
         /// <returns></returns>
         public ScanResults Scan()
         {
-            return ExecuteCommand(null);
+            return ExecuteScan(null, null);
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Automation/SnapshotCommand.cs
```diff
@@ -58,9 +62,6 @@ private static ScanResults ProcessResults(A11yElement element, Guid elementId, C
 
         private static OutputFile WriteOutputFiles(OutputFileFormat outputFileFormat, IScanTools scanTools, A11yElement element, Guid elementId)
         {
-            if (scanTools == null) throw new ArgumentNullException(nameof(scanTools));
```
DaveTryon: These checks now exist in the Execute method, so they never get triggered in this method

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -251,41 +283,52 @@ public void Execute_NoErrors_NoOutputFiles()
...
             _targetElementLocatorMock.VerifyAll();
+            _outputFileHelperMock.VerifyAll();
             _actionsMock.VerifyAll();
             _resultsAssemblerMock.VerifyAll();
         }
 
         [TestMethod]
         [Timeout(1000)]
-        public void Execute_NullOutputFileHelper_ThrowsException()
+        public void Execute_NullOutputFilePath_ThrowsException()
```
DaveTryon: The diff here gets a little muddled because a method was moved. It's clearer to review the 2 separate commits:
[Commit 1 (removed test is commented out)](https://github.com/microsoft/axe-windows/pull/276/commits/c7efa3c438d6485afe7ad3b57a3ca269b571c63b)
[Commit 2 (commented out test is deleted)](https://github.com/microsoft/axe-windows/pull/276/commits/37c9ede56193b6211c0fee99cf891cd38f455374)

src/Automation/SnapshotCommand.cs
```diff
@@ -18,8 +18,9 @@ static class SnapshotCommand
         /// <param name="config">A set of configuration options</param>
         /// <param name="scanTools">A set of tools for writing output files,
         /// creating the expected results format, and finding the target element to scan</param>
+        /// <param name="outputFileNameWithoutExtension">The name (without extension) for the output file (can be null)</param>
         /// <returns>A SnapshotCommandResult that describes the result of the command</returns>
-        public static ScanResults Execute(Config config, IScanTools scanTools)
+        public static ScanResults Execute(Config config, IScanTools scanTools, string outputFileNameWithoutExtension = null)
```
RobGallo: This file doesn't need to be touched at all. You can set the given file name in the Scanner.Scan code before calling Execute.

DaveTryon: I considered setting the file name in `Scanner.Scan`, but it's not guaranteed to work if multiple threads call `Scanner.Scan` concurrently and pass in different file names. The `OutputFileHelper` object persists for the lifetime of the scanner, and is shared by all scans. Since `Scanner.Scan` provides no serialization, a race condition would be created--it would be a "last writer wins" scenario, but it's indeterminate which thread would be the last writer.

Serialization is already provided by `ExecutionWrapper.ExecuteCommand`, so by setting the output file _inside_ the lambda, we completely remove the race condition, no matter how many threads are involved.

src/Automation/Scanner.cs
```diff
@@ -16,18 +17,46 @@ internal Scanner(Config config, IScanTools scanTools)
         {
             if (config == null) throw new ArgumentNullException(nameof(config));
             if (scanTools == null) throw new ArgumentNullException(nameof(scanTools));
+            if (scanTools.OutputFileHelper == null) throw new ArgumentException(ErrorMessages.ScanToolsOutputFileHelperNull, nameof(scanTools));
```
RobGallo: Should be ArgumentNullException.

PR #278, opened 2020-01-13
Enable DotNetCore tests in PR and signed builds

DaveTryon:
#### Describe the change
Our old build loops run tests via vstest.exe, which only understands .NET Framework unit tests. As a result, we were losing tests as we migrated to .NET Core-based tests. This is the root cause of #152. We need to add a task to run the .Net Core CLI task in both the PR and signed build loop definitions. 

Validation in signed build: https://dev.azure.com/mseng/1ES/_build/results?buildId=11295667&view=results

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -59,7 +59,7 @@ jobs:
   - task: DotNetCoreCLI@2
     displayName: 'Test Assemblies (.NET Core) **\*test*.csproj'
     inputs:
-      arguments: --no-build --blame --verbosity detailed --configuration release
+      arguments: --no-build --blame --verbosity normal --configuration release
       command: test
       projects: |
         **\*test*.csproj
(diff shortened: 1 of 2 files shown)


build/prbuild.yml
```diff
@@ -56,6 +56,14 @@ jobs:
       configuration: release
       rerunFailedTests: true
 
+  - task: DotNetCoreCLI@2
```
jalkire: Should we add something like "run-dotnetcore-tests" as a template stored in a separate file? This would mean a future change to the task's arguments would only need to touch one line, not five. debug/release could be passed in as a parameter, [like with the ai-win UI test job](https://github.com/microsoft/accessibility-insights-windows/blob/db0c9b0435f8694b3841bcf61efdfeb3899b1112/build/prbuild.yml#L21)

DaveTryon: Let's consider this as a separate YAML refactoring PR. The objective is this PR is to ensure that our .NET Core unit tests are running correctly

PR #282, opened 2020-01-17
Convert remaining five projects to .Net Standard

RobGallo:
#### Describe the change

Projects converted:

- Actions
- ActionsTests
- Automation
- AutomationTests
- OldFileVersionCompatibilityTests

All these projects needed to be converted at the same time because AutomationTests and OldFileVersionCompatibilityTests both have a dependency on the .Net Framework dll System.Drawing, which conflicts at runtime with the System.Drawing.Common dll used in .Net Standard and .Net Core. In other words, it appears the two dlls can not be used in the same process simultaneously. And this caused unit tests to fail.

Changes:

- Add Win32 NativeMethods.GetCursorPos function to replace System.Windows.Forms.Control.MousePosition which is not available in .Net Standard
- Change paths to sample applications used by integration tests because the output directories have changed with the .Net Standard and .Net Core project types
- Update the nuspec file to reflect that all the projects in the Axe.Windows package now support .Net Standard
- Add a try/catch block in EventListenerFactory to catch the PlatformNotSupportedException thrown in .Net Core when Thread.Abort is called. Aborting threads is not supported in .Net Core. The only way I found to cancel asynchronous work in .Net Core is to use Threading.Tasks. It's unclear that would be compatible with the com requirements of UIA. And in any event, it would be too big a change to be part of this commit.
- Add the Axe.Windows.Win32.HighContrast struct to replace the use of System.Windows.Forms.SystemInformation.HighContrast which is not supported in .Net Standard


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -36,7 +36,7 @@ public class AutomationIntegrationTests
         public AutomationIntegrationTests()
         {
             ValidationAppFolder = Path.GetFullPath(
-                Path.Combine(Directory.GetCurrentDirectory(), @"..\..\..\CurrentFileVersionCompatibilityTests\bin",
+                Path.Combine(Directory.GetCurrentDirectory(), @"../..\..\..\CurrentFileVersionCompatibilityTests\bin",
```
DaveTryon: For consistency, let's use forward slashes throughout
```suggestion
                Path.Combine(Directory.GetCurrentDirectory(), @"../../../../CurrentFileVersionCompatibilityTests\bin",
```

src/Win32/HighContrast.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Runtime.InteropServices;
+
+namespace Axe.Windows.Win32
+{
+    [StructLayout(LayoutKind.Sequential, CharSet = CharSet.Ansi)]
+    internal struct HighContrast
+    {
+        public UInt32 Size { get; private set; }
+        public Int32 Flags { get; private set; }
+        public string DefaultScheme { get; private set; }
```
DaveTryon: DefaultScheme seems to never get used. Should we remove it?

PR #286, opened 2020-01-24
Add a .NET Core CLI to trigger scans via Axe.Windows

DaveTryon:
#### Describe the change

We want a standalone .NET CLI (command line interface) to run Axe-Windows scans via arbitrary scanning engines (as opposed to requiring something that can call into the .NET code). This adds the initial version of the scanner, as well as associated unit tests. Most of the changes are in the CLI and CLITest projects. Changes outside of these projects include:
- Granting CLITest project access to internals of Automation assembly (build test objects)
- Allowing IScanner.Scan(string) to accept a null parameters (early thinking was that we didn't want to allow this. but now that we're writing consuming code, it makes things much simpler)
- Adding 2 new projects to the solution

Subsequent changes will include:
- E2E integration tests using the CLI
- Enabling signing in the build loop
- Requesting UIAccess permissions
- Creating an MSI to make the CLI easier to distribute (MSI is preferable to ZIP due to system requirements to grant UIAccess permissions)


src/AxeWindows.sln
```diff
@@ -14,25 +14,25 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "AutomationTests", "Automati
 		{DCD83A54-54DA-4131-A852-2A129DB3ACB8} = {DCD83A54-54DA-4131-A852-2A129DB3ACB8}
 	EndProjectSection
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Core", "Core\Core.csproj", "{2DA36FA6-4FF8-4DB4-AA7E-2F1CACAE6E35}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Core", "Core\Core.csproj", "{2DA36FA6-4FF8-4DB4-AA7E-2F1CACAE6E35}"
```
DaveTryon: VS changed these guids--no idea why ☹️

src/CLI/IOutputGenerator.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Axe.Windows.Automation;
+
+namespace AxeWindowsScanner
```
DaveTryon: The namespace is the name of the app. Happy to change things if there are more appropriate values

src/CLI/CLI.csproj
```diff
@@ -0,0 +1,25 @@
...
+
+  <ItemGroup>
+    <PackageReference Include="CommandLineParser" Version="2.7.82" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Automation\Automation.csproj" />
+  </ItemGroup>
+
+  <Target Name="WriteAxeWindowsVersionFile">
```
RobGallo: Why did this need to be added here? This should have been written already by the first project to be compiled.

src/CLI/CLI.csproj
```diff
@@ -0,0 +1,25 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <AssemblyName>AxeWindowsScanner</AssemblyName>
```
RobGallo: How about naming this AxeWindowsCLI? It seems more informative than "scanner".

src/CLI/ErrorCollector.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace AxeWindowsScanner
+{
+    public class ErrorCollector : IErrorCollector
+    {
+        public IReadOnlyList<string> ParameterErrors => _parameterErrors;
+
+        public IReadOnlyList<string> ScanErrors => _scanErrors;
```
RobGallo: I notice this property is not defined in the IErrorCollector interface. I'm not sure if that's intentional.

src/CLI/CLI.csproj
DaveTryon: Copy/Paste error. Now removed.

src/CLI/CLI.csproj
DaveTryon: Done

src/CLI/ErrorCollector.cs
DaveTryon: Nice catch--this was an early design thought that I later abandoned (but failed to clean up completely)

src/CLI/Options.cs
```diff
@@ -0,0 +1,69 @@
...
+                }
+                else
+                {
+                    ErrorCollector.AddParameterError("Invalid verbosity level: " + verbosity);
+                }
+            }
+
+            if (processId != 0)
+            {
+                _processId = processId;
```
RobGallo: Thought: Should the options class be doing this? I kind of feel like the options class should be meant to simply contain the options and not have much program logic.

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+        public OutputGenerator(TextWriter writer)
+        {
+            if (writer == null) throw new ArgumentNullException(nameof(writer));
+
+            _writer = writer;
+        }
+
+        public void ShowOutput(IOptions options, IErrorCollector errorCollector, ScanResults scanResults)
+        {
+            bool failedToComplete = errorCollector.ParameterErrors.Any() || errorCollector.Exceptions.Any() || (scanResults == null);
```
RobGallo: Could this logic be part of the ErrorCollector? It might make the code a little more succinct.

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+                ShowScanResults(options, scanResults);
+            }
+        }
+
+        public void ShowBanner(IOptions options)
+        {
+            ShowBanner(options, VerbosityLevel.Default);
+        }
+
+        private void ShowBanner(IOptions options, VerbosityLevel miniumVerbosity)
```
RobGallo: "minium"

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+            if (writer == null) throw new ArgumentNullException(nameof(writer));
+
+            _writer = writer;
+        }
+
+        public void ShowOutput(IOptions options, IErrorCollector errorCollector, ScanResults scanResults)
+        {
+            bool failedToComplete = errorCollector.ParameterErrors.Any() || errorCollector.Exceptions.Any() || (scanResults == null);
+
+            ShowBanner(options, failedToComplete ? VerbosityLevel.Quiet : VerbosityLevel.Default);
```
RobGallo: Why is the banner dependent on the verbosity? Isn't this controlled by the options?

RobGallo: Come to think of it, why is the banner shown here? Why is the banner not shown before anything else has run?

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        private void ShowScanResults(IOptions options, ScanResults scanResults)
+        {
+            if (options.VerbosityLevel == VerbosityLevel.Quiet)
+            {
+                return;
+            }
+
+            if (scanResults.ErrorCount == 1)
```
RobGallo: Perhaps extract to a function called WriteErrorCount?

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+                _writer.WriteLine("Results were written to \"{0}\"", scanResults.OutputFile.A11yTest);
+            }
+        }
+
+        private void ShowVerboseResults(ScanResults scanResults)
+        {
+            int errorCount = 0;
+            foreach (ScanResult scanResult in scanResults.Errors)
+            {
+                _writer.WriteLine("Error {0}: {1}", ++errorCount, scanResult.Rule.Description);
```
RobGallo: Does writing an error index provide value?

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+                if (scanResult.Element.Properties != null && scanResult.Element.Properties.Any())
```
RobGallo: Extract to a function called WriteElementProperties? nested for loops... 👃

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+                {
+                    _writer.WriteLine("  Element Properties:");
+                    foreach (KeyValuePair<string, string> pair in scanResult.Element.Properties)
+                    {
+                        _writer.WriteLine("    {0} = {1}", pair.Key, pair.Value);
+                    }
+                }
+                if (scanResult.Element.Patterns != null && scanResult.Element.Patterns.Any())
```
RobGallo: I would extract writing patterns to a function as well.

src/CLI/OutputGenerator.cs
```diff
@@ -0,0 +1,152 @@
...
+        private bool _bannerHasBeenShown;
+
+        public OutputGenerator(TextWriter writer)
+        {
+            if (writer == null) throw new ArgumentNullException(nameof(writer));
+
+            _writer = writer;
+        }
+
+        public void ShowOutput(IOptions options, IErrorCollector errorCollector, ScanResults scanResults)
```
RobGallo: I think it might be more accurate to call all the functions in this class which start with "Show, "Write".

src/CLI/ProcessHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        public int FindProcessByName(string processName)
+        {
+            var processes = _processAbstraction.GetProcessesByName(processName);
+
+            if (processes == null || processes.Length == 0)
+            {
+                _errorCollector.AddParameterError("Unable to find process with name " + processName);
+                return IProcessHelper.InvalidProcessId;
```
RobGallo: I realize this is a little easier, but I think throwing an exception might make the code a better. Dr. Bob recommends throwing exceptions rather than returning "magic" numbers. (Feel free to push back.)

src/CLI/OutputGenerator.cs
DaveTryon: The banner is omitted in quiet mode if execution completes normally. In non-quiet modes, we call ShowBanner immediately (to provide feedback that the app is doing something). We call ShowBanner again if we failed to complete, but we can't do that until after the app has run. This is to balance both scenarios.

src/CLI/OutputGenerator.cs
DaveTryon: fixed

src/CLI/OutputGenerator.cs
DaveTryon: I think it does if you're trying to refer to a specific issue in the output. It's basically free to include, and I can't think of any potential problem that it would represent

src/CLI/OutputGenerator.cs
DaveTryon: Done

src/CLI/ReturnValueChooser.cs
```diff
@@ -0,0 +1,30 @@
...
+        public const int ScanCompletedAndFoundErrors = 1;
+        public const int ScanFailedToComplete = 2;
+        public const int BadInputParameters = 255;
+
+        public static int GetReturnValue(IErrorCollector errorCollector, ScanResults scanResults)
+        {
+            if (errorCollector.ParameterErrors.Any())
+                return BadInputParameters;
+
+            if (errorCollector.Exceptions.Any() || scanResults == null)
```
RobGallo: This condition could probably be a property or method on ErrorCollector.

src/CLI/OutputGenerator.cs
DaveTryon: Done

src/CLI/ErrorCollector.cs
```diff
@@ -0,0 +1,39 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace AxeWindowsScanner
+{
+    public class ErrorCollector : IErrorCollector
```
RobGallo: It doesn't much matter because this is an exe and not a library, but is there any reason this class is public and not internal? Our typical pattern is to make classes like this internal.

src/CLI/OutputGenerator.cs
DaveTryon: Done

src/CLI/OutputGenerator.cs
DaveTryon: Done

src/CLITests/ErrorCollectorUnitTests.cs
```diff
@@ -0,0 +1,114 @@
...
+
+        IErrorCollector _errorCollector;
+
+        [TestInitialize]
+        public void BeforeEachTest()
+        {
+            _errorCollector = new ErrorCollector();
+        }
+
+        private void ValidateListSizes(int expectedParameterErrors = 0, int expectedExceptions = 0)
```
RobGallo: I'm not a fan of default parameter values because it requires me to remembe what the defaults are, or that there _are_ defaults at all. Even after reading this function, I still found it to be a speed bump when I encountered the method call with no parameters.

src/CLITests/ErrorCollectorUnitTests.cs
```diff
@@ -0,0 +1,114 @@
...
+        [TestMethod]
+        [Timeout(1000)]
+        public void AddParameterError_MutipleValidValues_StoresValuesCorrectly()
+        {
+            foreach (string value in ValidValueArray)
+            {
+                _errorCollector.AddParameterError(value);
+            }
+
+            for (int loop = 0; loop < ValidValueArray.Length; loop++)
```
RobGallo: I'm curious about the use of "loop" instead of 'i'.

src/CLI/ProcessHelper.cs
DaveTryon: Dr. Dave generally concurs, but this patient makes it difficult. This code gets called by the ctor for the Options class, which is called by CommandLineParser (not our code). CommandLineParser eats all Exceptions without allowing us to access the original information, so the final user experience is unpleasant. To get around that, I defined this interface to not throw (for the sake of messaging), which meant that it needed to return _something_ recognizable.

src/CLI/ReturnValueChooser.cs
DaveTryon: ls `errorCollector.AnyExceptions()` significantly clearer or more succinct than `errorCollector.Exceptions.Any()` ?

src/CLI/Options.cs
DaveTryon: Options knows about the input parameters, then the rest of the program treats the options as immutable. I went back and forth a couple of times on how to best use this. Happy to talk offline...

src/CLI/OutputGenerator.cs
DaveTryon: Wrapped this in IErrorCollector.Any

src/CLITests/OutputGeneratorUnitTests.cs
```diff
@@ -0,0 +1,670 @@
...
+        {
+            _optionsMock.Setup(x => x.VerbosityLevel).Returns(verbosityLevel);
+            _optionsMock.Setup(x => x.ProcessName).Returns(processName);
+            _optionsMock.Setup(x => x.ProcessId).Returns(processId);
+            _optionsMock.Setup(x => x.ScanId).Returns(scanId);
+        }
+
+        [TestMethod]
+        [Timeout(1000)]
+        public void Ctor_WriterIsNull_ThrowsArgumnentNullException()
```
RobGallo: "Argumnent"

src/CLITests/ProcessHelperUnitTests.cs
```diff
@@ -0,0 +1,159 @@
...
+using AxeWindowsCLI;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Moq;
+using System;
+using System.Diagnostics;
+
+namespace CLITests
+{
+    [TestClass]
+    public class ProcessHelperUnitTests
```
RobGallo: I think this class is missing a successful check of GetProcessById.

src/CLI/OutputGenerator.cs
RobGallo: "We call ShowBanner again..." How come?

src/CLI/OutputGenerator.cs
RobGallo: Maybe eventually we could have a parameter for it, but I don't think we need to do that now.

src/CLI/ReturnValueChooser.cs
RobGallo: Nevermind on this one. I think I read the code slightly wrong. :)

src/CLITests/ErrorCollectorUnitTests.cs
DaveTryon: This class is going away, but I find defaults to builders or validators extremely useful in unit tests

src/CLITests/ErrorCollectorUnitTests.cs
DaveTryon: Personal preference, I guess. I find "i" much less descriptive than "loop" or "index" (depending on the specific circumstance)

src/CLITests/OutputGeneratorUnitTests.cs
DaveTryon: Fixed

src/CLITests/ProcessHelperUnitTests.cs
DaveTryon: Added

src/CLI/OutputGenerator.cs
DaveTryon: Discussed offline

src/CLI/ErrorCollector.cs
DaveTryon: Changed. Options needs to remain public (for CommandLineParser), and VerbosityLevel needs to be remain public (since it's referenced by Options)

PR #287, opened 2020-01-31
Add an MSI to install the Axe.Windows CLI on a user's machine

DaveTryon:
#### Describe the change

Create an MSI to install the CLI onto a user's machine. We opted for an MSI to eventually enable UIAccess support. The images in the UI are the same used in AIWin's installer. The EULA is the AIWin EULA with the product name updated. Screenshots of the installer:

image

image

image

image

image

**Control Panel after install:**

image

@ferBonnin, would you please proof the visuals--you're probably most tied into what our current guidelines are with respect to EULA, etc. 


ferBonnin:
a few comments:

- are we using the same EULA from Accessibility insights for Windows? if no, can I get the full EULA we are adding there?
- We should spell out at least the first time what CLI means in the first setup screen
- CC @adarshrema, this will need a quick accessibility test pass since there is some UI.
- CC @adarshrema. can we get numbers on how many downloads?

DaveTryon:
@ferBonnin, I've updated the product name and the corresponding screenshots, as suggested. The complete EULA can be found in the PR itself: https://github.com/microsoft/axe-windows/blob/840a3333712a48142ab144f659611ae9a5d4d844/src/MSI/Resources/eula.rtf

As for downloads, AIWin currently gets download counts directly from GitHub. I would anticipate that the same mechanism could be used to give us a reasonable feel for download counts.

PR #291, opened 2020-02-03
Add unit tests to check old a11ytest file version compatibility

RobGallo:
#### Describe the change

Changing the OldFileVersionCompatibilityTests project so that it is a console app rather than a test app.

The problem was that since we changed the project to .Net Core, running all three unit tests which checked for old file version compatibility caused all three tests to appear not to run. The investigation into the problem determined there was a missing dll: Microsoft.IntelliTrace.Core. The trail went cold at that point: there is no documentation on this problem except that it exists and that it should have been fixed in the latest version of Visual Studio and .Net Core, although there are examples of users who still experience the problem with the latest versions just like us.

Since the tests run and pass individually, the solution (implemented herein) is to run each test by calling a console app and specifying the particular test function to run.

It isn't pretty, but there it is.

### DIFF
--- a/src/OldFileVersionCompatibilityTests/TestAssertionException.cs
+++ b/src/OldFileVersionCompatibilityTests/TestAssertionException.cs
@@ -1,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using System.Diagnostics;
+
+namespace OldFileVersionCompatibilityTests
+{
+    class TestAssertionException : Exception
+    {
+        public string FileName { get; }
+        public int LineNumber { get; }
+
+        public TestAssertionException(string message)
+            : base(message)
+        {
+            var trace = new StackTrace(true);
+            var frame = trace.GetFrame(2);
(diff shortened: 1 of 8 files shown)


src/OldFileVersionCompatibilityTests/TestAssertionException.cs
```diff
@@ -0,0 +1,24 @@
...
+    class TestAssertionException : Exception
+    {
+        public string FileName { get; }
+        public int LineNumber { get; }
+
+        public TestAssertionException(string message)
+            : base(message)
+        {
+            var trace = new StackTrace(true);
+            var frame = trace.GetFrame(2);
```
DaveTryon: This parameter of 2 seems a bit sketchy, since it assumes knowledge about the caller. Any chance we could make it a parameter of the ctor?

RobGallo: I think that's a good point. But I argue that the class is part of a closed system where it has a prescribed purpose where the exception is intended always to be created from an assert function. And so far, all the assert functions are only one level deep. And I have no expectation they will go deeper.

PR #292, opened 2020-02-04
Add command line support to display third party notices in default browser

DaveTryon:
#### Describe the change

This change adds the `--ShowThirdPartyNotices` command line parameter to the CLI, so that we have a way for users to fulfill licensing requirements. When asked, the CLI will open `thirdpartynotices.html` in the user's default browser. This file is auto-generated, and will be refreshed in a separate PR.

One noteworthy change to call out--we used to use the `Target` group in the commands that we passed to CommandLineParser. It didn't seem to make sense to include `--ShowThirdPartyNotices` in the `Target` group, so I removed the group. This means that the `OptionsEvaluator` class now has the responsibility to ensure that a target is specified.

Since CommandLineParser no longer requires input parameters, we lose the automatic display of parameters if the CLI is run without arguments. To cover this case, the CLI program detects empty input arguments and explicitly passes the `--help` parameter to CommandLineParser. It's something of a hack, but it's extremely non-invasive (a single line of code in the outermost method of the app).


src/CLITests/OptionsEvaluatorUnitTests.cs
```diff
@@ -135,28 +138,30 @@ public void ProcessInputs_SpecifiesInvalidVerbosity_ThrowsParameterException()
...
             ValidateOptions(OptionsEvaluator.ProcessInputs(input, _processHelperMock.Object),
-                processName: null, processId: TestProcessId, verbosityLevel: VerbosityLevel.Quiet);
+                processId: TestProcessId, verbosityLevel: VerbosityLevel.Quiet);
             VerifyAllMocks();
         }
 
         [TestMethod]
         [Timeout(1000)]
-        public void ProcessInputs_SpecifiesDefault_SetsQuietVerbosity()
+        public void ProcessInputs_SpecifiesDefault_SetsDefaultVerbosity()
```
DaveTryon: Test was misnamed before

src/CLITests/OptionsUnitTests.cs
```diff
@@ -36,41 +35,20 @@ private int FailIfCalled(IEnumerable<Error> errors)
...
-            List<Error> errorList = errors.ToList<Error>();
-            Assert.AreEqual(1, errorList.Count);
-            Assert.AreEqual(expectedError, errorList[0].Tag);
-            return ExpectedParseFailure;
-        }
-
         private int ValidateOptions(Options options, string processName = null,
             int processId = 0, string outputDirectory = null, string scanId = null,
-            string verbosity = null)
+            string verbosity = null, bool showThirdPartyNotices = false)
```
DaveTryon: This lets us leave the mock for ShowThirdPartyNotices unprogrammed in the couple of cases that need it that way

src/CLITests/OutputGeneratorUnitTests.cs
```diff
@@ -56,12 +57,17 @@ private void VerifyAllMocks()
 
         private void SetOptions(VerbosityLevel verbosityLevel = VerbosityLevel.Default,
             string processName = null, int processId = -1,
-            string scanId = null)
+            string scanId = null, bool setScanId = true,
```
DaveTryon: This lets us leave the mock for ShowThirdPartyNotices unprogrammed in the couple of cases that need it that way

src/CLITests/OutputGeneratorUnitTests.cs
```diff
@@ -365,7 +392,7 @@ public void WriteOutput_ScanResultsMultipleErrors_VerbosityIsDefault_WritesBanne
         }
 
         [TestMethod]
-        //[Timeout(1000)]
```
DaveTryon: This shouldn't have been commented out before :(

src/CLI/BrowserAbstraction.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Diagnostics;
+
+namespace AxeWindowsCLI
+{
+    internal class BrowserAbstraction : IBrowserAbstraction
+    {
+        public void Open(string pathToFile)
+        {
+            Process.Start(new ProcessStartInfo("cmd", $"/c \"" + pathToFile + "\"")
```
DaveTryon: This command line has been validated on Win10 after installing via the MSI

PR #297, opened 2020-02-07
Build zipped form of AxeWindowsCLI

DaveTryon:
#### Describe the change

We have an MSI installer for installing the CLI, but there are likely to be automation scenarios where people want to avoid the hassle of an MSI. To support this scenario, we will allow people to access the CLI as a ZIP file. For a given release of Axe.Windows, we'll probably publish both the MSI and ZIP formats, so people can use whichever best fits their needs.

This PR adds a PowerShell script to generate the ZIP file. The script is launched after successfully building the CLI project. Since not all dev machines are configured to allow PowerShell scripts to run automatically, we put the execution of the script behind a new environment variable: `CreateAxeWindowsZippedCLI`. This environment variable is set in the PR and signed builds, so the ZIP file will be available as an artifact for publishing.

To aid in debugging, the debug version of AxeWindowsCLI.zip also includes PDB files. We probably won't publish this one, but we'll have it available for our use if we need it.

### DIFF
--- a/tools/scripts/BuildZippedCLI.ps1
+++ b/tools/scripts/BuildZippedCLI.ps1
@@ -20,6 +20,9 @@ param(
     [Parameter(Mandatory=$true)][String]$TargetDir
 )
 
+Set-StrictMode -Version Latest
+$script:ErrorActionPreference = 'Stop'
+
 # Uncomment the next line for debugging
 #$VerbosePreference='continue'
 
@@ -72,9 +75,9 @@ function Create-Zipfile($src, $app, $zipFile, [string[]]$extensionsToPrune, [str
     Remove-Item $scratch -Force -Recurse
 }
 
-function CreateCLIZip($confguration, $targetDir, $zipFileName){
+function CreateCLIZip($configuration, $targetDir, $zipFileName){
     Write-Host "Creating a CLI zip file from files in $targetDir"
-    if ($confguration -eq 'debug') {
+    if ($configuration -eq 'debug') {
         $extensionsToPrune=@('dev.json')
     } else {
         $extensionsToPrune=@('pdb', 'dev.json')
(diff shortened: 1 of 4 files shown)


tools/scripts/BuildZippedCLI.ps1
```diff
@@ -0,0 +1,99 @@
...
+.Example Usage
+.\BuildZippedCLI.ps1 -Configuration release -TargetDir c:\myrepo\src\CLI\bin\release\netcoreapp3.0
+#>
+
+param(
+    $Configuration,
+    $TargetDir
+)
+
+# Uncomment the next line for debugging
```
dbjorge: I recommend using the following boilerplate at the top of any powershell script; it will catch things like typos in variable names and fail fast rather than silently if non-terminating errors you aren't expecting get emitted:

```powershell
Set-StrictMode -Version Latest
$script:ErrorActionPreference = 'Stop'
```

tools/scripts/BuildZippedCLI.ps1
```diff
@@ -0,0 +1,99 @@
...
+    }
+    foreach ($folder in $foldersToPrune) {
+        Remove-FoldersByName $scratch $folder
+    }
+    Create-Archive $scratch $zipFile
+
+    Remove-Item $scratch -Force -Recurse
+}
+
+function CreateCLIZip($confguration, $targetDir, $zipFileName){
```
dbjorge: sp: configuration

tools/scripts/BuildZippedCLI.ps1
DaveTryon: Thanks--added

tools/scripts/BuildZippedCLI.ps1
DaveTryon: Corrected

PR #303, opened 2020-02-11
Add support for targeting UWP apps

DaveTryon:
#### Describe the change

While experimenting with UWP apps, I found apps where the process name is ".App" instead of ".exe", and that's exactly how it shows up in the list of processes. To allow targeting of these apps. update the process name targeting to only remove the extension if it's ".exe". Add unit tests to cover the various code paths.

### DIFF
--- a/src/CLI/OptionsEvaluator.cs
+++ b/src/CLI/OptionsEvaluator.cs
@@ -22,13 +22,8 @@ namespace AxeWindowsCLI
             }
             else if (!string.IsNullOrEmpty(processName))
             {
-                string p = Path.GetFileName(processName);
-                if (p.EndsWith(".exe", StringComparison.OrdinalIgnoreCase))
-                {
-                    p = Path.GetFileNameWithoutExtension(p);
-                }
-                processName = p;
-                processId = processHelper.ProcessIdFromName(p);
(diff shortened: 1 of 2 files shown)


src/CLI/OptionsEvaluator.cs
```diff
@@ -22,7 +22,11 @@ public static IOptions ProcessInputs(Options rawInputs, IProcessHelper processHe
             }
             else if (!string.IsNullOrEmpty(processName))
             {
-                string p = Path.GetFileNameWithoutExtension(processName);
+                string p = Path.GetFileName(processName);
```
RobGallo: How about breaking the new logic out into a function?

DaveTryon: Done

src/CLI/OptionsEvaluator.cs
```diff
@@ -56,5 +55,15 @@ public static IOptions ProcessInputs(Options rawInputs, IProcessHelper processHe
                 VerbosityLevel = verbosityLevel,
             };
         }
+
+        private static string GetReducedProcessName(string processName)
```
RobGallo: Since "Reduced" means remove exe, "TrimDotExe" might be a better name for the function.

PR #314, opened 2020-02-19
Create self-contained CLI in zip format

DaveTryon:
#### Describe the change

The current CLI requires the presence of the .NET Core 3.0 runtime, which works well for most development systems, but often isn't installed in automated test systems. To better support automated test systems, we're creating a standalone version of the CLI that includes private copies of all of the requires runtime assemblies. This install is large (about 240 files and more than 60MB of disk space), so we generate it as a zip file that can hopefully be easily integrated into existing test frameworks. Other packaging formats can be added in the future, based on customer request.


src/CLI_Full/BuildNotes.md
```diff
@@ -0,0 +1,5 @@
+# Build notes for CLI_Full project
+
+The `CLI_Full` project exists to gather the CLI into a self-contained format where all necessary .NET Core runtime files are gathered by the compiler. We compile the code from the `CLI` project is used, but specify extra flags to create the self-contained version. We end up two versions of `AxeWindowsCLI.exe`, which are not interchangeable due to compiler-generated differences inside the binaries.
```
RobGallo:
```suggestion
The `CLI_Full` project exists to gather the CLI into a self-contained format (.zip) where all necessary .NET Core runtime files are gathered by the compiler. We compile the code from the `CLI` project is used, but specify extra flags to create the self-contained version. We end up with two versions of `AxeWindowsCLI.exe`, which are not interchangeable due to compiler-generated differences inside the binaries.
```

src/CLI_Full/BuildNotes.md
```diff
@@ -0,0 +1,5 @@
...
+
+Changes that impact _common functionality_ need to be made in the `CLI` project. changes that impact the generation of the self-contained version need to be made in teh `CLI_Full` project.
```
RobGallo:
```suggestion
Changes that impact _common functionality_ need to be made in the `CLI` project. changes that impact the generation of the self-contained version need to be made in the `CLI_Full` project.
```

PR #321, opened 2020-03-09
Fix Issue #320 - AI-Win not properly reporting text pattern on Win32 up down control

RobGallo:
#### Describe the change

The problem was that the event listener code registered for focus events on the CUIAutomation object, as opposed to the CUIAutomation8 object. We don't know exactly why CUIAutomation.PoleForPotentialSupportedPatterns  does not provide the text pattern as a supported pattern on the Win32 edit field. But CUIAutomation8 does and the text pattern is retrieved successfully.

### DIFF
--- a/src/Desktop/UIAutomation/EventHandlers/EventListenerFactory.cs
+++ b/src/Desktop/UIAutomation/EventHandlers/EventListenerFactory.cs
@@ -349,7 +349,8 @@ namespace Axe.Windows.Desktop.UIAutomation.EventHandlers
                     case EventType.UIA_AutomationFocusChangedEventId:
                         if (this.EventListenerFocusChanged == null)
                         {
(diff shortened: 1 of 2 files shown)


src/Desktop/UIAutomation/EventHandlers/EventListenerFactory.cs
```diff
@@ -349,7 +349,8 @@ private void RegisterEventListener(EventListenerFactoryMessage msgData)
                     case EventType.UIA_AutomationFocusChangedEventId:
                         if (this.EventListenerFocusChanged == null)
                         {
-                            this.EventListenerFocusChanged = new FocusChangedEventListener(this.UIAutomation, msgData.Listener);
+                            var uia = (IUIAutomation) UIAutomation8 ?? UIAutomation;
```
DaveTryon:
```suggestion
                            IUIAutomation uia = UIAutomation8 ?? UIAutomation;
```

Issue #322, opened 2020-03-09
[BUG] “An element's BoundingRectangle must not obscure its container element” should exclude dialogs


RobGallo:
## Description

For the rule “An element's BoundingRectangle must not obscure its container," we should make an exception for element's with the type of "Pane" and with the "IsDialog" property set to true.

The rule already contains an exemption for elements of type "Window" because windows may obscure their container elements. The same reasoning applies to dialog boxes because they too may obscure their containers.

element.”

RobGallo:
Fixed in 0.9.1.

PR #323, opened 2020-03-09
Fix Issue #322 - [BUG] “An element's BoundingRectangle must not obscure its container element” should exclude dialogs

RobGallo:
#### Describe the change

- Add the BoolProperties.IsDialog condition
- Update BoundingRectangleCompletelyObscuresContainer.CreateCondition to exempt panes which are dialogs
- Add unit tests
- Update Rules project to generate source files when .resx files change


src/RulesTest/Library/BoundingRectangleCompletelyObscuresContainerTest.cs
```diff
@@ -1,11 +1,12 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Types;
 using System.Drawing;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Axe.Windows.Rules.PropertyConditions;
```
DaveTryon: Can we sort these using statements alphabetically while we're here?
```suggestion
using Axe.Windows.Core.Bases;
using Axe.Windows.Core.Types;
using Axe.Windows.Rules.PropertyConditions;
using Microsoft.VisualStudio.TestTools.UnitTesting;
using System.Drawing;
```

src/RulesTest/Library/BoundingRectangleCompletelyObscuresContainerTest.cs
```diff
@@ -165,5 +166,31 @@ public void BoundingRectangleCompletelyObscuresContainer_BottomFail()
             Assert.IsTrue(Rule.Condition.Matches(e));
             Assert.AreEqual(EvaluationCode.Error, Rule.Evaluate(e));
         }
+
+        [TestMethod]
+        public void BoundingRectangleCompletelyObscuresContainer_IsDialog_NotApplicable()
```
DaveTryon: Should there also be a test that ensures that a Window is excluded?

src/Rules/Library/BoundingRectangleCompletelyObscuresContainer.cs
```diff
@@ -34,7 +34,10 @@ protected override Condition CreateCondition()
         {
             // Windows can be any size, regardless of their parents
```
DaveTryon: This comment now feels "detached" from the check. Should it be closer to the return statement, and possibly updated to reference dialogs?

src/Rules/Rules.csproj
```diff
@@ -35,4 +35,79 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup>
```
DaveTryon: Are the changes in rules.csproj necessary? It seems like things were working without them before...

src/RulesTest/PropertyConditions/BoolPropertiesTest.cs
```diff
@@ -315,5 +315,36 @@ public void TestIsControlElementDoesNotExistFalse()
                 Assert.IsFalse(IsControlElementDoesNotExist.Matches(e));
             } // using
         }
+
+        [TestMethod]
+        public void IsDialog_true()
```
DaveTryon: We've shifted the namign convention of these tests. If we use the existing convention, this might be
```suggestion
        public void IsDialogTrue()
```

DaveTryon: Naming conventions are a bit different in the new methods :(

src/Rules/Rules.csproj
RobGallo: They weren't causing any problems before because the strings in the resx file hadn't changed since the conversion to .Net Standard. Now that the strings have been updated, if I didn't make this change, the *.designer.cs files wouldn't be updated. And thus the code wouldn't compile.

src/RulesTest/PropertyConditions/BoolPropertiesTest.cs
RobGallo: Yeah, I thought about that, but this is the convention we try to use in newer tests. I don't like the mismatch either, but it doesn't bother me enough to change all the tests. Thoughts?

src/RulesTest/Library/BoundingRectangleCompletelyObscuresContainerTest.cs
RobGallo: Done.

src/Rules/Library/BoundingRectangleCompletelyObscuresContainer.cs
RobGallo: Updated the comment.

src/RulesTest/Library/BoundingRectangleCompletelyObscuresContainerTest.cs
RobGallo: Added.

src/RulesTest/PropertyConditions/BoolPropertiesTest.cs
DaveTryon: Looks like other tests are using upper case after each underscore. If we do that, we get:
```suggestion
        public void IsDialog_True()
```

src/RulesTest/PropertyConditions/BoolPropertiesTest.cs
```diff
@@ -315,5 +315,36 @@ public void TestIsControlElementDoesNotExistFalse()
...
+            using (var e = new MockA11yElement())
+            {
+                var a11yProperty = new A11yProperty(PropertyType.UIA_IsDialogPropertyId, false);
+                e.Properties.Add(PropertyType.UIA_IsDialogPropertyId, a11yProperty);
+                Assert.IsFalse(IsDialog.Matches(e));
+            } // using
+        }
+
+        [TestMethod]
+        public void IsDialog_NotSet_false()
```
DaveTryon:
```suggestion
        public void IsDialog_NotSet_False()
```

src/RulesTest/PropertyConditions/BoolPropertiesTest.cs
```diff
@@ -315,5 +315,36 @@ public void TestIsControlElementDoesNotExistFalse()
...
+            using (var e = new MockA11yElement())
+            {
+                var a11yProperty = new A11yProperty(PropertyType.UIA_IsDialogPropertyId, true);
+                e.Properties.Add(PropertyType.UIA_IsDialogPropertyId, a11yProperty);
+                Assert.IsTrue(IsDialog.Matches(e));
+            } // using
+        }
+
+        [TestMethod]
+        public void IsDialog_false()
```
DaveTryon:
```suggestion
        public void IsDialog_False()
```

src/Rules/Rules.csproj
DaveTryon: We also have resx files in Actions, Automation, Core, Desktop, and RuleSelection. Are those projects currently in a state where changes wouldn't get picked up?

RobGallo: I will check after this PR.

Issue #161, opened 2019-10-04
[Feature Request] Add rule to ensure the descendants of off screen elements are also off screen.


RobGallo:
### new rule

Based on https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.automationelement.automationelementinformation.isoffscreen?view=netframework-4.8:

“When the value is true for a container, it is also true for the container element's descendants.”

We should add a rule which validates this policy.

ferBonnin:
Reviewed this with PMs and Leads, this looks like a good suggestion, but we need to schedule time to do this feature @RobGallo  is this a 2-day thing or a 2-week thing?

RobGallo:
@ferBonnin, It's more of a two day thing.

RobGallo:
Fixed in 0.9.1.

PR #326, opened 2020-03-13
Fix Issue #161 - [Feature Request] Add rule to ensure the descendants of off screen elements are also off screen.

RobGallo:
#### Describe the change

The evaluation strategy is to  test each element against its parent. This avoids starting with a root element and testing all its descendants. The latter would be very inefficient because the test would be re-run for each descendant of the root.

Also, the chosen strategy has the added benefit of failing while on the element with the mismatched IsOffscreen property value, which means it is easier to point users to the element with the problem. This would have been much more difficult if the non matching element were a deep descendant of the element being tested.

### DIFF
--- a/src/RulesTest/Library/IsOffScreenChild.cs
+++ b/src/RulesTest/Library/IsOffScreenChild.cs
@@ -1,0 +1,65 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using EvaluationCode = Axe.Windows.Rules.EvaluationCode;
+
+namespace Axe.Windows.RulesTest.Library
+{
+    [TestClass]
+    public class IsOffScreenChildTests
+    {
+        private static Axe.Windows.Rules.IRule Rule = new Axe.Windows.Rules.Library.IsOffScreenChild();
+
+        [TestMethod]
+        public void IsOffScreenChild_Condition_Matches()
+        {
+            using (var e = new MockA11yElement())
+            {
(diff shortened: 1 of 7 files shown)


src/RulesTest/Library/IsOffScreenChild.cs
```diff
@@ -0,0 +1,65 @@
...
+            } // using
+        }
+
+        [TestMethod]
+        public void IsOffScreenChild_Evaluate_Error()
+        {
+            using (var e = new MockA11yElement())
+            using (var parent = new MockA11yElement())
+            {
+                parent.IsOffScreen = true;
```
DaveTryon: Should we explicitly set e.IsOffScreen, rather than rely on the default?

src/RulesTest/Library/IsOffScreenChild.cs
```diff
@@ -0,0 +1,65 @@
...
+            } // using
+        }
+
+        [TestMethod]
+        public void IsOffScreenChild_Evaluate_Pass()
+        {
+            using (var e = new MockA11yElement())
+            using (var parent = new MockA11yElement())
+            {
+                parent.IsOffScreen = false;
```
DaveTryon: Should we explicitly set e.IsOffScreen, rather than rely on the default?

src/RulesTest/Library/IsOffScreenChild.cs
RobGallo: The value for e is never checked in the call to Evaluate.

src/RulesTest/Library/IsOffScreenChild.cs
RobGallo: The value for e.IsOffScreen is never checked in the call to Evaluate.

PR #334, opened 2020-03-25
Restore netstandard20 references in test project

DaveTryon:
#### Describe the change

Since updating to newer assemblies, PR builds are generating the following error during build:

##[warning]EXEC(0,0): Warning : NU5128: Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location.

This change restores the netstandard20 references that were there before the changes.


DaveTryon:
This seems to be the wrong approach. Closing and will create a separate PR

PR #350, opened 2020-04-08
Remove hardcoded calls to System.Diagnostics.Debugger.Break

DaveTryon:
#### Describe the change

There are 5 places in the code where Axe.Windows is hardcoded to call Debugger.Break. These were added to provide a way to catch unexpected scenarios in Axe.Windows, but they are a real nuisance when trying to debug an app that consumes Axe.Windows (and where the exceptions aren't really that interesting). Of the 5 places where Debugger.Break was called:
- 2 were in unused methods
- 2 were in a block where an exception is reported (which means that the exception will be detectable in telemetry)
- 1 was in a scenario that telemetry suggested never occurred

### DIFF
--- a/src/Core/Misc/ExtensionMethods.cs
+++ b/src/Core/Misc/ExtensionMethods.cs
@@ -420,11 +420,7 @@ namespace Axe.Windows.Core.Misc
                 }
                 else if (HasTestResults(tss, ScanStatus.NoResult))
                 {
-                    // it means that there was no rule executed or 
-                    if (System.Diagnostics.Debugger.IsAttached)
-                    {
-                         System.Diagnostics.Debugger.Break();
-                    }
+                    return ScanStatus.NoResult;
                 }
 
                 return ScanStatus.Pass;
--- a/src/Core/Misc/ExtensionMethods.cs
+++ b/src/Core/Misc/ExtensionMethods.cs
@@ -418,10 +418,6 @@ namespace Axe.Windows.Core.Misc
                 {
                     return ScanStatus.Uncertain;
                 }
-                else if (HasTestResults(tss, ScanStatus.NoResult))
-                {
-                    return ScanStatus.NoResult;
-                }
 
                 return ScanStatus.Pass;
             }
--- a/src/Rules/Conditions/Condition.cs
+++ b/src/Rules/Conditions/Condition.cs
@@ -74,5 +74,8 @@ namespace Axe.Windows.Rules
 
         public static Condition True = Create(e => true, ConditionDescriptions.True);
         public static Condition False = Create(e => false, ConditionDescriptions.False);
+#if DEBUG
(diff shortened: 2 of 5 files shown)


src/Core/Misc/ExtensionMethods.cs
```diff
@@ -420,11 +420,7 @@ public static ScanStatus GetAggregatedScanStatus(this IEnumerable<ScanStatus> ts
                 }
                 else if (HasTestResults(tss, ScanStatus.NoResult))
                 {
-                    // it means that there was no rule executed or 
-                    if (System.Diagnostics.Debugger.IsAttached)
-                    {
-                         System.Diagnostics.Debugger.Break();
-                    }
+                    return ScanStatus.NoResult;
```
DaveTryon, line 423: @RobGallo, did we decide to return ScanStatus.NoResult (changing the behavior) or just to continue to let this return ScanStaus.Pass (keeping the current behavior) ?

RobGallo: I would continue the prior behavior.

src/Rules/Conditions/Condition.cs
```diff
@@ -74,6 +74,5 @@ public static Condition Create(DelegateCondition.MatchesDelegate d, string descr
 
         public static Condition True = Create(e => true, ConditionDescriptions.True);
         public static Condition False = Create(e => false, ConditionDescriptions.False);
-        public static Condition DebugBreak = Create(e => { System.Diagnostics.Debugger.Break(); return true; });
```
RobGallo, line 77: I'm not sure I agree with removing this. It will never get called unintentionally because it's never used. It's only a convenience for debugging rules, which are otherwise difficult to set breakpoints on. On the other hand, if you object to keeping it, it's pretty easy to add back as needed for debugging.

src/RulesTest/MonsterTest.cs
```diff
@@ -1021,15 +1020,5 @@ public void MonsterUserControl()
             var results = Axe.Windows.Rules.Rules.RunAll(e);
             return results.ToDictionary(r => r.RuleInfo.ID, r => r.EvaluationCode);
         }
-
-        private static void WriteResultsToDebugOutput(Dictionary<RuleId, EvaluationCode> results)
```
RobGallo, line 1025: Please don't remove this, I sometimes need it when rules change and the monster tests fail.

src/Rules/Conditions/Condition.cs
DaveTryon: I made it debug only as a compromise--it'll never make it into the NuGet package, and it's still easily accessible for testing and debugging

src/Core/Misc/ExtensionMethods.cs
DaveTryon: Done

src/RulesTest/MonsterTest.cs
DaveTryon: Restored

PR #353, opened 2020-04-13
Fix Issue #352 False-negative result for the rule "An element's IsOffScreen property must be false when its clickable point is on screen."

### LINKED ISSUE #352: [BUG] False-negative result for the rule "An element's IsOffScreen property must be false when its clickable point is on screen."
## Issue

False-negative result for the rule "An element's IsOffScreen property must be false when its clickable point is on screen."

Using WPF .NET 4.6.1 all hidden controls have false-negative results regarding IsOffscreen==True and ClickablePoint. It happens because, in spite of the fact that WPF returns IsOffScreen==True, it also returns some value for ClickablePoint.

According to [MSDN](https://docs.microsoft.com/en-us/dotnet/api/system.windows.automation.automationelement.automationelementinformation.isoffscreen?view=netframework-4.8):
> If the element has a clickable point that can cause it to be focused, it is considered to be onscreen even when a portion of the element is off screen. 

The main thing in the statement above is "that **can cause** it to be focused". But there is no way how an element can be focused if it is hidden. So, I see an incorrect implementation of this rule.

## To Reproduce
```
<Window x:Class="IsOffscreenIssue.MainWindow" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
        xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
        xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" mc:Ignorable="d" Title="MainWindow"
        Height="450" Width="800">
    <StackPanel>
        <Button Content="btn 1" />
        <Button Content="btn 2" Visibility="Collapsed" />
    </StackPanel>
</Window>
```
## Expected behavior

Accessibility Insight tool mustn't report that error for hidden elements.

## Actual behavior

Accessibility Insight tool reports that error for hidden elements.

## Screenshots or .GIF

image

## Desktop

 - OS: Windows 10 1303
 - Accessibility Insights for Windows Version: 1.1.1152.1

## Additional context

Related issue: #162

RobGallo:
#### Describe the change

Fix #352

The rule is meant to ensure element's with valid clickable points are marked as on screen so that ATs can recognize those elements as clickable. In the case given in the bug, the button has a valid clickable point and is not off screen (according to UIA), which is an error. However, the element is also not focusable, which means it cannot be focused or (theoretically) invoked. In this case, the button does not actually appear on screen.

When the element is not keyboard focusable, the rules cannot determine if the element's IsOffScreen property is correct or not. Therefore, the rule will not be run when the element is not keyboard focusable.

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -16,7 +16,7 @@ namespace Axe.Windows.AutomationTests
     public class AutomationIntegrationTests
     {
         // These values should change only in response to intentional rule modifications
-        const int WildlifeManagerKnownErrorCount = 13;
+        const int WildlifeManagerKnownErrorCount = 12;
         const int Win32ControlSamplerKnownErrorCount = 0;
         const int WindowsFormsControlSamplerKnownErrorCount = 6;
         const int WpfControlSamplerKnownErrorCount = 5;
--- a/src/Rules/Library/ClickablePointOnScreen.cs
+++ b/src/Rules/Library/ClickablePointOnScreen.cs
@@ -30,7 +30,7 @@ namespace Axe.Windows.Rules.Library
 
         protected override Condition CreateCondition()
         {
-            return ClickablePoint.OnScreen;
+            return IsKeyboardFocusable & ClickablePoint.OnScreen;
         }
     } // class
 } // namespace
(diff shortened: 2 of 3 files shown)


DaveTryon:
@RobGallo I think you also need to update the expected results from scanning WildlifeManager (since the count went down by 1)

PR #363, opened 2020-04-28
Add FXCopAnalyzers to production projects

RobGallo:
#### Describe the change

The FSCopAnaalyzers were meant to be used in all our production projects, but through some error they were not. The change adds them to production projects and fixes the errors found by the analyzers.

Most errors were of the following types:

- Weak reference in lock
- Methods could be static
- Should catch more specific exception

There is one case where an error check which threw an exception was removed in SnapshotCommand.Execute. This is okay because the call to Scan will immediately check the nullity of the returned value which used to be checked.

The warning CA1303 was suppressed throughout the entirety of the CLI project. The warning flags quoted strings. We have chosen to suppress the warning because the work to address it would only be useful if we planned to localize the project, which we currently do not.

### DIFF
--- a/src/Actions/Actions/SelectAction.cs
+++ b/src/Actions/Actions/SelectAction.cs
@@ -67,7 +67,7 @@ namespace Axe.Windows.Actions
         // Backing object for POIElementContext - is disposed via POIElementContext property
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2213:DisposableFieldsShouldBeDisposed", MessageId = "_POIElementContext")]
         ElementContext _POIElementContext = null;
-        private readonly Object _elementContextLock = new object();
+        private readonly object _elementContextLock = new object();
 
         /// <summary>
         /// Element Context of Point of Interest(a.k.a. selected)
--- a/src/Actions/Actions/SelectAction.cs
+++ b/src/Actions/Actions/SelectAction.cs
@@ -67,6 +67,7 @@ namespace Axe.Windows.Actions
         // Backing object for POIElementContext - is disposed via POIElementContext property
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2213:DisposableFieldsShouldBeDisposed", MessageId = "_POIElementContext")]
         ElementContext _POIElementContext = null;
+        private readonly Object _elementContextLock = new object();
 
         /// <summary>
         /// Element Context of Point of Interest(a.k.a. selected)
@@ -181,7 +182,7 @@ namespace Axe.Windows.Actions
         /// <param name="el"></param>
         public void SetCandidateElement(A11yElement el)
         {
-            lock (this)
+            lock (_elementContextLock)
             {
                 if (el != null)
                 {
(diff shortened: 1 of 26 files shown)


src/Actions/Actions/SelectAction.cs
```diff
@@ -67,6 +67,7 @@ public double IntervalMouseSelector
         // Backing object for POIElementContext - is disposed via POIElementContext property
         [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Usage", "CA2213:DisposableFieldsShouldBeDisposed", MessageId = "_POIElementContext")]
         ElementContext _POIElementContext = null;
+        private readonly Object _elementContextLock = new object();
```
DaveTryon, line 70: >O [](start = 25, length = 1)

Does VS suggest object (with a lower case) here? One's an alias of the other, and I know that VS often suggests the use of the lower case version.

src/Automation/SnapshotCommand.cs
```diff
@@ -31,7 +31,6 @@ public static ScanResults Execute(Config config, IScanTools scanTools)
             scanTools.NativeMethods.SetProcessDPIAware();
 
             var rootElement = scanTools.TargetElementLocator.LocateRootElement(config.ProcessId);
-            if (rootElement == null) throw new InvalidOperationException(nameof(rootElement));
```
DaveTryon, line 34: >rootElement [](start = 16, length = 11)

Why remove this?

src/CLI/CLI.csproj
```diff
@@ -19,6 +19,16 @@
...
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
+    <!-- Suppress the warning (CA1303) for string literals.
+    this warning would only be useful if we decided to localize the project. -->
+    <NoWarn>CA1303;$(NoWarn)</NoWarn>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702,CA1303</NoWarn>
```
DaveTryon, line 29: >1701;1702, [](start = 12, length = 10)

Why exclude these in debug?

src/CLI/Program.cs
```diff
@@ -53,10 +53,15 @@ private int Run()
             Exception caughtException = null;
             try
             {
-                CaseInsensitiveParser().ParseArguments<Options>(_args)
+                using (var parser = CaseInsensitiveParser())
```
DaveTryon, line 56: >using [](start = 16, length = 5)

I wasn't sure about the lifetime here, since we use the helpWriter that we pass into the settings that the parser consumes. Looking at [https://github.com/commandlineparser/commandline/blob/master/src/CommandLine/ParserSettings.cs](https://github.com/commandlineparser/commandline/blob/master/src/CommandLine/ParserSettings.cs), though, it explicitly says that the helpWriter is ours.

src/Actions/Actions/SelectAction.cs
RobGallo: I've made the change to lowercase 'o'. I thought about using "_lockObject," which I think is fine. But I didn't see the harm in being more specific.

src/CLI/CLI.csproj
RobGallo: Just an error. Fixed.

src/CLI/Program.cs
RobGallo: It makes sense that the Parser doesn't dispose of the HelpWriter because it doesn't allocate the TextWriter. I wouldn't expect this to change.

PR #368, opened 2020-05-05
Re-address Accessibility Insights for Windows changes events bug

RobGallo:
#### Describe the change

See [Accessibility Insights for Windows bug #694](https://github.com/microsoft/accessibility-insights-windows/issues/694)
and PR #356.

The original fix was to change the order in which events were registered. However, the documentation for
[UiaChangeInfo](https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcore/ns-uiautomationcore-uiachangeinfo)
indicates the ids being passed into AddChangesEventHandler were wrong: the code was passing event ids and should have been passing property, annotation, style, or text attribute ids.

This change obviates the need for the change in PR #356 where event registration could optionally be ordered, so that change has been undone as part of this change.

In order to get changes events, I had to make sure the ids included in TextAttributeType did not contain UIA_TextAttributeId. To do that, I removed UIA_TextAttributeId and refactored the code which included it so that it is no longer necessary.

I added unit tests to ensure my changes to TypeBase work, but I did not include unit tests for the entire TypeBase class. Some of the Testing of type base may be tested by the testing of other types which inherit from TypeBase. At any rate, testing TypeBase in its entirety is a subject for another PR.

I also did a small amount of refactoring in TypeBase to abstract the logic of determining whether or not fields should be included in the dictionary and extracting the logic of adding the field data to the dictionary. I think those changes made the code easier to read.

I refactored TextAttributeType to remove the GetTemplate function, which I don't think should belong to that class. I also made the GetTemplate code easier to read by emplying some typedefs and creation functions.


src/Core/Types/TypeBase.cs
```diff
@@ -12,34 +15,45 @@ namespace Axe.Windows.Core.Types
...
+        const string DefaultNamePattern = "UIA_";
         protected Dictionary<int, string> Dic { get; private set; }
+        private readonly string _namePattern = null;
 
-        protected TypeBase() : this(NamePattern) { }
+        protected TypeBase() : this(DefaultNamePattern) { }
 
         protected TypeBase(string np)
         {
+            _namePattern = np;
```
DaveTryon, line 26: I have mixed feelings about storing _namePattern as a class member. It's only needed during the constructor, so this could just as easily be handled as a parameter.

src/Desktop/Types/TextAttributeType.cs
```diff
@@ -17,6 +18,7 @@ namespace Axe.Windows.Desktop.Types
     public class TextAttributeType : TypeBase
     {
 #pragma warning disable CA1707 // Identifiers should not contain underscores
+        [NotType]
```
DaveTryon, line 21: This is awkward--we're declaring a list of types but saying that this isn't a type. It would probably help if the attribute had a more specific name--name `NotChangeEventType` (or something similar) instead of `NotType`

src/Desktop/UIAutomation/EventHandlers/ChangesEventListener.cs
```diff
@@ -11,51 +13,27 @@ namespace Axe.Windows.Desktop.UIAutomation.EventHandlers
...
-        };
+            // Ids based on expected values for UiaChangeInfo.uiaId
+            // https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcore/ns-uiautomationcore-uiachangeinfo
+            // Note: property ids are not included as they are already handled by the property changed event handler.
+
+            var changeTypes = new List<int>();
+            changeTypes.AddRange(AnnotationType.GetInstance().Values);
+            changeTypes.AddRange(Styles.StyleId.GetInstance().Values);
+            changeTypes.AddRange(TextAttributeType.GetInstance().Values);
+            changeTypes.Add(ChangeInfoType.UIA_SummaryChangeId);
```
DaveTryon, line 28: Would it make sense to list the IDs in the same order that they appear in the referenced link? In that link, they appear as:
UIA_SummaryChangeId
Property Attributes (excluded, per comment)
Text Attributes
Annotation Types
Styles

src/Core/Types/TypeBase.cs
RobGallo: I had those same mixed feelings. But I chose the member variable so that I didn't have to keep passing the value around as a parameter.

src/Desktop/UIAutomation/EventHandlers/ChangesEventListener.cs
RobGallo: Will do.

PR #392, opened 2020-07-06
Enable authenticode signing with 3rd party cert

jalkire:
Currently, we do not digitally sign axe-windows assemblies. This PR enables digital signing for its assemblies using the proper Microsoft 3rd party cert. [Successful signed build here](https://dev.azure.com/mseng/1ES/_build/results?buildId=12530233&view=results).

### DIFF
--- a/build/settings.targets
+++ b/build/settings.targets
@@ -19,6 +19,7 @@
         <Otherwise>
           <ItemGroup>
             <FilesToSign Include="@(DropSignedFile)">
+              <Authenticode>3PartySHA2</Authenticode>
               <StrongName>StrongName</StrongName>
             </FilesToSign>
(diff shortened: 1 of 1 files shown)


build/settings.targets
```diff
@@ -19,6 +19,7 @@
         <Otherwise>
```
jalkire, line 19: The lines above this sign the axe-windows CLI project with a different (1st party) cert. I believe this should be left as-is, but let me know if that doesn't seem right.

DaveTryon: The MSI properties list Microsoft as the publisher, so I think this would make sense. @RobGallo or @peterdur might have more recent insights based on the thread, but that could certainly be in a separate PR, if needed.

PR #395, opened 2020-07-09
Fix #285 Add .Net Framework tests in addition to .Net Core tests

### LINKED ISSUE #285: [TASK]  Add .Net Framework tests in addition to .Net Core tests
## Describe the bug

Now that we've converted axe-windows to .Net Standard, the unit test projects have all been converted to .Net Core. However, we should be able to build the test projects as .Net Core and .Net Framework simultaneously. This task is a reminder to investigate and, if possible, do that.

RobGallo:
#### Describe the change

Using .Net Framework 4.72 because that is what Accessibility Insights for Windows uses.

Moved target frameworks property to a common .target file where it can easily be reused. The condition in the .target file is needed because CLITests references CLI, which is built for the .Net Core Framework and not .Net Standard (So .Net Framework is not compatible with CLI).

### DIFF
--- a/build/NetStandardTest.targets
+++ b/build/NetStandardTest.targets
@@ -6,8 +6,8 @@
   </PropertyGroup>
 
   <PropertyGroup>
-        <IsPackable>false</IsPackable>
-<NoWarn>1701;1702;CA1707</NoWarn>
+    <IsPackable>false</IsPackable>
+    <NoWarn>1701;1702;CA1707</NoWarn>
   </PropertyGroup>
 
   <Import Project=".\NetStandardAll.targets" />
--- a/build/NetStandardTest.targets
+++ b/build/NetStandardTest.targets
@@ -1,7 +1,13 @@
 ﻿<Project>
 
+  <!-- The following condition should be true for almost all test projects -->
+  <PropertyGroup Condition="$(TargetFramework) == '' AND $(TargetFrameworks) == ''">
+    <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup>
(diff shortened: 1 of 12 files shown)


build/NetStandardTest.targets
```diff
@@ -1,7 +1,13 @@
 ﻿<Project>
 
+  <!-- The following condition should be true for almost all test projects -->
+  <PropertyGroup Condition="$(TargetFramework) == '' AND $(TargetFrameworks) == ''">
+    <TargetFrameworks>netcoreapp3.1;net472</TargetFrameworks>
+  </PropertyGroup>
+
   <PropertyGroup>
-    <NoWarn>1701;1702;CA1707</NoWarn>
+        <IsPackable>false</IsPackable>
```
jalkire, line 9:
```suggestion
    <IsPackable>false</IsPackable>
```
Spacing fix

build/NetStandardTest.targets
```diff
@@ -1,7 +1,13 @@
...
+<NoWarn>1701;1702;CA1707</NoWarn>
```
jalkire, line 10:
```suggestion
    <NoWarn>1701;1702;CA1707</NoWarn>
```
Spacing fix

src/Win32Tests/Win32Tests.csproj
```diff
@@ -1,12 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
-
-    <IsPackable>false</IsPackable>
-  </PropertyGroup>
-
-  <ItemGroup>
+    <ItemGroup>
```
jalkire, line 3:
```suggestion
  <ItemGroup>
```
Spacing fix

PR #397, opened 2020-07-15
Sign external dependencies

jalkire:
#### Describe the change
Currently, we have two non-Microsoft-signed dependencies. This PR ensures they are signed with the 3rd party certificate (but not strong name signed by us) in both the CLI installer and zip. Signed build [here](https://dev.azure.com/mseng/1ES/_build/results?buildId=12609892&view=results).

### DIFF
--- a/src/CLI_Installer/Product.wxs
+++ b/src/CLI_Installer/Product.wxs
@@ -59,33 +59,33 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
 
     <ComponentGroup Id="ProductComponents" Directory="INSTALLFOLDER">
       <Component Id="ProductComponent" Guid="548A9965-21AD-4CDF-99AD-A6F2A47D0AE8">
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\AxeWindowsCLI.exe" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\AxeWindowsCLI.dll" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\AxeWindowsCLI.deps.json" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\AxeWindowsCLI.runtimeconfig.json" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\Axe.Windows.Actions.dll" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\Axe.Windows.Automation.dll" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\Axe.Windows.Core.dll" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\Axe.Windows.Desktop.dll" />
-        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\Axe.Windows.Rules.dll" />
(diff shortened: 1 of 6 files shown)


src/CLI_Installer/Product.wxs
```diff
@@ -59,36 +59,36 @@ Licensed under the MIT license. See LICENSE file in the project root for full li
...
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\Axe.Windows.Telemetry.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\Axe.Windows.Win32.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\CommandLine.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\Microsoft.Win32.SystemEvents.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\Newtonsoft.Json.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\System.Drawing.Common.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\System.IO.Packaging.dll" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\thirdpartynotices.html" />
-        <File Source="..\CLI\bin\Release\netcoreapp3.1\README.MD" />
+        <File Source="..\CLI_Full\bin\Release\netcoreapp3.1\win7-x86\AxeWindowsCLI.exe" />
```
DaveTryon, line 62: This may be a problem. IIRC, the EXE from the CLI_full project is stamped in a way that prevents it from loading any of the locally-installed .NET assemblies. If that's the case, then the MSI-installed CLI will be broken as soon as it needs to use anything that isn't installed in the MSI. Easy to check--install the MSI and see if the CLI can still function as expected.

jalkire: Reverted Product.wxs to original state, moved 3rd party signing to CLI, and updated zip creation script to source from CLI and CLI_Full as appropriate.

tools/scripts/BuildZippedCLI.ps1
```diff
@@ -66,22 +66,47 @@ function Create-Zipfile([string]$srcDir, [string]$app, [string]$zipFile, [string
     $scratch=(Get-UniqueTempFolder $app)[0]
     Write-Verbose "scratch = $scratch"
 
-    Copy-Files $srcDir '*.*' $scratch
+    Prepare-Zipfile $srcDir $otherSignedSrcDir $scratch $patternsToRemove $otherSignedAssemblies
```
DaveTryon, line 69: Umm... does this run? PowerShell requires functions to appear _before_ the caller. I'd expect an error from calling Prepare-ZipFile before it's defined

jalkire: Yes, you can find its output in the artifacts of the signed build linked to above.

DaveTryon: Got it--Create-ZipFile is called from CreateCLIZip, which appears _after_ Prepare-ZipFile. By the time the interpreter executes the call on line 69, it's already encountered Prepare-ZipFile, so it works.

Just to be safe for future changes, you might consider moving Prepare-ZipFile so it appears before Create-ZipFile

jalkire: Order updated

PR #398, opened 2020-07-16
Set Codesign Validation scan directory when creating zip

jalkire:
#### Describe the change
Currently, the auto injected Codesign Validation task scans our entire project for signing violations. Since we only ship specific 
assemblies, this leads to hundreds of false positives. A build variable is available to set the root directory that the task scans. This PR sets that variable to the temp directory we create while making our zipped CLI package. This package contains a superset of all the assemblies that we ship (MSI, zip, and Nuget package), so it seems like a good choice for where to direct the Codesign Validation task--I'm open to other suggestions for where to scan, though. You can find an example of the impact on the build task [here](https://dev.azure.com/mseng/1ES/_build/results?buildId=12617518&view=logs&j=7c3ebc89-e7dd-5791-8c02-2ea1f3a27b50&t=95036218-ae8d-5d14-6fa9-97adee74970f) (different branch, same variable setting).

### DIFF
--- a/tools/scripts/BuildZippedCLI.ps1
+++ b/tools/scripts/BuildZippedCLI.ps1
@@ -38,7 +38,13 @@ function Get-UniqueTempFolder([string] $app){
     $uniqueName=$app + '_' + $guid
     $tempFolder=Join-Path $env:temp $uniqueName
 	New-Item -Path $tempFolder -ItemType Directory
+    Set-CodeSignValidationDirectory $tempFolder
     return $tempFolder
+}
+
+function Set-CodeSignValidationDirectory([string] $dir){
+    Write-Verbose "Setting ADO job variable GDN_CODESIGN_TARGETDIRECTORY to $dir"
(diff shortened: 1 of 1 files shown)


tools/scripts/BuildZippedCLI.ps1
```diff
@@ -95,8 +101,6 @@ function Create-Zipfile([string]$srcDir, [string]$otherSignedSrcDir, [string]$ap
     Prepare-Zipfile $srcDir $otherSignedSrcDir $scratch $patternsToRemove $otherSignedAssemblies
 
     Create-Archive $scratch $zipFile
-
-    Remove-Item $scratch -Force -Recurse
```
jalkire, line 99: We need to leave the directory with contents for the task to scan. Since script only runs when a specific environment variable is set (CreateAxeWindowsZippedCLI) and the validation task runs at the end of our pipeline, leaving the files in the system temp directory seemed acceptable to me.

PR #399, opened 2020-07-16
Update nuspec metadata per guidance

jalkire:
#### Describe the change
This PR updates our nuget package metadata based on internal guidance.


RobGallo:
The Authors element should say "Microsoft and collaborators". The owners element should say "Microsoft".

PR #405, opened 2020-07-20
Set an ADO variable for the CodeCov version

RobGallo:
#### Describe the change

This will make it possible for dependabot to update the version of CodeCov without an engineer having to manually change the version used in prbuild.yml.

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -111,7 +111,7 @@ jobs:
   - task: CmdLine@2
     displayName: 'Upload coverage to codecov.io'
     inputs:
-      script: $(USERPROFILE)\.nuget\packages\codecov\1.12.1\tools\win7-x86\codecov.exe -t $(CODECOV_TOKEN) -f **/*coverage.cobertura.xml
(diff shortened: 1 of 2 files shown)


build/prbuild.yml
```diff
@@ -111,7 +111,7 @@ jobs:
   - task: CmdLine@2
     displayName: 'Upload coverage to codecov.io'
     inputs:
-      script: $(USERPROFILE)\.nuget\packages\codecov\1.12.1\tools\win7-x86\codecov.exe -t $(CODECOV_TOKEN) -f **/*coverage.cobertura.xml
+      script: $(USERPROFILE)\.nuget\packages\codecov\$(CODECOV_VERSION)\tools\win7-x86\codecov.exe -t $(CODECOV_TOKEN) -f **/*coverage.cobertura.xml
```
DaveTryon, line 114: Would it make sense to have a script that auto-detects the version, either by inspecting the versions under .nuet\packages\codecov, or by examining the build file where it's specified? This approach feels like we need to manually tweak it every time we pick up a new drop :(

DaveTryon: Thanks for explaining!

PR #426, opened 2020-08-13
Fix minor typo in Telemetry.md

decriptor:
Fixed a minor typo, `When` instead of `Wehn`


DaveTryon (approved):
Thanks!

PR #430, opened 2020-08-25
Update thirdpartynotices.html

jalkire:
#### Describe the change
We had not updated our third party notices for a while. Generated using the Notice -> Download option [here](https://dev.azure.com/mseng/1ES/_componentGovernance/320?_a=components&typeId=24175&alerts-view-option=active).

### DIFF
--- a/thirdpartynotices.html
+++ b/thirdpartynotices.html
@@ -34,109 +34,26 @@
             <li>
     <details>
         <summary>
-            NuGet.Frameworks 5.0.0 - Apache-2.0
+            Codecov 1.12.2 - MIT
         </summary>
         
         <ul><li>(c) 2008 VeriSign, Inc.</li>
-<li>(c) Microsoft Corporation.</li></ul>
+<li>Copyright (c) by P.J. Plauger</li>
+<li>copyright Excepting NotParsed</li>
+<li>Copyright (c) 2017-Present Larz White, Kim J. Nordmo</li>
+<li>4Copyright (c) 2017-Present Larz White, Kim J. Nordmo</li>
+<li>Copyright (c) 2016 JetBrains http://www.jetbrains.com</li>
+<li>Copyright (c) 2005 - 2020 Giacomo Stelluti Scala &amp; Contributors</li></ul>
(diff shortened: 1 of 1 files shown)


thirdpartynotices.html
```diff
@@ -18,4 +484,511 @@
...
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+        </pre>
+    </details>
+</li>
+<li>
+    <details>
+        <summary>
+            WiX 3.11.2 - MS-RL
```
dbjorge, line 941: I'm not sure this is accurate; we distribute an MSI built with WiX, do we really distribute WiX itself? If we do, I think the MS-RL demands certain copyleft obligations our release process probably isn't including right now

dbjorge: (if it's not accurate, that probably suggests that you need to go disable this package from the "Configure NOTICE generation" UI in component governance, then regenerate the file)

dbjorge (changes requested):
One suggested change for the actual NOTICE content! Talked with John offline about automating this with the OSE build task, but that can be separate from this PR

thirdpartynotices.html
jalkire: We do not distribute WiX separately from the MSI we build with it. Our current [ai-win thirdpartynotices file includes the same notice](https://github.com/microsoft/accessibility-insights-windows/blob/e896138d51854a1eb928e6aa3a509791dd3e7a8b/thirdpartynotices.html#L669). @DaveTryon, do you have any background on the inclusion of the WiX notice?

DaveTryon: @jalkire, @dbjorge, our last guidance from legal was that we didn't need the MS-RL license, but that we had an "alternative license" for WiX. I'll forward you both the email but I don't have specifics about what "alternative license" means here...

jalkire: Removed wix & regenerated file per email guidance

dbjorge (approved):

PR #439, opened 2020-09-28
fix: Update rules concerning disallowed characters

DaveTryon:
#### Describe the change

Axe.Windows currently has 2 rules that check for "Special Characters". After discussing internally, we decided that checking for characters in the Unicode private area is more generally applicable, and should replace the existing checks.

| Old Rule | New Rule | Comment |
| --- | --- | --- |
| LocalizedLandmarkTypeExcludesSpecialCharacters | LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters | Private characters rule is new |
| NameExcludesSpecialCharacters | NameExcludesPrivateUnicodeCharacters | Private characters rule already existed |


src/Rules/Library/LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters.cs
```diff
@@ -8,21 +8,21 @@
...
-    class LocalizedLandmarkTypeExcludesSpecialCharacters : Rule
+    [RuleInfo(ID = RuleId.LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters)]
+    class LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters : Rule
     {
-        public LocalizedLandmarkTypeExcludesSpecialCharacters()
+        public LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters()
         {
-            this.Info.Description = Descriptions.LocalizedLandmarkTypeExcludesSpecialCharacters;
-            this.Info.HowToFix = HowToFix.LocalizedLandmarkTypeExcludesSpecialCharacters;
+            this.Info.Description = Descriptions.LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters;
```
jalkire, line 16: Is there a particular reason we aren't reusing the `Descriptions.PropertyExcludesPrivateUnicodeCharacters` and `HowToFix.PropertyExcludesPrivateUnicodeCharacters` strings that the other ExcludesPrivateUnicodeCharacters rules use ([example](https://github.com/microsoft/axe-windows/blob/master/src/Rules/Library/NameExcludesPrivateUnicodeCharacters.cs))?

src/Rules/Resources/HowToFix.resx
```diff
@@ -544,7 +539,7 @@ If possible, use a predefined (non-custom) control type and the default localize
     <value>If the sibling hyperlink with the same Name navigates to a different destination, change the Name of either hyperlink.</value>
   </data>
   <data name="PropertyExcludesPrivateUnicodeCharacters" xml:space="preserve">
```
jalkire, line 541: Similar to my other comment, it seems like this string covers the `LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters` string added above

src/Rules/Library/LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters.cs
DaveTryon: The SpecialCharacters rule has a fairly generic string for Description but a very specific string for HowToFix. I stayed with the existing pattern but am more than willing to change the formatting. @RobGallo, any preferences?

RobGallo: If the text is generic enough then using a single string setting would be okay . usually the text contains the name of the property in question; and I think that may be useful.

jalkire: That makes sense. It seems like in this case, the Descriptions string (`Descriptions.PropertyExcludesPrivateUnicodeCharacters`) should be reused because it is identical to the new string, but the HowToFix string needs to be unique because it contains extra info. Does that sound right?

jalkire (approved):

RobGallo (approved):

PR #448, opened 2020-10-22
fix: Normalize mouse tracking

DaveTryon:
#### Describe the change
This is the axe.windows portion of https://github.com/microsoft/accessibility-insights-windows/issues/974. It modifies the mouse tracking code to normalize the element based on the current TreeViewMode. It also makes a couple of unrelated changes (2 comments, 1 extra newline, and sorting the using blocks) that I happened to notice while I was working on this change.

The change itself is simple. The SelectAction already has the TreeViewMode, so we just need to pass it to the MouseTracker, then use this to normalize the element to the TreeViewMode. The code to check the IsContentElement and IsControlElement properties could be removed if we always calling Normalize, but that complicates the object management, so we make an extra check (which the compiler will optimize easily) and only call Normalize when we know our existing element won't match the current TreeViewMode.

I've validated this new code in AIWin, and it works as expected. Using the sample app from the original issue, I did the following with both old and new code:
1. Start AIWin with "always on top" set to true, let it wait at the welcome screen
2. Start the test app. I'd previously positioned the windows so AIWin would overlap the edit control containing the placeholder (the placeholder is implemented as a TextBlock that returns false for both IsContentElement and IsControlElement properties)
3. Close the AIWin welcome screen
4. Hover over the edit control with the placeholder
5. Check the properties (and grab a screenshot)

Before this change, the TextBlock appears in the hierarchy tree, which in incorrect since TreeViewMode is set to Control. After this change, the TextBlock no longer appears in the hierarchy tree. In the screenshot below, the old behavior is on the top (the TextBlock's entry in the hierarchy tree is highlighted in red), and the new behavior is on the bottom.

image

I've also checked TreeViewMode values of Raw and Content, and everything worked as expected.

### DIFF
--- a/src/Desktop/UIAutomation/A11yAutomation.cs
+++ b/src/Desktop/UIAutomation/A11yAutomation.cs
@@ -1,9 +1,5 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using System;
-using System.Diagnostics;
-using System.Linq;
-using System.Runtime.InteropServices;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
 using Axe.Windows.Core.Misc;
@@ -11,6 +7,10 @@ using Axe.Windows.Core.Types;
 using Axe.Windows.Desktop.UIAutomation.TreeWalkers;
 using Axe.Windows.Telemetry;
 using Axe.Windows.Win32;
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Runtime.InteropServices;
 using UIAutomationClient;
 
 namespace Axe.Windows.Desktop.UIAutomation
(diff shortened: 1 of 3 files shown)


src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -213,7 +213,25 @@ public static A11yElement GetAppElement(A11yElement e)
         }
 
         /// <summary>
-        /// Get IUIAutomationTreeWalker based on inidcated mode.
+        /// Normalize an element to the specified TreeViewMode
+        /// </summary>
+        /// <param name="element">A11yElement</param>
+        /// <param name="treeViewMode">mode to normalize to</param>
+        /// <returns></returns>
+        public static A11yElement GetNormalizedElement(A11yElement element, TreeViewMode treeViewMode)
```
RobGallo, line 221: I think the word "normalized" should be removed. passing in the tree scope communicates well the purpose of the function and normalized doesn't carry much meaning in this context.

RobGallo (approved):

src/Desktop/UIAutomation/A11yAutomation.cs
DaveTryon: We discussed this and didn't find a better alternative, so leaving it as is...

DaveTryon:
Closing/reopening to jump-start PR update

DaveTryon:
Closing/reopening again after refreshing the trigger

DaveTryon:
Closing and reopening one more time

PR #455, opened 2020-10-29
Chore: Break ControlShouldSupportTablePattern rule into two rules

RobGallo:
#### Describe the change

The old rule returned two possible `ErrorCode` values (in addition to `ErrorCode.Pass`), whereas each of the new separate rules only returns one. This makes it possible for us to specify the failure error code as part of the `RuleInfo` data, which will allow us to automate documentation for the rules regarding which error codes are returned by which rules.

### DIFF
--- a/src/Core/Enums/RuleId.cs
+++ b/src/Core/Enums/RuleId.cs
@@ -126,6 +126,7 @@ namespace Axe.Windows.Core.Enums
         ControlShouldSupportSpreadsheetItemPattern,
         ControlShouldSupportTableItemPattern,
         ControlShouldSupportTablePattern,
+        ControlShouldSupportTablePatternInEdge,
         ControlShouldSupportTogglePattern,
         ControlShouldSupportTransformPattern,
         ControlShouldSupportTextPattern,
--- a/src/Rules/Library/ControlShouldSupportTablePattern.cs
+++ b/src/Rules/Library/ControlShouldSupportTablePattern.cs
@@ -25,7 +25,7 @@ namespace Axe.Windows.Rules.Library
         {
             if (e == null) throw new ArgumentNullException(nameof(e));
 
-                return Patterns.Table.Matches(e) ? EvaluationCode.Pass : EvaluationCode.Error;
+            return Patterns.Table.Matches(e) ? EvaluationCode.Pass : EvaluationCode.Error;
         }
 
         protected override Condition CreateCondition()
--- a/src/RulesTest/Library/ControlShouldSupportTablePattern.cs
+++ b/src/RulesTest/Library/ControlShouldSupportTablePattern.cs
@@ -51,7 +51,7 @@ namespace Axe.Windows.RulesTest.Library
             int[] expectedTypes = { ControlType.Calendar, ControlType.Table };
             var unexpectedTypes = ControlType.All.Difference(expectedTypes);
             
-                var e = new MockA11yElement();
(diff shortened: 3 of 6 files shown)


src/RulesTest/Library/ControlShouldSupportTablePatternInEdge.cs
```diff
@@ -0,0 +1,75 @@
...
+            Assert.AreEqual(PropertyType.UIA_IsTablePatternAvailablePropertyId, this.Rule.Info.PropertyID);
+        }
+
+        [TestMethod]
+        public void ConditionMatchesExpectedElement()
+        {
+            int[] expectedTypes = { ControlType.Calendar, ControlType.Table };
+            var unexpectedTypes = ControlType.All.Difference(expectedTypes);
+            
+                var e = new MockA11yElement();
```
DaveTryon, line 54: One extra indent on this line
```suggestion
            var e = new MockA11yElement();
```

src/RulesTest/Library/ControlShouldSupportTablePattern.cs
```diff
@@ -52,5 +44,31 @@ public void CheckPropertyIdIsSet()
...
             Assert.AreEqual(PropertyType.UIA_IsTablePatternAvailablePropertyId, this.Rule.Info.PropertyID);
         }
+
+        [TestMethod]
+        public void ConditionMatchesExpectedElement()
+        {
+            int[] expectedTypes = { ControlType.Calendar, ControlType.Table };
+            var unexpectedTypes = ControlType.All.Difference(expectedTypes);
+            
+                var e = new MockA11yElement();
```
DaveTryon, line 54: This line is indented one tab too far
```suggestion
            var e = new MockA11yElement();
```

src/Core/Enums/RuleId.cs
```diff
@@ -126,6 +126,7 @@ public enum RuleId
         ControlShouldSupportSpreadsheetItemPattern,
         ControlShouldSupportTableItemPattern,
         ControlShouldSupportTablePattern,
+        ControlShouldSupportTablePatternInEdge,
```
DaveTryon, line 129: @RobGallo, should we update the RuleDescriptions.md file as rules are modified or when we have a release? I tend to lean towards keeping the code and the description in sync in the repo, independent of the release schedule. Thoughts?

RobGallo: I think we should keep them updated as code changes, not on release. I'll add that as a change.

src/Rules/Library/ControlShouldSupportTablePattern.cs
```diff
@@ -25,15 +25,12 @@ public override EvaluationCode Evaluate(IA11yElement e)
         {
             if (e == null) throw new ArgumentNullException(nameof(e));
 
-            if (Patterns.Table.Matches(e))
-                return EvaluationCode.Pass;
-
-            return StringProperties.Framework.Is(Framework.Edge).Matches(e) ? EvaluationCode.Warning : EvaluationCode.Error;
+                return Patterns.Table.Matches(e) ? EvaluationCode.Pass : EvaluationCode.Error;
```
DaveTryon, line 28: This row is indented with one extra tab
```suggestion
            return Patterns.Table.Matches(e) ? EvaluationCode.Pass : EvaluationCode.Error;
```

DaveTryon (approved):

PR #457, opened 2020-10-30
Chore: Add unit tests for the code which runs all rules

RobGallo:
#### Describe the change

Extracted code from `Rules` into `RuleRunner` in order to test the rule running logic. The extraction was essentially a cut and paste operation.

The idea is to have rule running logic thoroughly unit tested before an upcoming modification which will add another mechanism for evaluating rules and returning `EvaluationCode` values.

### DIFF
--- a/src/Rules/RuleRunner.cs
+++ b/src/Rules/RuleRunner.cs
@@ -1,0 +1,102 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
+using Axe.Windows.Telemetry;
+using System;
+using System.Collections.Generic;
+using static System.FormattableString;
+
+namespace Axe.Windows.Rules
+{
+    class RuleRunner
+    {
+        private readonly IRuleProvider _provider;
+
+        public RuleRunner(IRuleProvider provider)
+        {
+            if (provider == null) throw new ArgumentNullException(nameof(provider));
+
+            _provider = provider;
+        }
+
+        public RunResult RunRuleByID(RuleId id, IA11yElement element)
+        {
+            var rule = _provider.GetRule(id);
+            if (rule == null)
+            {
+                return new RunResult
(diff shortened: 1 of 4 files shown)


src/Rules/Rules.cs
```diff
@@ -43,19 +42,7 @@ public static class Rules
         /// <returns></returns>
         public static RunResult RunRuleByID(RuleId id, IA11yElement element)
```
DaveTryon, line 43: RuneRuleByID currently has no unit test coverage. How hard would it be to add this test?

src/Rules/RuleRunner.cs
```diff
@@ -0,0 +1,102 @@
...
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
+using Axe.Windows.Telemetry;
+using System;
+using System.Collections.Generic;
+using static System.FormattableString;
+
+namespace Axe.Windows.Rules
+{
+    class RuleRunner
```
DaveTryon, line 12: There's already RuleSelection.RuleRunner class, and adding Rules.RuleRunner could lead to some potential confusion. Should we consider a different name?

RobGallo: I think the benefit of having an accurately named class outweighs the potential for confusion. I also think namespaces go a long way toward clarifying which class is being used.

src/Rules/Rules.cs
RobGallo: I wouldn't say it's "hard" so much as it provides very little value. Any change would also involve modifying how other assemblies in the solution call rules, which I am reluctant to do.

DaveTryon: I can see the point of "adds very little value", but since this is public, it should be trivial to add a test similar to `MonsterTest.GetTestResultsAsDictionary`, but call `Rules.RunRuleByID` instead of `Rules.RunAll`.  Another option would be to remove `Rules.RunRuleByID`, since (if memory serves), it was added for a specific scenario that is no longer in play.

DaveTryon (approved):
Approved with a suggestion about removing Rules.RunRuleByID

PR #458, opened 2020-11-02
Chore: Add RuleInfo.ErrorCode and IRule.PassesTests

RobGallo:
#### Describe the change

This change makes it so that if `IRule.Info.ErrorCode` exists for a givenRule`, `IRule.PassesTest` will be called instead of `IRule.Evaluate` to determine if the rule passes its test. In addition, if the rule fails its test, the value of `IRule.Info.ErrorCode` will be returned.

This means that each rule can now specify in its info the `EvaluationCode` to be returned if the rule's test does not pass. And that, in turn, allows us to get the error code for each rule via reflection and display it in our documentation.

### DIFF
--- a/src/Rules/RuleRunner.cs
+++ b/src/Rules/RuleRunner.cs
@@ -87,7 +87,12 @@ namespace Axe.Windows.Rules
             if (!ShouldRun(rule.Condition, element))
                 return result;
 
-            result.EvaluationCode = rule.Evaluate(element);
+            // The PassesTest call is the method we want to convert all rules to use
+            // Using RuleInfo.ErrorCode will allow us to use reflection to document the result of each rule
+            if (rule.Info.ErrorCode.HasValue)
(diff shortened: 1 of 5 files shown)


src/Rules/RuleRunner.cs
```diff
@@ -87,7 +87,12 @@ private static RunResult RunRuleWorker(IRule rule, IA11yElement element)
             if (!ShouldRun(rule.Condition, element))
                 return result;
 
-            result.EvaluationCode = rule.Evaluate(element);
+            // The PassesTest call is the method we want to convert all rules to use
+            // Using RuleInfo.ErrorCode will allow us to use reflection to document the result of each rule
+            if (rule.Info.ErrorCode.HasValue)
```
DaveTryon, line 92: Should there be a try/catch around these, so that we can accommodate the rules that currently return ErrorCode.RuleExecutionError? I was thinking that the catch block would set result.EvaluationCode to RuleExecutionError...

DaveTryon (approved):

PR #459, opened 2020-11-03
Chore: Add unit test for Rule.Evaluate not implemented

RobGallo:
#### Describe the change

The new test ensures that if `Rule.Info.ErrorCode` has no value, `Rule.Evaluate` is expected to be overridden. This is the converse of the existing test where if `Rule.Info.ErrorCode` has a value, `Rule.PassesTest` is expected to be overridden.

Also added tests to confirm the test rule implementations match expectations for the other tests.

Made `Rule.Evaluate` and `Rule.PassesTest` `virtual` so they could be overridden by subclasses.

### DIFF
--- a/src/Rules/Rule.cs
+++ b/src/Rules/Rule.cs
@@ -51,9 +51,14 @@ namespace Axe.Windows.Rules
             return null;
         }
 
-        public abstract EvaluationCode Evaluate(IA11yElement element);
+        public virtual EvaluationCode Evaluate(IA11yElement element)
+        {
+            // This base class function should never be called
+            // once all rules have been converted to use RuleInfo.EvaluationCode, this function will be removed
+            throw new NotImplementedException();
+        }
 
-        public bool PassesTest(IA11yElement element)
+        public virtual bool PassesTest(IA11yElement element)
         {
             // This base class function should never be called
             // once all rules have been converted to use RuleInfo.EvaluationCode, this function will be designated abstract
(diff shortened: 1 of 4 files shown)


src/Rules/Rule.cs
```diff
@@ -51,9 +51,14 @@ private RuleInfo GetRuleInfoFromAttributes()
...
+            // This base class function should never be called
+            // once all rules have been converted to use RuleInfo.EvaluationCode, this function will be removed
+            throw new NotImplementedException();
+        }
 
-        public bool PassesTest(IA11yElement element)
+        public virtual bool PassesTest(IA11yElement element)
         {
             // This base class function should never be called
             // once all rules have been converted to use RuleInfo.EvaluationCode, this function will be designated abstract
```
DaveTryon, line 64: It seems like this would follow the same pattern as the Evaluate method and be removed once the migration is complete. If that's true, then the comment is incorrect. Maybe:
```suggestion
            // once all rules have been converted to use RuleInfo.EvaluationCode, this function will be removed
```

src/RulesTest/RuleImplementations/RuleWithPassesTestNotImplemented.cs
```diff
@@ -18,7 +18,8 @@ public RuleWithPassesTestNotImplemented() : base()
 
         public override EvaluationCode Evaluate(IA11yElement element)
         {
-            throw new NotImplementedException();
+            // Don't throw an exception here so we can be sure the only exception thrown is in Rule.PassesTest
+            return EvaluationCode.Pass;
```
DaveTryon, line 22: General comment about this class: Would it be worth a comment that the entire class will go away once the migration to PassesTest is complete?

DaveTryon (approved):
Approved with a couple of comments

PR #461, opened 2020-11-05
fix: Update behavior for no output file to not interact with file system

### LINKED ISSUE #460: [bug] Exception being thrown while creating a scanner using default parameters inside a UWP app
## Is your feature request related to a problem? Please describe.

Not really a feature, but writing test results does take some time. Usually it doesn't really make a difference, but if there are hundreds of test runs, this might make a difference.
## Describe the solution you'd like

This is where it becomes a bit difficult. Currently an empty path is being replaced with a default path, setting the FileFormat to "None" still creates files (I think). The scanner factory could expose a function that skips the creation of the file helper entirely, something like "BuildNoFileLogging". However having the configuration object still have the filepath and fileformat property might look a bit weird.

## Describe alternatives you've considered

None yet as this isn't a big issue. It would be nice to have.

chingucoding:
#### Describe the change
I updated the behavior for the creation of scanner tools to also allow a null OutputfileHelper. That way, we skip creating a folder which can break if a path is not valid.

To verify that changes, I updated existing tests to also scan with no file specified and check if the scans run the same as file runs (error results wise).

Fixes #460 

Open question(s):
- Should there be a UWP test app that creates scanners?
- More generally, should there be a UWP app to scan against?

### DIFF
--- a/src/Automation/ScanTools.cs
+++ b/src/Automation/ScanTools.cs
@@ -14,7 +14,6 @@ namespace Axe.Windows.Automation
 
         public ScanTools(IOutputFileHelper outputFileHelper, IScanResultsAssembler resultsAssembler, ITargetElementLocator targetElementLocator, IAxeWindowsActions actions, INativeMethods nativeMethods)
         {
-            if (outputFileHelper == null) throw new ArgumentNullException(nameof(outputFileHelper));
             if (resultsAssembler == null) throw new ArgumentNullException(nameof(resultsAssembler));
             if (targetElementLocator == null) throw new ArgumentNullException(nameof(targetElementLocator));
             if (actions == null) throw new ArgumentNullException(nameof(actions));
--- a/src/Automation/ScanToolsBuilder.cs
+++ b/src/Automation/ScanToolsBuilder.cs
@@ -24,8 +24,9 @@ namespace Axe.Windows.Automation
 
         public IScanTools Build()
         {
+            // A null file helper indicates no writing to file and as such is allowed.
             return new ScanTools(
-                _outputFileHelper ?? _factory.CreateOutputFileHelper(null),
+                    _outputFileHelper,
                     _factory.CreateResultsAssembler(),
                     _factory.CreateTargetElementLocator(),
                     _factory.CreateAxeWindowsActions(),
--- a/src/Automation/Scanner.cs
+++ b/src/Automation/Scanner.cs
@@ -17,7 +17,6 @@ namespace Axe.Windows.Automation
         {
             if (config == null) throw new ArgumentNullException(nameof(config));
             if (scanTools == null) throw new ArgumentNullException(nameof(scanTools));
-            if (scanTools.OutputFileHelper == null) throw new ArgumentException(ErrorMessages.ScanToolsOutputFileHelperNull, nameof(scanTools));
 
             _config = config;
             _scanTools = scanTools;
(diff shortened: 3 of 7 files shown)


RobGallo (changes requested):
Thanks so much for the contribution and for pointing out the problem! I am marking this PR as requesting changes so that it isn't accidentally commited. I have devised an alternative fix to address the issue in PR #462 
My fix is just a bit more targeted in the files it touches, although it basically accomplishes the same thing as this PR in a different way.

chingucoding:
@RobGallo So should I close my PR then?

RobGallo:
Yes, please. 😊

DaveTryon:
Returning to the questions about UWP apps:

Q: Should we have a UWP app that triggers scans?
A: We don't try to create sample scanning apps for other frameworks, so it would seem odd to single out UWP as a sample, especially if UWP sandboxing is likely to interfere with the scanner.

Q: Should we have a UWP app that we scan?
A: We currently have "sampler" apps based on Win32, WPF, and WinForms, so having a UWP app here would extend an existing pattern. These sampler apps are intended to serve as "early warning systems", so if we add or change a rule that is likely to impact a broad number of users due to a framework bug, our unit tests will catch it and we can then assess whether or not we want to continue with that change. If we had a UWP app that followed that same pattern and could be integrated into unit tests, then a UWP app would probably make sense. The UWP sampler app would need to be runnable directly from the command line, without requiring an installation process.

chingucoding:
> A: We don't try to create sample scanning apps for other frameworks, so it would seem odd to single out UWP as a sample, especially if UWP sandboxing is likely to interfere with the scanner.

Makes sense.

> We currently have "sampler" apps based on Win32, WPF, and WinForms, so having a UWP app here would extend an existing pattern. These sampler apps are intended to serve as "early warning systems", so if we add or change a rule that is likely to impact a broad number of users due to a framework bug, our unit tests will catch it and we can then assess whether or not we want to continue with that change. If we had a UWP app that followed that same pattern and could be integrated into unit tests, then a UWP app would probably make sense. The UWP sampler app would need to be runnable directly from the command line, without requiring an installation process.

Well running UWP apps is a bit of problem. UWP apps can be started from the command line. The bad news: As far as I know it is not possible to run a UWP without installation. The good news is that the UWP apps can generate an `Install.ps1` file as part of their build output. Running that script installs the app silently without creating a popup. After that, the app can be ran through the commandline.

DaveTryon:
@chingucoding, automatically running tests in UWP may not have an ideal answer. Silently installing a UWP app during the test could lead to unexpected behavior in our shared environment build pipeline--it introduces interesting cleanup questions, as well as the risk of having concurrently running builds competing over whether or not the UWP app should be installed. Sideloading apps comes to mind as a possibility, although it may still cause contention between concurrently running builds.

We could theoretically have a UWP test app in the project, and then each engineer could run the project locally from inside the debugger and run the CLI against that app. In reality, though, it will very rarely get run.

There may be a middle ground, although it's again non-ideal. We could create a test class that only runs when some external condition is satisfied. We do this, for example, in https://github.com/microsoft/accessibility-insights-windows/tree/master/src/UITests, where the tests short-circuit and call Assert.Inconclusive if WinAppDriver isn't running. If there is a reliable way to detect that the UWP sampler app is installed, then the tests can execute against that app and compare the scan results against a baseline, just we do with the other apps.

If we move forward, it should probably be in a new feature request, since we've repurposed this issue into a bug that fixed the Exception.

PR #462, opened 2020-11-05
Fix: Create directory for output files only when files are requested

### LINKED ISSUE #460: [bug] Exception being thrown while creating a scanner using default parameters inside a UWP app

RobGallo:
#### Describe the change

Fixes #460

This fix still allows the exception mentioned in the bug to be propagated to the user. However, the user will not encounter the exception if no output file is requested.

The OutputFileHelper class must still exist as part of the IScanTools interface even though it will not be used if no output files are requested. This helps to keep the change small and to  keep the code concise.

### DIFF
--- a/src/Automation/OutputFileHelper.cs
+++ b/src/Automation/OutputFileHelper.cs
@@ -14,6 +14,7 @@ namespace Axe.Windows.Automation
     {
         private readonly string _outputDirectory;
         private readonly ISystemDateTime _dateTime;
+        private readonly ISystemIODirectory _directory;
 
         private string _scanId;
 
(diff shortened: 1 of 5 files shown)


src/AutomationTests/OutputFileHelperUnitTests.cs
```diff
@@ -157,7 +155,7 @@ public void OutputFileHelperCtor_WithOutputDirectory_UsesProvidedDirectory()
 
         [TestMethod]
         [Timeout(1000)]
-        public void OutputFileHelperCtor_CreatesNonexistentDirectory()
+        public void OutputFileHelper_EnsureOutputDirectoryExists_CreatesNonexistentDirectory()
```
DaveTryon, line 158: It feels like we're only covering 2 of 3 cases. We validate the exception and if the directory doesn't exist, but we don't cover the case where directory already exists (and CreateDirectory never gets called)

chingucoding (approved):
Looks good to me.

src/Automation/OutputFileHelper.cs
```diff
@@ -67,6 +66,20 @@ private static void VerifyPathOrThrow(string path)
             }
         }
 
+        /// <summary>
+        /// Ensures the output directory for the file path returned by <see cref="GetNewA11yTestFilePath"/>
```
DaveTryon, line 70: Should this just say _outputDirectory?

DaveTryon (approved):
Approved with one suggestion

PR #469, opened 2020-11-11
chore: Let callers of ColorContrast tests specify the config

DaveTryon:
#### Describe the change

In the current code, the ColorContrastConfig is hardcoded in Axe.Windows, and is optimized for use in Android, where CPU resources are limited and where image resolution is high. In AIWin, we have plenty of CPU resources and lower resolution images, so it makes sense to allow AIWin to control the config parameters. To make that possible, I've converted ColorContrastConfig from a static class to an interface, and used dependency injection to specify the config when constructing the BitmapCollection object. For backward compatibility with existing AIWin code, I've added a temporary legacy constructor to BitmapCollection that defaults to the same config that used to be hardcoded. The temporary legacy constructor ought to be removed once AIWin starts calling the ctor with its own implementation of IColorContastConfig.


src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
```diff
@@ -1,44 +0,0 @@
...
-         * contains Text.
-         */
-        public static readonly int MinNumberColorTransitions = 4;
-
-        /**
-         * We will inevitably find other color combinations in a row. To be highly confident
-         * we have text, we expect the most likely color combination to be at least this dominant.
-         * Type: Double between 0 and 1.
-         */
-        public static readonly double TextColorPairDominanceValue = .2;
```
DaveTryon, line 25: The `TextColorPairDominanceValue` field is not used anywhere in the code, so I went ahead and removed it from the interface

karanbirsingh (approved):

Issue #441, opened 2020-09-28
Add how-to-fix info for 'custom' DataGridCell behavior in old .NET Framework versions


karanbirsingh:
This issue tracks adding more actionable guidance for the DataGridCell issue in https://github.com/microsoft/accessibility-insights-windows/issues/931. 

It was fixed by directing the user to either update to .NET 4.7.1 or to use accessibility switches. This DataGridCell issue (alongside some other updates) was explicitly called out in the .NET 'What's new in accessibility' [documentation here](https://docs.microsoft.com/en-us/dotnet/framework/whats-new/whats-new-in-accessibility#whats-new-in-accessibility-in-net-framework-471).

If axe-windows detects this error, we should suggest updating/accessibility switches in the tool itself.

Details:
- create a new failing rule that targets DataGrids with this issue (can look for custom control type/localized control type, wpf framework, and class name)
- include how-to-fix information that mentions the suggested fix from above & how-to-fix information from `LocalizedControlTypeIsNotCustom`
- update the behavior of LocalizedControlTypeIsNotCustom: https://github.com/microsoft/axe-windows/blob/95b5fe9f41b9e9f8c5b9b7ddc8534451ca63f2f4/src/Rules/Library/LocalizedControlTypeIsNotCustom.cs#L16 to exclude testing elements that are tested by the new rule

An alternative approach would require updating our how-to-fix mappings to be chosen dynamically from within a rule, rather than a 1:1 mapping from a rule to HowToFix entities. When discussing with Rob, the new-rule approach seemed more aligned with the existing codebase.

We should also check whether this rule has an associated how-to-fix link. Based on that, it might make sense to add or update the doc with a note about this issue.

RobGallo:
Addressed in PR #478

DaveTryon:
@karanbirsingh, this is in the build at https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1467.01. Please validate & close at your convenience 😃

PR #478, opened 2020-12-08
Fix: Issue #441

RobGallo:
#### Describe the change

Add extra how-to-fix information about accessibility switches in the case where the `ControlType` property is `Custom`, the `Framework` is "WPF", and the `ClassName` is "DataGridCell". The how-to-fix information in question applies to .Net versions older than 4.7.1.

To do this, I created a new rule: `LocalizedControlTypeIsNotCustomWPFGridCell`. The new rule should apply in the same cases as `LocalizedControlTypeIsNotCustom`, but only if the platform is `WPF` and the `ClassName` is `DataGridCell`. The existing rule, `LocalizedControlTypeIsNotCustom`, should not be run in the case of a WPF DataGridCell.

I used string formatting for the htf info to avoid duplicating text in documentation which could become mismatched.

I reworked the unit tests in LocalizedControlTypeIsNotCustom.cs a bit to (hopefully) make them a little more robust and a little easier to understand.

### DIFF
--- a/src/Rules/Library/LocalizedControlTypeIsNotCustomWPFGridCell.cs
+++ b/src/Rules/Library/LocalizedControlTypeIsNotCustomWPFGridCell.cs
@@ -4,22 +4,23 @@ using System;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
 using Axe.Windows.Core.Types;
-using Axe.Windows.Rules.PropertyConditions;
 using Axe.Windows.Rules.Resources;
 using static Axe.Windows.Rules.PropertyConditions.BoolProperties;
 using static Axe.Windows.Rules.PropertyConditions.ControlType;
 using static Axe.Windows.Rules.PropertyConditions.Relationships;
+using static Axe.Windows.Rules.PropertyConditions.ElementGroups;
 using static Axe.Windows.Rules.PropertyConditions.StringProperties;
+using System.Globalization;
 
 namespace Axe.Windows.Rules.Library
 {
-    [RuleInfo(ID = RuleId.LocalizedControlTypeNotCustom)]
-    class LocalizedControlTypeIsNotCustom : Rule
+    [RuleInfo(ID = RuleId.LocalizedControlTypeNotCustomWPFGridCell)]
+    class LocalizedControlTypeIsNotCustomWPFGridCell : Rule
     {
-        public LocalizedControlTypeIsNotCustom()
+        public LocalizedControlTypeIsNotCustomWPFGridCell()
         {
             this.Info.Description = Descriptions.LocalizedControlTypeNotCustom;
-            this.Info.HowToFix = HowToFix.LocalizedControlTypeNotCustom;
+            this.Info.HowToFix = String.Format(CultureInfo.CurrentCulture, HowToFix.LocalizedControlTypeNotCustomWPFGridCell, HowToFix.LocalizedControlTypeNotCustom);
             this.Info.Standard = A11yCriteriaId.ObjectInformation;
             this.Info.PropertyID = PropertyType.UIA_LocalizedControlTypePropertyId;
         }
@@ -33,13 +34,10 @@ namespace Axe.Windows.Rules.Library
 
         protected override Condition CreateCondition()
         {
-            var dataGridDetailsPresenter = ClassName.Is("DataGridDetailsPresenter") & Parent(DataItem);
-
             return Custom
                 & IsKeyboardFocusable
                 & LocalizedControlType.NotNullOrEmpty
-                & ~dataGridDetailsPresenter
-                & ~ElementGroups.WPFDataGridCell;
+                & WPFDataGridCell;
         }
     } // class
 } // namespace
--- a/src/Rules/Library/LocalizedControlTypeIsNotCustomWPFGridCell.cs
+++ b/src/Rules/Library/LocalizedControlTypeIsNotCustomWPFGridCell.cs
@@ -1,16 +1,15 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
-using System;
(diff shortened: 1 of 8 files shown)


src/Rules/PropertyConditions/ElementGroups.cs
```diff
@@ -19,6 +19,7 @@ static class ElementGroups
         // the following occurs for xaml expand/collapse controls
         private static Condition FocusableGroup = Group & IsKeyboardFocusable & (StringProperties.Framework.Is(Core.Enums.Framework.WPF) | StringProperties.Framework.Is(Core.Enums.Framework.XAML));
 
+        private static Condition WPF = Framework.Is(Core.Enums.Framework.WPF);
```
DaveTryon, line 22: Out of curiosity, what's the criterion for using a Create method or just putting these inline? Most of the Conditions use Create methods, so should we lean that way, even for simple ones? As an aside, we currently have 2 inline creators that are each more complex than the CreateWPRDataGridCellCondition method

src/Rules/Resources/HowToFix.resx
```diff
@@ -478,6 +478,11 @@ Provide a string for the LocalizedControlType property that concisely describes
 
 Better:
 If possible, use a predefined (non-custom) control type and the default localized control type. The UIA framework will provide the correct localized control type automatically.</value>
+  </data>
+  <data name="LocalizedControlTypeNotCustomWPFGridCell" xml:space="preserve">
+    <value>{0}
+    
+If your application targets a version of .Net Framework before 4.7.1, but is running on a .Net version of 4.7.1 or higher, you may be able to address this issue by taking advantage of accessibility enhancements in .Net 4.7.1. To do this, you will need to enable accessibility switches as described on the following page: https://docs.microsoft.com/en-us/dotnet/framework/whats-new/whats-new-in-accessibility#accessibility-switches</value>
```
DaveTryon, line 485: Capitalization: ".NET" is capitalized, this this becomes:
```suggestion
If your application targets a version of .NET Framework before 4.7.1, but is running on a .NET version of 4.7.1 or higher, you may be able to address this issue by taking advantage of accessibility enhancements in .NET 4.7.1. To do this, you will need to enable accessibility switches as described on the following page: https://docs.microsoft.com/en-us/dotnet/framework/whats-new/whats-new-in-accessibility#accessibility-switches</value>
```

DaveTryon, line 485: "running on .NET version of 4.7.1 or higher" somehow seems strange to my ear. Would it make sense to add a bit to explain how this could happen, something like "running on a system where .NET version 4.7.1 or higher is installed" ?

src/RulesTest/Library/LocalizedControlTypeIsNotCustom.cs
```diff
@@ -10,51 +11,101 @@ public class LocalizedControlTypeIsNotCustom
     {
         private Axe.Windows.Rules.IRule Rule = new Axe.Windows.Rules.Library.LocalizedControlTypeIsNotCustom();
 
-        [TestMethod]
-        public void TestControlTypeIdIsNotCustomControl()
+        private MockA11yElement CreateElementExpectedToMatchCondition()
```
DaveTryon, line 14: Maybe use a test fixture and give this a `BeforeEach` tag?

src/RulesTest/Library/LocalizedControlTypeIsNotCustomWPFGridCell.cs
```diff
@@ -0,0 +1,110 @@
...
+using EvaluationCode = Axe.Windows.Rules.EvaluationCode;
+
+namespace Axe.Windows.RulesTest.Library
+{
+    [TestClass]
+    public class LocalizedControlTypeIsNotCustomWPFGridCell
+    {
+        private Axe.Windows.Rules.IRule Rule = new Axe.Windows.Rules.Library.LocalizedControlTypeIsNotCustomWPFGridCell();
+
+        private MockA11yElement CreateElementExpectedToMatchCondition()
```
DaveTryon, line 14: Another likely candidate for a test fixture

src/RulesTest/PropertyConditions/ElementGroupsTests.cs
```diff
@@ -200,5 +200,23 @@ public void WPFScrollBarPageButtons_NotPageUpOrPageDown_False()
                 Assert.IsFalse(ElementGroups.WPFScrollBarPageButtons.Matches(e));
             } // using
         }
+
+        [TestMethod]
+        public void WPFDataGridCell_MatcheAsExpected()
```
DaveTryon, line 205: Looks like there's a typo in the function name here

src/Rules/Library/LocalizedControlTypeIsNotCustomWPFGridCell.cs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
```
DaveTryon, line 3: Can we please sort the using blocks alphabetically?

src/Rules/Resources/HowToFix.resx
RobGallo: Done

src/RulesTest/Library/LocalizedControlTypeIsNotCustom.cs
RobGallo: Perhaps if there were multiple setup steps. But as is, I don't think it adds much.

src/Rules/PropertyConditions/ElementGroups.cs
RobGallo: I'll change this. I made it a method because I originally thought it would be more complicated.

DaveTryon (approved):
LGTM

Issue #483, opened 2020-12-16
[BUG] Control should not support table pattern error on SysListView32


RobGallo:
## Describe the bug

The table pattern is implemented on Win32 list views (class "SysListView32"), but the user has no way to remove it and the platform is no longer supported.

**To Reproduce**

Occurs for the CList class when building an app using MFC. We don't have a specific example since the problem was discovered using an a11ytest file and the app being tested belongs to a third party.

## Expected behavior

The error should not be thrown

## Actual behavior

An error is thrown.

DaveTryon:
@RobGallo, this is fixed in the current Canary build: https://github.com/microsoft/accessibility-insights-windows/releases/tag/v1.1.1467.01. Please validate and close at your convenience. Thanks!

RobGallo:
Fixed in PR #484

PR #484, opened 2020-12-16
Fix: Address bug #483

RobGallo:
#### Describe the change

No longer check that the table pattern doesn't exist for list elements with the class name "SysListView32" because developers can't remove the pattern and the framework is unlikely to change.

### DIFF
--- a/src/RulesTest/Library/ControlShouldNotSupportTablePattern.cs
+++ b/src/RulesTest/Library/ControlShouldNotSupportTablePattern.cs
@@ -1,0 +1,72 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Types;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Linq;
+using EvaluationCode = Axe.Windows.Rules.EvaluationCode;
+
+namespace Axe.Windows.RulesTest.Library
+{
+    [TestClass]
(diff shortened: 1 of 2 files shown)


src/RulesTest/Library/ControlShouldNotSupportTablePattern.cs
```diff
@@ -0,0 +1,72 @@
...
+            e.Patterns.Add(new A11yPattern(e, PatternType.UIA_TablePatternId));
+
+            Assert.AreEqual(EvaluationCode.Error, Rule.Evaluate(e));
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ArgumentNull_Exception()
+        {
+            Rule.Evaluate(null);
```
DaveTryon, line 38: I've recently started to use a new format for expected exceptions, patterned after what is used in the TypeScript library. Certainly not required, but just for consideration. Remove the ExpectedException attribute from the test, then declare the expectation inside the function:

Assert.ThrowsException<ArgumentNullException)(() => Rule.Evaluate(null));

I personally prefer having the declaration of the expectation on the same line as the call that triggers the Exception

DaveTryon (approved):
Made one suggestion. Not sure why CodeCov is claiming that we're losing coverage in EventListenerFactory.cs with these changes.

src/RulesTest/Library/ControlShouldNotSupportTablePattern.cs
RobGallo: I like it better too. I don't think I'll change it now though. Perhaps in the future as we alter tests we can change to that format.

PR #489, opened 2021-01-05
Chore: Replace lengthy framework conditions with shorter ones

RobGallo:
#### Describe the change

This replaces a `Condition` statement in a rule like

`StringProperties.Framework.Is(FrameworkId.Edge)`

with

`Edge`

This makes the code easier to comprehend.

### DIFF
--- a/src/Rules/PropertyConditions/ElementGroups.cs
+++ b/src/Rules/PropertyConditions/ElementGroups.cs
@@ -168,7 +168,6 @@ namespace Axe.Windows.Rules.PropertyConditions
 
         private static bool IsParentWPFDataItem(IA11yElement e)
         {
-
             // GetUIFramework searches the ancestors for a valid framework value in addition to checking the element itself
             return e.GetUIFramework() == Core.Enums.FrameworkId.WPF
                 && e.Parent != null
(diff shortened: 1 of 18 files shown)


src/Rules/PropertyConditions/ElementGroups.cs
```diff
@@ -169,14 +168,16 @@ private static Condition CreateNonFocusableSliderButtonsCondition()
 
         private static bool IsParentWPFDataItem(IA11yElement e)
         {
+
```
DaveTryon, line 171: (nit): Remove this empty line?

src/Rules/PropertyConditions/ElementGroups.cs
```diff
@@ -169,14 +168,16 @@ private static Condition CreateNonFocusableSliderButtonsCondition()
...
+            // GetUIFramework searches the ancestors for a valid framework value in addition to checking the element itself
             return e.GetUIFramework() == Core.Enums.FrameworkId.WPF
                 && e.Parent != null
                 && e.Parent.ControlTypeId == Axe.Windows.Core.Types.ControlType.UIA_DataItemControlTypeId;
         }
 
         private static bool HasAllowedPlatformPropertiesForText(IA11yElement e)
         {
-            return !IsPlatformWinForms(e)
+            return !WinForms.Matches(e)
```
DaveTryon, line 180: Are we concerned about e ever being null at this point? The old code would have simply returned false due to the null propagation. I think we'll get a NullReferenceException if a null makes it to here, and I'm not finding any guard against it

DaveTryon (commented):
I love the idea here! I had a couple of questions before approving

src/Rules/PropertyConditions/ElementGroups.cs
RobGallo: Eventually this code will make it to the extension method GetUIFramework, which has a null check in it. When that happens, null is returned for the value of the string. So this should not throw an exception if e is null

DaveTryon (approved):

PR #492, opened 2021-01-06
fix: Update WXS file to use updated components

DaveTryon:
#### Describe the change

Since the last release, we've updated several assemblies that used to use the assemblies from the OS-installed .NET runtime. The revised assembly versions are pulled into the app's manifest, so we need to reference them in the WXS file. I also updated the names of two file ID's to be a little more descriptive (it seemed to make sense now that we have more files).

Tested by building the MSI locally, installing it, and ensuring that it works as expected.


jalkire (approved):
Works for me!

PR #494, opened 2021-01-07
fix: Update www.access-board.gov links with current URL

jalkire:
#### Describe the change
We noticed in AI-Windows that sometime late last year, the www.access-board.gov site we link to changed. Now, our links are redirected to the correct page, but the fragment bringing you to the right place on the page is lost. You can see the full repro steps in this AI-Win issue: https://github.com/microsoft/accessibility-insights-windows/issues/1017.

This PR fixes the issue by updating the URL. These were the only places in either repo referencing access-board.gov.


DaveTryon (approved):
These look good to me. I wish we had a good way to catch external changes like this

PR #506, opened 2021-01-28
Chore: Update more rules to use the new PassesTest method instead of Evaluate

RobGallo:
#### Describe the change

Adding the evaluation code to the rule's info will allow us to automatically generate documentation which provides the rule's expected error code.

### DIFF
--- a/src/Rules/Library/ButtonInvokeAndTogglePatterns.cs
+++ b/src/Rules/Library/ButtonInvokeAndTogglePatterns.cs
@@ -16,13 +16,14 @@ namespace Axe.Windows.Rules.Library
             this.Info.Description = Descriptions.ButtonInvokeAndTogglePatterns;
             this.Info.HowToFix = HowToFix.ButtonInvokeAndTogglePatterns;
             this.Info.Standard = A11yCriteriaId.NameRoleValue;
+            this.Info.ErrorCode = EvaluationCode.Error;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            var rule = Relationships.All(Patterns.Invoke, Patterns.Toggle);
+            var rule = ~(Relationships.All(Patterns.Invoke, Patterns.Toggle));
 
-            return rule.Matches(e) ? EvaluationCode.Error : EvaluationCode.Pass;
+            return rule.Matches(e);
         }
 
         protected override Condition CreateCondition()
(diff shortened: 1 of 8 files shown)


src/Rules/Library/ButtonInvokeAndTogglePatterns.cs
```diff
@@ -16,13 +16,14 @@ public ButtonInvokeAndTogglePatterns()
...
+            this.Info.ErrorCode = EvaluationCode.Error;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            var rule = Relationships.All(Patterns.Invoke, Patterns.Toggle);
+            var rule = ~(Relationships.All(Patterns.Invoke, Patterns.Toggle));
 
-            return rule.Matches(e) ? EvaluationCode.Error : EvaluationCode.Pass;
```
DaveTryon, line 25: Would it be any more efficient to leave rule unchanged, then negate the return value from Matches? That saves us creating a Condition object, and I think you've followed this pattern in previous changes

src/Rules/Library/ButtonToggleAndExpandeCollapsePatterns.cs
```diff
@@ -16,13 +16,14 @@ public ButtonToggleAndExpandCollapsePatterns()
...
             this.Info.HowToFix = HowToFix.ButtonToggleAndExpandCollapsePatterns;
             this.Info.Standard = A11yCriteriaId.NameRoleValue;
+            this.Info.ErrorCode = EvaluationCode.Error;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            var rule = Relationships.All(Patterns.Toggle, Patterns.ExpandCollapse);
+            var rule = ~(Relationships.All(Patterns.Toggle, Patterns.ExpandCollapse));
```
DaveTryon, line 24: Would it be any more efficient to leave rule unchanged, then negate the return value from Matches? That saves us creating a Condition object, and I think you've followed this pattern in previous changes

DaveTryon (approved):
Approved with a couple of comments

src/Rules/Library/ButtonInvokeAndTogglePatterns.cs
RobGallo: I did that originally, and then changed it to this way. Honestly, I'm not crazy about the '!' negation, even in the cases I've used it. I think it makes it harder to easily groc the intent of the method.

PR #507, opened 2021-01-29
docs: add instructions for silent installation

karanbirsingh:
This PR adds information about how to install the MSI installer silently. We're adding docs like this for several of our Windows installers (AI-Windows, AI-Unified, etc) so users who wish to avoid the graphical interface can do so.

### DIFF
--- a/src/CLI/README.MD
+++ b/src/CLI/README.MD
@@ -2,6 +2,15 @@
 
 This tool allows you to scan your application for accessibility issues from your currently existing scripting system.
 
+## Installing the CLI
+
+You can install the CLI by executing the MSI installer on the [releases](https://github.com/microsoft/axe-windows/releases/) page. If you prefer to install silently, open a command prompt as administrator, change to the directory containing the file, and include the `/quiet` flag:
+
+```
+cd %HOMEPATH%\Downloads
+msiexec.exe /i AxeWindowsCLI-1.0.6.msi /quiet
+```
+
 ## Running the tool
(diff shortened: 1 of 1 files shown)


src/CLI/README.MD
```diff
@@ -2,6 +2,15 @@
 
 This tool allows you to scan your application for accessibility issues from your currently existing scripting system.
 
+## Installing the CLI
+
+You can install the CLI by executing the MSI installer on the [releases](https://github.com/microsoft/axe-windows/releases/) page. If you prefer to install silently, open a command prompt as administrator, change to the directory containing the file, and include the `/quiet` flag:
+
+```
+cd %HOMEPATH%\Downloads
```
DaveTryon, line 10: Rather than putting this here, what if we put it into the comments, something like:

If you prefer to install silently, open a command prompt as administrator, change to the directory containing the file (typically your downloads folder), and execute the MSI file directly, including the `/quiet` flag:

`msiexec.exe /i AxeWindowsCLI-1.0.6.msi /quiet`

One other thought--by including the version number here, is this adding another thing to change when we publish a new version?

karanbirsingh: Updated!

chingucoding: I think it's best to have something like `msiexec.exe /i AxeWindowsCLI-VERSION.msi /quiet` or `msiexec.exe /i AxeWindowsCLI-X.Y.Z.msi /quiet` so that this won't be outdated or have to be updated every release.

RobGallo (approved):

PR #510, opened 2021-02-04
Chore: Update rules to use the new PassesTest method instead of Evaluate

RobGallo:
#### Describe the change

    Adding the evaluation code to the rule's info will allow us to automatically generate documentation which provides the rule's expected error code.

This is a large change. When reviewing please try to make sure the expected error code for each rule is consistent with its former value. (Those changes were made by hand.)

All the changes in the test files were made by a search and replace.

### DIFF
--- a/src/Rules/Library/IsContentElementTrueOptional.cs
+++ b/src/Rules/Library/IsContentElementTrueOptional.cs
@@ -25,7 +25,7 @@ namespace Axe.Windows.Rules.Library
 
         public override bool PassesTest(IA11yElement e)
         {
-            return true;
+            return false;
         }
 
         protected override Condition CreateCondition()
--- a/src/Rules/Library/IsContentElementTrueOptional.cs
+++ b/src/Rules/Library/IsContentElementTrueOptional.cs
@@ -20,11 +20,12 @@ namespace Axe.Windows.Rules.Library
             this.Info.HowToFix = HowToFix.IsContentElementTrueOptional;
             this.Info.Standard = A11yCriteriaId.ObjectInformation;
             this.Info.PropertyID = PropertyType.UIA_IsContentElementPropertyId;
+            this.Info.ErrorCode = EvaluationCode.Open;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            return EvaluationCode.Open;
+            return true;
         }
 
         protected override Condition CreateCondition()
--- a/src/Rules/Library/ListItemSiblingsUnique.cs
+++ b/src/Rules/Library/ListItemSiblingsUnique.cs
@@ -34,7 +34,7 @@ namespace Axe.Windows.Rules.Library
                 & Name.Is(e.Name)
                 & LocalizedControlType.Is(e.LocalizedControlType));
             var count = siblings.GetValue(e);
-            if (count < 1) throw new Exception(String.Format(CultureInfo.CurrentCulture, ErrorMessages.NoElementFound, this.Info.ID));
+            if (count < 1) throw new Exception(String.Format(CultureInfo.InvariantCulture, ErrorMessages.NoElementFound, this.Info.ID));
 
             return count == 1;
         }
--- a/src/Rules/Library/ListItemSiblingsUnique.cs
+++ b/src/Rules/Library/ListItemSiblingsUnique.cs
@@ -5,6 +5,7 @@ using Axe.Windows.Core.Enums;
 using Axe.Windows.Rules.PropertyConditions;
 using Axe.Windows.Rules.Resources;
 using System;
+using System.Globalization;
 using static Axe.Windows.Rules.PropertyConditions.BoolProperties;
 using static Axe.Windows.Rules.PropertyConditions.ControlType;
 using static Axe.Windows.Rules.PropertyConditions.Relationships;
@@ -20,9 +21,10 @@ namespace Axe.Windows.Rules.Library
             this.Info.Description = Descriptions.ListItemSiblingsUnique;
             this.Info.HowToFix = HowToFix.ListItemSiblingsUnique;
             this.Info.Standard = A11yCriteriaId.NameRoleValue;
+            this.Info.ErrorCode = EvaluationCode.Warning;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
             if (e == null) throw new ArgumentNullException(nameof(e));
             if (e.Parent == null) throw new ArgumentException(ErrorMessages.ElementParentNull, nameof(e));
@@ -32,9 +34,9 @@ namespace Axe.Windows.Rules.Library
                 & Name.Is(e.Name)
                 & LocalizedControlType.Is(e.LocalizedControlType));
             var count = siblings.GetValue(e);
-            if (count < 1) return EvaluationCode.RuleExecutionError;
+            if (count < 1) throw new Exception(String.Format(CultureInfo.CurrentCulture, ErrorMessages.NoElementFound, this.Info.ID));
 
-            return count == 1 ? EvaluationCode.Pass : EvaluationCode.Warning;
+            return count == 1;
         }
 
         protected override Condition CreateCondition()
--- a/src/Rules/Library/LocalizedControlTypeIsNotEmpty.cs
+++ b/src/Rules/Library/LocalizedControlTypeIsNotEmpty.cs
@@ -19,7 +19,6 @@ namespace Axe.Windows.Rules.Library
             this.Info.HowToFix = HowToFix.LocalizedControlTypeNotEmpty;
             this.Info.Standard = A11yCriteriaId.ObjectInformation;
             this.Info.PropertyID = PropertyType.UIA_LocalizedControlTypePropertyId;
-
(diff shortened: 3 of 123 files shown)


src/Rules/Library/IsContentElementTrueOptional.cs
```diff
@@ -20,11 +20,12 @@ public IsContentElementTrueOptional()
...
             this.Info.Standard = A11yCriteriaId.ObjectInformation;
             this.Info.PropertyID = PropertyType.UIA_IsContentElementPropertyId;
+            this.Info.ErrorCode = EvaluationCode.Open;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            return EvaluationCode.Open;
+            return true;
```
DaveTryon, line 28: Won't this result in an evaluation code of Pass instead of Open?

src/Rules/Library/ListItemSiblingsUnique.cs
```diff
@@ -32,9 +34,9 @@ public override EvaluationCode Evaluate(IA11yElement e)
                 & Name.Is(e.Name)
                 & LocalizedControlType.Is(e.LocalizedControlType));
             var count = siblings.GetValue(e);
-            if (count < 1) return EvaluationCode.RuleExecutionError;
+            if (count < 1) throw new Exception(String.Format(CultureInfo.CurrentCulture, ErrorMessages.NoElementFound, this.Info.ID));
```
DaveTryon, line 37: Not entirely sure why the Exception is different, but I'd suggest CultureInfo.InvariantCulture instead of CultureInfo.CurrentCulture (the message isn't localized, so it doesn't make sense to localize the ID)

src/Rules/Library/LocalizedControlTypeIsNotEmpty.cs
```diff
@@ -19,13 +19,15 @@ public LocalizedControlTypeIsNotEmpty()
             this.Info.HowToFix = HowToFix.LocalizedControlTypeNotEmpty;
             this.Info.Standard = A11yCriteriaId.ObjectInformation;
             this.Info.PropertyID = PropertyType.UIA_LocalizedControlTypePropertyId;
+
```
DaveTryon, line 22: Remove extra newline on line 22 (you didn't add it in other classs so I'm assuming this is an oversight)

src/Rules/Library/NameIsEmptyButElementIsNotKeyboardFocusable.cs
```diff
@@ -21,13 +22,14 @@ public NameIsEmptyButElementIsNotKeyboardFocusable()
...
             this.Info.Standard = A11yCriteriaId.ObjectInformation;
             this.Info.PropertyID = PropertyType.UIA_NamePropertyId;
+            this.Info.ErrorCode = EvaluationCode.Open;
         }
 
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            if (e == null) return EvaluationCode.RuleExecutionError;
+            if (e == null) throw new ArgumentNullException(nameof(e));
```
DaveTryon, line 30: This Exception makes sense, but there are lots of other rules that will merrily throw a NulReferenceException in this case (for example, NameIsInformative (chosen just because it's the next in the review) has no guard. It feels like the guard ought to be consistent in all rules (could be a separate PR)

src/Rules/Library/ParentChildShouldNotHaveSameNameAndLocalizedControlType.cs
```diff
@@ -18,14 +19,15 @@ public ParentChildShouldNotHaveSameNameAndLocalizedControlType()
...
-        public override EvaluationCode Evaluate(IA11yElement e)
+        public override bool PassesTest(IA11yElement e)
         {
-            if (e == null) return EvaluationCode.RuleExecutionError;
-            if (e.Parent == null) return EvaluationCode.RuleExecutionError;
+            if (e == null) throw new ArgumentNullException(nameof(e));
+            if (e.Parent == null) throw new ArgumentException(ErrorMessages.ElementParentNull, nameof(e));
 
-            return e.Name == e.Parent.Name && e.LocalizedControlType == e.Parent.LocalizedControlType ? EvaluationCode.Error : EvaluationCode.Pass;
+            return e.Name != e.Parent.Name || e.LocalizedControlType != e.Parent.LocalizedControlType;
```
DaveTryon, line 30: I wish we had some parentheses here, just to be explicit about grouping:
```suggestion
            return (e.Name != e.Parent.Name) || (e.LocalizedControlType != e.Parent.LocalizedControlType);
```

src/Rules/Library/SiblingUniqueAndFocusable.cs
```diff
@@ -48,9 +50,9 @@ public override EvaluationCode Evaluate(IA11yElement e)
                 & Name.Is(e.Name)
                 & LocalizedControlType.Is(e.LocalizedControlType));
             var count = siblings.GetValue(e);
-            if (count < 1) return EvaluationCode.RuleExecutionError;
+            if (count < 1) throw new Exception(string.Format(CultureInfo.CurrentCulture, ErrorMessages.NoElementFound, this.Info.ID));
```
DaveTryon, line 53: Not entirely sure why the Exception is different, but I'd suggest CultureInfo.InvariantCulture instead of CultureInfo.CurrentCulture (the message isn't localized, so it doesn't make sense to localize the ID)
```suggestion
            if (count < 1) throw new Exception(string.Format(CultureInfo.InvariantCulture, ErrorMessages.NoElementFound, this.Info.ID));
```

src/Rules/Library/SiblingUniqueAndNotFocusable.cs
```diff
@@ -47,9 +49,9 @@ public override EvaluationCode Evaluate(IA11yElement e)
                 & Name.Is(e.Name)
                 & LocalizedControlType.Is(e.LocalizedControlType));
             var count = siblings.GetValue(e);
-            if (count < 1) return EvaluationCode.RuleExecutionError;
+            if (count < 1) throw new Exception(string.Format(CultureInfo.CurrentCulture, ErrorMessages.NoElementFound, this.Info.ID));
```
DaveTryon, line 52: Not entirely sure why the Exception is different, but I'd suggest CultureInfo.InvariantCulture instead of CultureInfo.CurrentCulture (the message isn't localized, so it doesn't make sense to localize the ID)
```suggestion
            if (count < 1) throw new Exception(string.Format(CultureInfo.InvariantCulture, ErrorMessages.NoElementFound, this.Info.ID));
```

DaveTryon (changes requested):
I've only reviewed the main classes so far (no unit tests), but since I found issues there, I'll review the unit tests on the next iteration

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -150,6 +150,9 @@
   <data name="NoControlTypeEntryFound" xml:space="preserve">
     <value>No control type entry was found in dictionary</value>
   </data>
+  <data name="NoElementFound" xml:space="preserve">
+    <value>Based on the Rule.Condition match, there should have been at least one element found in the PassesTest method of {0}</value>
```
DaveTryon, line 154: Will this be exposed to users? It seems a little too obscure to really help them

src/RulesTest/Library/NameIsEmptyButElementIsNotKeyboardFocusableTest.cs
```diff
@@ -1,8 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
```
DaveTryon, line 4: In "using System;" added intentionally? No changes in this file seem to necessitate it

src/Rules/Library/IsContentElementTrueOptional.cs
RobGallo: I suspected I might make that mistake at some point. Nice catch.

src/Rules/Library/ListItemSiblingsUnique.cs
RobGallo: Why the exception is different from what?

And I am going to fix these instances in a single change so I can make sure they compile.

src/Rules/Library/NameIsEmptyButElementIsNotKeyboardFocusable.cs
RobGallo: Agreed.

src/Rules/Library/SiblingUniqueAndFocusable.cs
RobGallo: Fixed.

src/Rules/Library/SiblingUniqueAndNotFocusable.cs
RobGallo: Fixed.

src/Rules/Library/ListItemSiblingsUnique.cs
DaveTryon: I understand the Exception now--you're throwing here and RuleExecutionError will come from the Exception handler.

DaveTryon (approved):

PR #513, opened 2021-02-08
Chore: Remove Evaluate method from IRule

RobGallo:
#### Details

Now that we've moved away from using the `IRule.Evaluate` method in favor of the `Irule.PassesTest` method (so that the `EvaluationCode` can be stored in `IRule.Info.ErrorCode`), remove `Evaluate` altogether.

##### Motivation

We want to expose the error code each rule returns through reflection so that we can generate documentation about the rules directly from the code.

##### Context

I'm not satisfied with the fact that it's possible for the `RuleInfo.ErrorCode` property not to be set. I've added a unit test to make sure that it is. However, in future I think we should update this code so that all the required properties of `RuleInfo` are mandated at compile time.


src/Rules/RuleRunner.cs
```diff
@@ -87,12 +87,7 @@ private static RunResult RunRuleWorker(IRule rule, IA11yElement element)
             if (!ShouldRun(rule.Condition, element))
                 return result;
 
-            // The PassesTest call is the method we want to convert all rules to use
-            // Using RuleInfo.ErrorCode will allow us to use reflection to document the result of each rule
-            if (rule.Info.ErrorCode.HasValue)
-                result.EvaluationCode = rule.PassesTest(element) ? EvaluationCode.Pass : rule.Info.ErrorCode.Value;
-            else
-                result.EvaluationCode = rule.Evaluate(element);
+                result.EvaluationCode = rule.PassesTest(element) ? EvaluationCode.Pass : rule.Info.ErrorCode;
```
DaveTryon, line 90: Need to remove one level of indent here:
```suggestion
            result.EvaluationCode = rule.PassesTest(element) ? EvaluationCode.Pass : rule.Info.ErrorCode;
```

src/RulesTest/RuleRunnerTests.cs
```diff
@@ -341,11 +282,13 @@ public void RunAll_ReturnsExpectedResults()
 
         private static Mock<IRule> CreateRuleMock(Condition condition, EvaluationCode code, A11yElement e)
         {
-            var infoStub = new RuleInfo();
+            var infoStub = new RuleInfo {
+                ErrorCode = code
+        };
```
DaveTryon, line 287: Should be indented one level:
```suggestion
            };
```

src/RulesTest/RuleRunnerTests.cs
```diff
@@ -341,11 +282,13 @@ public void RunAll_ReturnsExpectedResults()
...
 
-            var ruleMock = new Mock<IRule>(MockBehavior.Strict);
+        var ruleMock = new Mock<IRule>(MockBehavior.Strict);
```
DaveTryon, line 289: Should be indented one level:
```suggestion
            var ruleMock = new Mock<IRule>(MockBehavior.Strict);
```

DaveTryon (changes requested):
Looks good overall, pointed out a few indentation issues

DaveTryon (approved):

PR #529, opened 2021-03-03
chore: Suppress warnings that VS 16.9.0 is suddenly generating

DaveTryon:
#### Details

VisualStudio 16.9.0 recently released, and builds with the new version are suddenly broken due to 10 new warnings that were not occurring before. We've reached out to the Visual Studio team to understand why VS suddenly decided to raise these warnings, but the quickest mitigation to get everyone unstuck is to suppress these warnings globally. 

##### Motivation

Allow the team to develop with VS 16.9.0

##### Context

This is a temporary mitigation to get unstuck. It comes with the understanding that we'll actually evaluate and maybe fix these issues going forward.

#### Pull request checklist

### DIFF
--- a/build/NetStandardRelease.targets
+++ b/build/NetStandardRelease.targets
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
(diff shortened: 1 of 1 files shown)


jalkire (approved):
Builds locally

PR #535, opened 2021-03-12
chore: Enable + fix CS8073 warnings by removing unnecessary null checks

jalkire:
#### Details
PR #529 disabled many warnings. This PR reenables the [CS8073 warning, which catches expressions that are always true or false](https://devblogs.microsoft.com/dotnet/automatically-find-latent-bugs-in-your-code-with-net-5/#warn-when-expression-is-always-true-or-false). The rule found a few null checks on Rectangles, which cannot be null, so I removed the null checks.

##### Motivation

##### Context
I considered switching these to check if Rectangle.IsEmpty == true, but that didn't make sense in context.

#### Pull request checklist


src/Rules/PropertyConditions/BoundingRectangle.cs
```diff
@@ -52,7 +52,6 @@ private static bool HasCorrectDataFormat(IA11yElement e)
         private static bool IsBoundingRectangleEmpty(IA11yElement e)
         {
             if (e == null) throw new ArgumentNullException(nameof(e));
-            if (e.BoundingRectangle == null) throw new ArgumentException(ErrorMessages.ElementBoundingRectangleNull, nameof(e));
 
             // Because IA11yElement.BoundingRectangle will return an empty rectangle when the property value is null,
```
DaveTryon, line 56: Since BoundingRectangle can never be null, should this comment also be removed?

jalkire: The way I read this comment, it seems to be saying we can't null check `IA11yElement.BoundingRectangle` here, which is correct. With that interpretation, the comment didn't make sense before this PR because we were attempting to null check, and with this PR now does make sense. Happy to remove it if you read the comment differently though!

DaveTryon: Might be best to run this by @RobGallo to see if he remembers the intent of the comment, then either reword or remove, as appropriate.

src/Rules/PropertyConditions/BoundingRectangle.cs
```diff
@@ -52,7 +52,6 @@ private static bool HasCorrectDataFormat(IA11yElement e)
         private static bool IsBoundingRectangleEmpty(IA11yElement e)
         {
             if (e == null) throw new ArgumentNullException(nameof(e));
-            if (e.BoundingRectangle == null) throw new ArgumentException(ErrorMessages.ElementBoundingRectangleNull, nameof(e));
```
DaveTryon, line 55: Should we remove the resource string for `ErrorMessages.ElementBoundingRectangleNull`, since it's an impossible condition?

src/Rules/PropertyConditions/BoundingRectangle.cs
jalkire: Removed the new line but left code comment in based on Rob's PR comment.

src/Rules/PropertyConditions/BoundingRectangle.cs
jalkire: Oh yeah, I meant to remove that! Thanks.

DaveTryon (approved):

PR #538, opened 2021-03-12
chore: Enable + fix CA1508 warnings by removing unused null checks

jalkire:
#### Details
PR #529 disabled many warnings. This PR reenables the [CA1508 warning, which detects dead conditional code](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/CA1508). I removed some null checks where the value could never be null. In one case, I suppressed the rule because it claimed the value couldn't be null, but I believe it was mistaken.

##### Motivation

##### Context
If you think `caughtException` can never be null in `ReturnValueChooser` line 24, let me know!

#### Pull request checklist


DaveTryon (approved):
I agree that the check in ReturnValueChooser.GetReturnValue is correct as written.

PR #552, opened 2021-03-17
chore: Enable and address CA1024 warnings

DaveTryon:
#### Details
PR #529 disabled many warnings. This PR reenables the [CA1024 warning](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1024), which nudges us towards properties, where applicable. The analyzer provides some good suggestions, but can be a bit noisy. A notable source of noise is things like `GetInstance` methods for singletons, something that occurs far more frequently in Axe.Windows than any of us would really like ☹️.

This PR might be simpler to review on a per-commit basis.

##### Motivation

##### Context

Unit tests are all happy.

This change will require a corresponding change in AIWIn when we pick up the new drop of Axe.Windows, since AIWin uses some of the methods that are being converted into properties

#### Pull request checklist


src/Actions/Actions/SelectAction.cs
```diff
@@ -376,22 +379,12 @@ public static SelectAction GetDefaultInstance()
...
             return sDefaultInstance;
         }
-
-        /// <summary>
-        /// Get an instance of SelectAction
-        /// this is for the case that you need 2nd or more instances of SelectAction. 
-        /// in most of cases, you can use GetDefaultInstance() for getting default SelectAction instance
-        /// </summary>
-        /// <returns></returns>
-        public static SelectAction GetInstance()
```
DaveTryon, line 391: This method isn't used anywhere outside of this class, so I streamlined it

src/Core/Bases/A11yElement.cs
```diff
@@ -638,14 +638,6 @@ IA11yElement IA11yElement.Parent
         /// </summary>
         public string IssueDisplayText { get; set; }
 
-        public ScanResults GetScannerResultInstance()
```
DaveTryon, line 641: This isn't used anywhere in Axe.Windows or AIWin

src/Desktop/UIAutomation/DesktopElementHelper.cs
```diff
@@ -1,66 +1,24 @@
...
 {
     /// <summary>
-    /// Class DesktopElementHelper
-    /// it is a helper class for DesktopElement to handle property/pattern related caching and more
-    /// it is a singleton object via static members
+    /// A helper class to manage the cached properties that we always request from UIA
     /// </summary>
     public static class DesktopElementHelper
     {
-
```
DaveTryon, line 17: These changes mirror those in #550. I'll plan on merging #550 first, then handle the merge into this PR

jalkire (approved):

PR #557, opened 2021-03-24
chore: Sort using blocks (done by a tool)

DaveTryon:
#### Details

Our using blocks have been left in a disarray after multiple renames that have occurred over the last year or two. This PR sorts the using blocks. It was done via a tool. A couple of test files that didn't previously end with a newline now have the newline--I left them that way since those files were outliers and normalizing them would make future tool-driven changes simpler.

##### Motivation

Code hygiene

##### Context

@RobGallo is on board with this change (we discussed it earlier today). 

This PR does not remove unneeded using blocks. VS is smart enough to handle this, but the tool that I created isn't.

Also, this makes no attempt to sort `using static` blocks or using blocks that are used as aliases.

#### Pull request checklist

### DIFF
--- a/src/DesktopTests/ColorContrastAnalyzer/ColorTests.cs
+++ b/src/DesktopTests/ColorContrastAnalyzer/ColorTests.cs
@@ -1,8 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Desktop.ColorContrastAnalyzer;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
-using Axe.Windows.Desktop.ColorContrastAnalyzer;
 
 namespace Axe.Windows.DesktopTests.ColorContrastAnalyzer
 {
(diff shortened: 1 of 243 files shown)


src/DesktopTests/ColorContrastAnalyzer/ColorTests.cs
```diff
@@ -49,4 +49,4 @@ public void LuminanceOf_Black()
             Assert.AreEqual(0, new Color(0, 0, 0).Luminance());
         }
     }
-}
\ No newline at end of file
+}
```
DaveTryon, line 52: This is the outlier file that didn't previously end with a newline

src/DesktopTests/ColorContrastAnalyzer/ImageTests.cs
```diff
@@ -190,4 +190,4 @@ public void BinaryRowSearchIterator()
             Assert.AreEqual(0, rowOrder.Count(), "We should have inspected all expected rows.");
         }
     }
-}
\ No newline at end of file
+}
```
DaveTryon, line 193: This is one of the outlier files that didn't previously end with a newline

jalkire (approved):
Glad to see this!

Issue #554, opened 2021-03-22
[Feature Request] Warn users when application is written in a Java framework that doesn't provide UIA information


karanbirsingh:
## Is your feature request related to a problem? Please describe.
When users test Java applications that don't provide meaningful UIA information (IntelliJ is an example), Axe.Windows & Accessibility Insights for Windows can show 0 errors. However, the UIA tree will show that very little UIA information was exposed. See this [FAQ question](https://accessibilityinsights.io/docs/en/windows/reference/faq/#can-i-use-accessibility-insights-for-windows-on-a-windows-app-written-with-java). Depending on how robustly we can detect these scenarios, it would be nice to add some warning, test result, or info in the application that explains this scenario to the user and encourages manual NVDA / JAWS testing.

One entry point could be using the classname to identify the framework. But maybe there are other solutions that use heuristics like empty UIA trees.

## Describe the solution you'd like
A user who encounters this situation should be given information about it inside the library/application.

## Describe alternatives you've considered
Currently users either:
- see the 0 errors and assume nothing is wrong, or
- see the empty UIA tree and need to debug / find more information

## Additional context
I'm not sure how "sure" we can be. We've seen this happen with the SunAwtFrame classname a few times, but I haven't explored whether it's possible to add UIA support a different way while maintaining that framework.

chingucoding:
A way to somewhat detect this would be to have the UIA tree at least contain a certain amount of elements, e.g. a window that just has 2 elements is very likely to be entirely inaccessible since nothing is exposed.

RobGallo:
I believe this feature request to have been address by PR #558

karanbirsingh:
released in https://github.com/microsoft/axe-windows/releases/tag/v1.0.7

PR #558, opened 2021-03-26
fix: Add unsupported framework rule

DaveTryon:
#### Details

Issue #554 requests a way to warn users if they're using a framework that doesn't support UIA. We don't yet have a good way to do this in live mode, but a first step is to add an automated check that triggers an error based on the Windows class name. So far, the cases we've encountered are Win32 frameworks that return class names of `SunAwtFrame` or `SunAwtDialog`. Other classes could exist, so we've opted to raise the error for any Win32 object with a reported ClassName of `SunAwt*`. We use a case-sensitive regular expression here.

##### Motivation

Issue #554 

##### Context

The HowToFix text may be a little too blunt. It says "Switch your application to a framework that supports UI Automation.". I'm open to gentler ways to express this sentiment of "Sorry, you chose the wrong framework".

#### Pull request checklist

### DIFF
--- a/src/Core/Misc/ExtensionMethods.cs
+++ b/src/Core/Misc/ExtensionMethods.cs
@@ -647,9 +647,11 @@ namespace Axe.Windows.Core.Misc
         {
             if (e == null) return null;
 
-            return (string.IsNullOrEmpty(e.Framework))
+            string framework = e.Framework;
+
+            return (string.IsNullOrEmpty(framework))
                 ? GetUIFramework(e.Parent)
-                : e.Framework;
+                : framework;
         }
 
         /// <summary>
--- a/src/Core/Misc/ExtensionMethods.cs
+++ b/src/Core/Misc/ExtensionMethods.cs
@@ -647,6 +647,7 @@ namespace Axe.Windows.Core.Misc
         {
             if (e == null) return null;
 
+            // Only query Framework property once to simplify unit test setup 
             string framework = e.Framework;
 
             return (string.IsNullOrEmpty(framework))
--- a/src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
+++ b/src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
@@ -4,6 +4,7 @@ using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
 using Axe.Windows.Rules.Resources;
 using System.Text.RegularExpressions;
+using static Axe.Windows.Rules.PropertyConditions.ControlType;
 using static Axe.Windows.Rules.PropertyConditions.Framework;
 
 namespace Axe.Windows.Rules.Library
@@ -11,7 +12,7 @@ namespace Axe.Windows.Rules.Library
     [RuleInfo(ID = RuleId.FrameworkDoesNotSupportUIAutomation)]
     class FrameworkDoesNotSupportUIAutomation : Rule
     {
-        private readonly static string[] KnownProblematiClasses =
+        private readonly static string[] KnownProblematicClasses =
         {
             @"^\s*SunAwt.*$",
         };
@@ -28,7 +29,7 @@ namespace Axe.Windows.Rules.Library
         {
             string className = e.ClassName;
 
-            foreach (var knownProblematicClass in KnownProblematiClasses)
+            foreach (var knownProblematicClass in KnownProblematicClasses)
             {
                 if (Regex.IsMatch(className, knownProblematicClass, RegexOptions.CultureInvariant))
                     return false;
@@ -39,7 +40,7 @@ namespace Axe.Windows.Rules.Library
 
         protected override Condition CreateCondition()
         {
-            return Win32Framework;
+            return Win32Framework & Window;
         }
     }
 }
--- a/src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
+++ b/src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
+using Axe.Windows.Rules.Resources;
+using System.Linq;
+using static Axe.Windows.Rules.PropertyConditions.Framework;
+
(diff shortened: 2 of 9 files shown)


src/Core/Misc/ExtensionMethods.cs
```diff
@@ -647,9 +647,11 @@ public static string GetUIFramework(this IA11yElement e)
         {
             if (e == null) return null;
 
-            return (string.IsNullOrEmpty(e.Framework))
+            string framework = e.Framework;
+
+            return (string.IsNullOrEmpty(framework))
```
DaveTryon, line 652: While running the unit tests, I noticed that e.Framework was being called twice. I had the choice to tell the unit tests to double their expected number of calls, or change the Extension method to only query the property once. The latter seemed like the better choice. Happy to discuss if you wish.

src/Core/Misc/ExtensionMethods.cs
```diff
@@ -647,9 +647,11 @@ public static string GetUIFramework(this IA11yElement e)
         {
             if (e == null) return null;
 
-            return (string.IsNullOrEmpty(e.Framework))
```
RobGallo, line 650: Not quite sure why this change is here.

src/Core/Misc/ExtensionMethods.cs
RobGallo: I think this needs a comment in the code.

src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
```diff
@@ -0,0 +1,45 @@
...
+using Axe.Windows.Rules.Resources;
+using System.Text.RegularExpressions;
+using static Axe.Windows.Rules.PropertyConditions.Framework;
+
+namespace Axe.Windows.Rules.Library
+{
+    [RuleInfo(ID = RuleId.FrameworkDoesNotSupportUIAutomation)]
+    class FrameworkDoesNotSupportUIAutomation : Rule
+    {
+        private readonly static string[] KnownProblematiClasses =
```
RobGallo, line 14: Spelling

src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
```diff
@@ -0,0 +1,45 @@
...
+                if (Regex.IsMatch(className, knownProblematicClass, RegexOptions.CultureInvariant))
+                    return false;
+            }
+
+            return true;
+        }
+
+        protected override Condition CreateCondition()
+        {
+            return Win32Framework;
```
RobGallo, line 42: Wondering if we want to narrow this down a little further so this test doesn't get run all the time. For example, are the SAWT* class types usually only on elements with a control type of Window or Pane?

src/Rules/Resources/Descriptions.resx
```diff
@@ -426,4 +426,7 @@
   <data name="ClickablePointOffScreen" xml:space="preserve">
     <value>An element's IsOffScreen property must be true when its clickable point is off screen.</value>
   </data>
+  <data name="FrameworkDoesNotSupportUIAutomation" xml:space="preserve">
+    <value>The framework in use does not support UI Automation.</value>
```
RobGallo, line 430:
```suggestion
    <value>The framework used to build this application does not support UI Automation.</value>
```

src/RulesTest/Library/FrameworkDoesNotSupportUIAutomationTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+        [TestMethod]
+        public void Condition_FrameworkIsNotWin32_ReturnsFalse()
+        {
+            string[] nonWin32Values = { FrameworkId.DirectUI, FrameworkId.Edge, FrameworkId.InternetExplorer, FrameworkId.WinForm, FrameworkId.WPF, FrameworkId.XAML, "NotWin32" };
+
+            foreach (string nonWin32Value in nonWin32Values)
+            {
+                _elementMock.Setup(m => m.Framework)
+                    .Returns(FrameworkId.DirectUI)
```
RobGallo, line 32: Don't you intend to return `nonWin32Value`?

src/RulesTest/Library/FrameworkDoesNotSupportUIAutomationTests.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            _elementMock.Setup(m => m.Framework).Returns(FrameworkId.Win32).Verifiable();
+
+            Assert.IsTrue(Rule.Condition.Matches(_elementMock.Object));
+
+            _elementMock.Verify(m => m.Framework, Times.Once());
+        }
+
+        [TestMethod]
+        public void PassesTest_FrameworkIsWin32_ClassNameIsNotKnownProblematic_ReturnsFalse()
```
RobGallo, line 52: I think "IsWin32" in the name is not necessary, and could easily become out-of-date.

RobGallo (changes requested):

src/Rules/Resources/Descriptions.resx
DaveTryon: Fixed

src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
DaveTryon: Narrowed the scope to just Windows (this is the only case that occurs in the sample files that we've received)

src/Rules/Library/FrameworkDoesNotSupportUIAutomation.cs
DaveTryon: Fixed

src/Core/Misc/ExtensionMethods.cs
DaveTryon: Comment added

src/RulesTest/Library/FrameworkDoesNotSupportUIAutomationTests.cs
DaveTryon: Yes. The first version had a hardcoded value and I forgot to update it when I added the loop

src/RulesTest/Library/FrameworkDoesNotSupportUIAutomationTests.cs
DaveTryon: Removed (here and in the case that returns true)

RobGallo (approved):

PR #561, opened 2021-03-30
fix: Exclude some classes from name required condition

DaveTryon:
#### Details

Issue #556 identifies some cases where we should not require the Name property because it adds little value and/or is difficult to set. This PR revises the `NameRequired` Condition to exclude these cases. Tested by running the CLI (with a time delay that will be in a separate PR) against AIWin with the context menu expanded (the same scenario that was called out in #556).

##### Motivation

Address issue #556

##### Context

The consequential changes here are very small. The `NamesTest` class had many lines where indents were off, and I went ahead and fixed these in this PR. The only consequential changes in the `NamesTest` class appear at the very end of the file, and I've marked them with a comment. I can split the indent change into a separate PR if that seems helpful--just let me know.

#### Pull request checklist

### DIFF
--- a/src/RulesTest/PropertyConditions/NameTest.cs
+++ b/src/RulesTest/PropertyConditions/NameTest.cs
@@ -283,7 +283,7 @@ namespace Axe.Windows.RulesTests.PropertyConditions
         [TestMethod]
         public void TestExcludedClassNames_ClassNamesDoNotMatch()
         {
-            string[] nonExcludedClassNames = { "NotPopup", "ContextMenuIsNotMyName" };
+            string[] nonExcludedClassNames = { null, "NotPopup", "ContextMenuIsNotMyName" };
 
             using (var e = new MockA11yElement())
             {


src/RulesTest/PropertyConditions/NameTest.cs
```diff
@@ -279,5 +279,21 @@ public void TestTextWithDisallowedPlatformProperties()
                 Assert.IsTrue(Misc.NameRequired.Matches(e));
             } // using
         }
+
```
DaveTryon, line 282: Consequential change to this class begin at this point

src/Rules/PropertyConditions/ElementGroups.cs
```diff
@@ -212,16 +214,16 @@ private static Condition CreateIsControlOptionalCondition()
...
-        * And for many, it's easy to imagine it being difficult to come up with a meaningful name beyond the control type.
-        * What else do you call a StatusBar for example?
-        * Custom elements are handled separately because there may be times
-        * when a name is "page 1" for example and the type is "page"
-        * which we currently believe is acceptable because screen readers will read the name then the type
-        * and "1 page" would sound awkward.
-        * Names that are particularly long are viewed as acceptable because they are
-        * more likely to be informative and not to cause double speaking by screen readers.
-        * */
+            * And therefore don't benefit from name-based distinction.
```
DaveTryon, line 217: This block of changes is just the indents

src/RulesTest/PropertyConditions/NameTest.cs
```diff
@@ -22,20 +22,20 @@ public NameTest()
             RequiredTypes = new List<int>
             {
                 Calendar, CheckBox, ComboBox,
-               DataGrid, DataItem, Document,
```
DaveTryon, line 25: With the exception of the marked block at the end of this file, these change are all correcting indentation issues

src/Rules/PropertyConditions/ElementGroups.cs
```diff
@@ -125,6 +125,7 @@ private static Condition CreateNameRequiredCondition()
             var trueWhenElementHasSiblingsOfSameType = (Header | StatusBar) & SiblingsOfSameType;
             var allowedImage = Image & NotParent(Button | ListItem | MenuItem | TreeItem);
             var allowedText = Text & Condition.Create(HasAllowedPlatformPropertiesForText);
+            var excludedByClassName = WPF & (StringProperties.ClassName.Is("Popup") | StringProperties.ClassName.Is("ContextMenu"));
```
RobGallo, line 128: If I had my druthers, I'd prefer a name which describes the match like `WPFPopupsAndContextMenus`.

src/RulesTest/PropertyConditions/NameTest.cs
```diff
@@ -279,5 +279,21 @@ public void TestTextWithDisallowedPlatformProperties()
...
         }
+
+        [TestMethod]
+        public void TestExcludedClassNames()
+        {
+            string[] excludedClassNames = { "Popup", "ContextMenu" };
+
+            using (var e = new MockA11yElement())
+            {
+                e.Framework = FrameworkId.WPF;
```
RobGallo, line 290: What do you think about adding a test that should assert true before you set the class names? To me, it at least indicates that it's the class names causing the condition not to match.

DaveTryon: Added both positive and negative tests. I set the ControlType as Window to satisfy the class condition

src/Rules/PropertyConditions/ElementGroups.cs
DaveTryon: Renamed the local variable to `isWPFPopupOrContextMenu`

src/RulesTest/PropertyConditions/NameTest.cs
```diff
@@ -281,17 +281,35 @@ public void TestTextWithDisallowedPlatformProperties()
         }
 
         [TestMethod]
-        public void TestExcludedClassNames()
+        public void TestExcludedClassNames_ClassNamesDoNotMatch()
+        {
+            string[] nonExcludedClassNames = { "NotPopup", "ContextMenuIsNotMyName" };
```
RobGallo, line 286: What about a null class name?

PR #562, opened 2021-03-30
chore: Add delayed triggering to CLI scanner

DaveTryon:
#### Details

The current CLI doesn't provide a way to scan transient controls. This PR adds the ability to delay up to 60 seconds to allow the user to get the app UI into the correct configuration for scanning.

##### Motivation

I needed this ability for testing #561, then realized that it might be useful in general.

##### Context

To facilitate unit testing, I implemented the 1 second delay via an `Action` that gets passed in from the Program. 

I changed `OutputGeneratorUnitTests` to use a test fixture. I was going to have to modify the constructor anyway, so I just went ahead and moved to a test fixture while I was at it.

#### Pull request checklist

### DIFF
--- a/src/CLI/OptionsEvaluator.cs
+++ b/src/CLI/OptionsEvaluator.cs
@@ -36,7 +36,7 @@ namespace AxeWindowsCLI
             else
             {
 #pragma warning disable CA1303 // Do not pass literals as localized parameters
-                throw new ParameterException("Please specify either processId or processName on the command line");
+                throw new ParameterException("Please specify either processId or processName on the command line.");
 #pragma warning restore CA1303 // Do not pass literals as localized parameters
             }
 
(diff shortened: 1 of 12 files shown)


src/CLITests/OptionsUnitTests.cs
```diff
@@ -19,16 +19,11 @@ public class OptionUnitTests
         const string VerbosityKey = "--verbosity";
         const string ScanIdKey = "--scanid";
         const string ShowThirdPartyNoticesKey = "--showthirdpartynotices";
+        const string DelayInSecondsKey = "--delayinseconds";
 
         const string TestProcessName = "MyProcess";
         const int TestProcessId = 42;
 
-        private int FailIfCalled(Options options)
```
DaveTryon, line 26: Removing unused method

src/CLI/OptionsEvaluator.cs
```diff
@@ -28,7 +36,7 @@ public static IOptions ProcessInputs(Options rawInputs, IProcessHelper processHe
             else
             {
 #pragma warning disable CA1303 // Do not pass literals as localized parameters
-                throw new ParameterException("Please specify either processId or processName on the command line");
+                throw new ParameterException("Please specify either processId or processName on the command line.");
```
DaveTryon, line 39: Punctuation counts!

DaveTryon:
@RobGallo, the "no delay in quiet mode" restriction was partially an artifact of where I plumbed in the ScanDelay, plus I couldn't identify a scenario where it made sense. Stepping back, I get a cleaner design by refactoring the plumbing, and I've removed the restriction.

RobGallo (approved):

PR #563, opened 2021-04-01
chore: Normalize test namespaces

DaveTryon:
#### Details

If you use the IDE to add a new test to one of our test projects, the test class ends up in a different namespace that other test classes in that folder. In addition, our current test namespaces are very inconsistent. This PR normalizes the namespaces and ensures that IDE-added classes follow the same pattern.

##### Motivation

Make it easier to work with tests (both creating and viewing the results) in the IDE.

##### Context

To test this, I added a new test class to each test assembly and confirmed that the new tests get added in the correct namespace. I never committed these files. I also broke up the commits per project, in case there's any desire to review these individually.

#### Pull request checklist

### DIFF
--- a/src/CLITests/CLITests.csproj
+++ b/src/CLITests/CLITests.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>AxeWindowsCLITests</RootNamespace>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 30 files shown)


src/CLITests/CLITests.csproj
```diff
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>AxeWindowsCLITests</RootNamespace>
```
dbjorge, line 5: Should this be `Axe.Windows.CLITests` for consistency with the rest?

DaveTryon: The CLI namespace is AxeWindowsCLI, so this is AxeWindowCLITests.

dbjorge (approved):

PR #565, opened 2021-04-07
Fix: Issue 540

RobGallo:
#### Details

Address [BUG] ComboBox "An element's BoundingRectangle must not obscure its container element" · Issue #540

##### Context

The light dismiss button is not intended to ever have focus and has little to no value for magnification users. It is also designed specifically to eclipse its parent window. Therefore, we will not check for BoundingRectangleCompletelyObscuresContainer on the light dismiss button.

### DIFF
--- a/src/Rules/Library/BoundingRectangleCompletelyObscuresContainer.cs
+++ b/src/Rules/Library/BoundingRectangleCompletelyObscuresContainer.cs
@@ -36,7 +36,7 @@ namespace Axe.Windows.Rules.Library
         protected override Condition CreateCondition()
         {
             // Windows and dialogs can be any size, regardless of their parents
-var isDialog = Pane & IsDialog;
+            var isDialog = Pane & IsDialog;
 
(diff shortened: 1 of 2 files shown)


src/Rules/Library/BoundingRectangleCompletelyObscuresContainer.cs
```diff
@@ -34,16 +36,19 @@ public override bool PassesTest(IA11yElement e)
         protected override Condition CreateCondition()
         {
             // Windows and dialogs can be any size, regardless of their parents
+var isDialog = Pane & IsDialog;
```
dbjorge, line 39: nit: it looks like L39 lost its indentation by accident

DaveTryon: Revised line:
```suggestion
            var isDialog = Pane & IsDialog;
```

dbjorge (approved):

PR #573, opened 2021-04-14
feat: Allow CLI to decouple screen captures from System.Drawing.Bitmap

DaveTryon:
#### Details

We know that Windows 10X doesn't include GDIPlus.dll, which means that we can't rely on `System.Drawing.Bitmap` to represent the screen capture (the version in `System.Drawing.Common` is just a port of `System.Drawing` and still relies on GDIPlus.dll). Creating one-off screen capture tech isn't part of our charter, but we can create an abstraction that will allow others to implement screen capture if they desire. With that in mind, here are the basic steps in this PR:

1. Create a pair of interfaces to decouple screen captures from `System.Drawing.Bitmap`. These interfaces represent all of the `System.Drawing.Bitmap` functionality needed within the CLI's operation:
  -  `Axe.Windows.SystemAbstractions.IBitmapCreator` implements a `FromScreenRectangle` method, which returns an `Axe.Windows.SystemAbstractions.IBitmap` object.
  - `Axe.Windows.SystemAbstractions.IBitmap` implements a `SavePngToStream` method. 
2. Create a pair of classes (`Axe.Windows.Desktop.IBitmapCreator` and `Axe.Windows.Desktop.IBitmap`) to provide implementations of the abstractions, using `System.Drawing.Bitmap` inside the abstraction.
3. Replace the `System.Drawing.Bitmap` in the `DataContext` class with an `IBitmap`. 
4. Update other pieces in Axe.Windows to be able to accommodate the change in `DataContext`. This meant adding a new method to each of the interfaces--`IBitmap` gets a `ToSystemDrawingBitmap` method to expose a `System.Drawing.Bitmap` from an `IBitmap`, and `IBitmapCreator` gets a `FromSystemDrawingBitmap` method to create an `IBitmap` from a `System.Drawing.Bitmap`. Crucially, neither of these last 2 methods are used by the CLI.
5. Update the CLI to allow a user to specify an assembly that exposes a public implementation of `IBitmapCreator`. We then use reflection to find the implementation, create it, and use it for the screen capture.
6. (not in this PR). Because of the change in the `DataContext` class, we'll need to update AIWin to call `IBitmap.ToSystemDrawingBitmap` and `IBitmapCreator.FromSystemDrawingBitmap` in a few places.

##### Motivation

Enable Windows 10X team to implement screen capture as an independent component.

##### Context

The PrivacyExtensions create the "yellow box" screen representation that we currently use as a fallback when loading a file without a screenshot. I considered using `IBitmap` instead of `System.Drawing.Bitmap` within that class, but didn't like the result, and since this code is only used when _loading_ a file (i.e., not when creating it), I decided to keep `IBitmap` as small as possible.

As called out above, there will need to be a separate PR in AIWin when we pick up the version of Axe.Windows that contains these changes.

#### Pull request checklist

### DIFF
--- a/src/Actions/Actions/DataManager.cs
+++ b/src/Actions/Actions/DataManager.cs
@@ -5,9 +5,9 @@ using Axe.Windows.Actions.Contexts;
 using Axe.Windows.Actions.Enums;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Misc;
+using Axe.Windows.SystemAbstractions;
 using System;
 using System.Collections.Generic;
-using System.Drawing;
 using System.Linq;
 
 namespace Axe.Windows.Actions
@@ -151,7 +151,7 @@ namespace Axe.Windows.Actions
         /// </summary>
         /// <param name="ecId"></param>
         /// <returns></returns>
-        public Bitmap GetScreenshot(Guid ecId)
+        public IBitmap GetScreenshot(Guid ecId)
         {
             var elementContext = GetElementContext(ecId);
             if (elementContext == null) return null;
--- a/src/Actions/Actions/LoadAction.cs
+++ b/src/Actions/Actions/LoadAction.cs
@@ -3,7 +3,9 @@
 using Axe.Windows.Actions.Attributes;
 using Axe.Windows.Actions.Enums;
 using Axe.Windows.Core.Bases;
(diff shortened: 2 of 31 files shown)


DaveTryon:
After discussion, we're not going to merge this at this time. We'll keep the branch available should we later choose to enable this ability

Issue #572, opened 2021-04-05
[Bug] AI for Windows flags error for Button content containing a private Unicode character despite having a separate AutomationProperties.Name set


cprivera:
AI Windows reports an error that "The Name property must not contain any characters in the private Unicode range" for buttons with private Unicode characters set as their content despite the button having its AutomationProperties.Name property already set.

Each image here shows the button in question, with the error flagged on the 'text' inside the button. You can see the name property set for both buttons here.
<img width="1127" alt="03-31-2021" src="https://user-images.githubusercontent.com/25541708/113605065-03527e80-95fb-11eb-8a7e-66363c414c45.png">

<img width="1128" alt="04-05-2021" src="https://user-images.githubusercontent.com/25541708/113605524-a3100c80-95fb-11eb-9ef9-e5a693af4785.png">

Result: AI Windows flags error "The Name property must not contain any characters in the private Unicode range" for text inside the button
Expected: The error should not be flagged because, according to [this documentation](https://docs.microsoft.com/en-us/accessibility-tools-docs/items/WPF/Button_Name), having AutomationProperties.Name set should override using the content of the button as the name.

I will note that in each case the button does not have a nested text block or anything, simply the content of the button is set to some character that we use as an icon to indicate the button's purpose.

DaveTryon:
Buttons are handled differently between .NET Framework and .NET Core. In .NET Framework, the Button element is a leaf node in UIA. In .NET Core (versions 3.1, 5.0, and 6.0 preview 3), a Button's text appears in UIA as a child element. The current rules aren't aware of this difference, so they are perceived as 2 separate controls, instead of 2 portions of the same element. We've reached out to the .NET Core accessibility team to try to better understand why .NET Core includes this new behavior, and will come up with a plan once we have a clearer picture of what our options are.

DaveTryon:
Ok, the problem is understood and a fix is in the works. I'm attaching a zip file of a test app for easy reference.
[TestApp.zip](https://github.com/microsoft/axe-windows/files/6423603/TestApp.zip)

DaveTryon:
This fix is in the code, but it hasn't yet been released

DaveTryon:
[...]

PR #575, opened 2021-05-04
fix: Exclude button TextBlocks from NameExcludesPrivateUnicodeCharacters rule

DaveTryon:
#### Details

By default, a WPF Button will use its Content as its UIA Name. As called out in #572, we're not taking this into account when checking for characters in the private Unicode range. This change creates a new Condition that is satisfied if all of the following are true:
1. The control type is Text
2. The control's parent is type Button

We then omit elements that satisfy this new condition from the `NameExcludesPrivateUnicodeCharacters` rule.

I've included an a11ytest file showing the results (compare this to the a11ytest file in #572)

[TestApp-Fixed.a11ytest.zip](https://github.com/microsoft/axe-windows/files/6428327/TestApp-Fixed.a11ytest.zip)

##### Motivation

Address #572 

##### Context

- These checks are framework-agnostic, on the assumption that the new Condition is already tightly scoped and that similar constructs could appear in any framework.
- There may be other rules where this new Condition should be taken into account, but I didn't try to identify or address them here. The Condition is easy to apply to other rules if/when they're identified.

#### Pull request checklist

- [x] Addresses an existing issue: #572


RobGallo (approved):
LGTM!

PR #576, opened 2021-05-06
chore: Specify culture invariant

DaveTryon:
#### Details

Specify CultureInfo.InvariantCulture when rendering strings.

##### Motivation

This is part of the GlobalReadiness work for the compliance feature. Only 1 change (called out in the code) is likely to have any real impact to the shipping code.

#### Pull request checklist

### DIFF
--- a/src/Core/Bases/A11yProperty.cs
+++ b/src/Core/Bases/A11yProperty.cs
@@ -5,6 +5,7 @@ using Axe.Windows.Core.Types;
 using Axe.Windows.Win32;
 using Newtonsoft.Json;
 using System;
+using System.Globalization;
 
 namespace Axe.Windows.Core.Bases
 {
@@ -147,7 +148,7 @@ namespace Axe.Windows.Core.Bases
             if ((double)arr[2] < 0 || (double)arr[3] < 0)
             {
                 // the 3rd and 4th values in array are negative value, we need to show value in different format like l,t,w,h
-                text = string.Format("[l={0},t={1},w={2},h={3}]", arr[0], arr[1], arr[2], arr[3]);
+                text = string.Format(CultureInfo.InvariantCulture, "[l={0},t={1},w={2},h={3}]", arr[0], arr[1], arr[2], arr[3]);
             }
             else
             {
--- a/src/Core/Types/PatternType.cs
+++ b/src/Core/Types/PatternType.cs
@@ -80,7 +80,7 @@ namespace Axe.Windows.Core.Types
 
             sb.Replace("UIA_", "");
             sb.Replace("Id", "");
-            //sb.AppendFormat("({0})", id); // skip Id for now
+            // sb.Append(Invariant($" ({id})")); // skip id for now
 
             return sb.ToString();
         }
--- a/src/Core/Types/PlatformPropertyType.cs
+++ b/src/Core/Types/PlatformPropertyType.cs
@@ -51,7 +51,7 @@ namespace Axe.Windows.Core.Types
 
             sb.Replace("Platform_", "");
             sb.Replace("PropertyId", "");
-            //sb.AppendFormat("({0})", id);
+            // sb.Append(Invariant($" ({id})"));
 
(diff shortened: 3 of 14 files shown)


src/Core/Bases/A11yProperty.cs
```diff
@@ -147,7 +148,7 @@ private string GetBoundingRectangleText()
             if ((double)arr[2] < 0 || (double)arr[3] < 0)
             {
                 // the 3rd and 4th values in array are negative value, we need to show value in different format like l,t,w,h
-                text = string.Format("[l={0},t={1},w={2},h={3}]", arr[0], arr[1], arr[2], arr[3]);
+                text = string.Format(CultureInfo.InvariantCulture, "[l={0},t={1},w={2},h={3}]", arr[0], arr[1], arr[2], arr[3]);
```
DaveTryon, line 151: This is the ship-impacting change. By default, the numbers in the string should pick up culture-specific separators. With CultureInfo.InvariantCulture, all separators should be skipped.

src/Core/Types/PatternType.cs
```diff
@@ -80,7 +80,7 @@ protected override string GetNameInProperFormat(string name, int id)
 
             sb.Replace("UIA_", "");
             sb.Replace("Id", "");
-            //sb.AppendFormat("({0})", id); // skip Id for now
+            // sb.Append(Invariant($" ({id})")); // skip id for now
```
DaveTryon, line 83: Updating the commented out code to follow the same pattern used in other types.

src/Core/Types/PlatformPropertyType.cs
```diff
@@ -51,7 +51,7 @@ protected override string GetNameInProperFormat(string name, int id)
 
             sb.Replace("Platform_", "");
             sb.Replace("PropertyId", "");
-            //sb.AppendFormat("({0})", id);
+            // sb.Append(Invariant($" ({id})"));
```
DaveTryon, line 54: Updating the commented out code to follow the same pattern used in other types.

src/Desktop/Styles/AnimationStyle.cs
```diff
@@ -61,6 +61,7 @@ protected override string GetNameInProperFormat(string name, int id)
 
             sb.Replace(Prefix, "");
             sb.Append(Invariant($" ({id})"));
+
```
DaveTryon, line 64: Newline added to match the formatting used in most styles.

src/Desktop/Styles/BulletStyle.cs
```diff
@@ -60,6 +60,7 @@ protected override string GetNameInProperFormat(string name, int id)
 
             sb.Replace(Prefix, "");
             sb.Append(Invariant($" ({id})"));
+
```
DaveTryon, line 63: Newline added to match the formatting used in most styles.

src/Desktop/Styles/CapStyle.cs
```diff
@@ -61,6 +61,7 @@ protected override string GetNameInProperFormat(string name, int id)
 
             sb.Replace(Prefix, "");
             sb.Append(Invariant($" ({id})"));
+
```
DaveTryon, line 64: Newline added to match the formatting used in most styles.

src/Core/Bases/A11yProperty.cs
RobGallo: Does it then make sense to use the current culture?

DaveTryon: Upon further investigation, separators are not impacted by the current culture, but negative signs definitely are. This is [documented](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.numberformatinfo.negativesign?view=net-5.0) and I confirmed it via a small test app. For a data file format, we should be using culture-agnostic formats both when reading and when writing, so a file created in culture X can be opened in culture Y.

RobGallo (approved):

PR #577, opened 2021-05-06
feat(compliance-may21): Use resource strings for AxeWindowsCLI

DaveTryon:
#### Details

Use resource strings for AxeWindowsCLI

##### Motivation

Global readiness. We aren't planning localization at this time, but this opens the door to localize the CLI should we later choose to do so.

##### Context

- This is limited in scope to just the CLI
- This doesn't yet use resources for the command line help text. I've found a sample that I might be able to follow, but it feels like a separate PR.

#### Pull request checklist


src/CLI/CLI.csproj
```diff
@@ -21,12 +21,6 @@
     <AxeDigitalSign>true</AxeDigitalSign>
   </PropertyGroup>
 
-  <PropertyGroup>
```
DaveTryon, line 24: This should have been removed previously, but we missed it

src/CLI/CLI.csproj
```diff
@@ -55,6 +49,21 @@
     <ProjectReference Include="..\Automation\Automation.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
DaveTryon, line 52: This block is boilerplate that Visual Studio added when adding the resource file to the project

src/CLI/OptionsEvaluator.cs
```diff
@@ -18,7 +19,7 @@ public static IOptions ProcessInputs(Options rawInputs, IProcessHelper processHe
 
             if (delayInSeconds < 0 || delayInSeconds > 60)
             {
-                throw new ParameterException(string.Format(CultureInfo.InvariantCulture, "Invalid delay: {0}. Please enter an integer value from 0 to 60.", delayInSeconds));
+                throw new ParameterException(string.Format(CultureInfo.CurrentCulture, DisplayStrings.ErrorInvalidDelayFormat, delayInSeconds));
```
DaveTryon, line 22: Any user-displayable messages will need CurrentCulture if/when localization is added, so switching to them now

src/CLI/OutputGenerator.cs
```diff
@@ -47,14 +48,14 @@ public void WriteBanner(IOptions options)
         public void WriteThirdPartyNoticeOutput(string pathToFile)
         {
             WriteAppBanner();
-            _writer.WriteLine("Opening Third Party Notices in default browser. If this fails, manually open \"{0}\"", pathToFile);
+            _writer.WriteLine(DisplayStrings.ThirdPartyNoticeFormat, pathToFile);
```
DaveTryon, line 51: The writer will already use CurrentCulture under the hood (it's Console.Out in production code)

src/CLI/ScanDelay.cs
```diff
@@ -25,35 +27,35 @@ public void DelayWithCountdown(IOptions options)
...
-            if (options.VerbosityLevel > VerbosityLevel.Quiet)
-            {
-                _writer.WriteLine(format, count, PluralSuffix(count));
-            }
+            string format = SelectString(stringIfCountIsOne, stringIfCountIsNotOne, count);
+            ConditionallyWriteMessage(options, () => string.Format(CultureInfo.InvariantCulture, format, count));
         }
 
-        private void ConditionallyWriteMessage(IOptions options, string message)
+        private void ConditionallyWriteMessage(IOptions options, Func<string> getMessage)
```
DaveTryon, line 48: This now takes a Func so that we can consolidate the options check in one place but still only build the strings if we really use them. This last bit helps the unit tests to better validate the behavior.

src/CLI_Full/CLI_Full.csproj
```diff
@@ -27,6 +27,10 @@
     <Compile Include="..\CLI\*.cs" Link="*.cs (From CLI)" />
   </ItemGroup>
 
+  <ItemGroup>
```
DaveTryon, line 30: CLI_Full "piggybacks" on the code from the CLI project. Since the resources are in a separate folder (CLI\Resources), we explicitly include that folder. I wasn't able to get glob patterns to work here ☹️

waabid (approved):
LGTM

src/CLI_Full/CLI_Full.csproj
chingucoding: Oh didn't think about those files. Yes in that case, I think two simple lines are way better than a single complex one.

PR #578, opened 2021-05-11
feat(compliance-may-21): Use localizable text for CLI options

DaveTryon:
#### Details

We have no current plans to localize Axe.Windows, but we want to remove obstacles as part of our compliance feature. This change simply moves some hardcoded strings into a new resources class, then specifies those strings in that class when defining the options that CommandLineParser uses. I verified that modifications to the resource strings appeared as expected in the output, but I never committed those changes. I verified that the old and new option prompts are identical (ran them both, piped them to a disk file, compared the files)

##### Motivation

Reduce obstacles for localization

##### Context

This PR impacts just the help text that we specify in the Options. It does not modify the following, which are hardcoded in CommandLineParser's `SentenceBuilder` class:
- The prompt for the --help command
- The prompt for the --version command
- Messaging triggered by command line parsing (for example, if you specify an invalid command)

 CommandLineParser has a [demo project](https://github.com/commandlineparser/commandline/tree/master/demo/ReadText.LocalizedDemo) that shows how to override the `SentenceBuilder` with localizable strings, but it means replicating some code from CommandLineParser and could break in the future if CommandLineParser changed the interface (in fact, the demo project already doesn't build due to a method being added after the demo project was created). I chose to defer that work to the time where we decide that we need it. The link to the demo project is preserved in the PR record, so it should be easy to find again later.

#### Pull request checklist


src/CLI/CLI.csproj
```diff
@@ -55,13 +55,22 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>DisplayStrings.resx</DependentUpon>
     </Compile>
+    <Compile Update="Resources\OptionsHelpText.Designer.cs">
```
DaveTryon, line 58: I put these into a separate resource file for 2 reasons:
1. There was already a logical grouping that I wanted to preserve
2. These strings need to be public for CommandLineParser to use them. The existing strings can be internal.

waabid (approved):
LGTM

PR #580, opened 2021-05-12
chore: Display output of validation app

DaveTryon:
#### Details

We have a test app that logs to the console, but that output doesn't make it into the pipeline logs. Capture the output from the spawned process and display it in our test logs.

##### Motivation

Better visibility into what this test is doing in the pipeline

#### Pull request checklist


DaveTryon:
This didn't work. Closing for now

PR #581, opened 2021-05-12
chore: Updates based on a repo-wide code overview

DaveTryon:
#### Details

I took a little time to scan through _all_ of the files in the project and address various issues that I noticed. This included:
- Whitespace issues (indents, spaces instead of tabs, newlines that seemed out of place, etc.)
- Unneeded using blocks (Visual Studio color codes them)
- Tweak some comments for accuracy/consistency
- Remove unneeded initialization
- Fix a couple of typos
- Simplify declarations of `List` objects that use list initializers
- Make the CI project truly a dummy (this was a very bare-bones CLI that can go away now that the CLI exists. I've also confirmed that the axe-pipeline-samples repo does not point to this code)

I strongly suggest reviewing these one commit at a time. One caveat--a test timeout was incorrectly commented out at one point, then I had to fix it in a later commit.

##### Motivation

Take advantage of the opportunity to look across the entire code base.

##### Context

Things intentionally excluded:
- Our naming of tests is pretty inconsistent, but I haven't tried to address that here.
- Our `IDisposable` implementations are inconsistent. They all start with the boilerplate, but the don't all uncomment the TODO step that maps the finalizer to `Dispose(false)`. It might make sense to look more carefully at these, but that feels like something that 

#### Pull request checklist


src/Core/Bases/A11yElement.cs
```diff
@@ -380,8 +380,8 @@ private static dynamic ConvertValueIfNecessary(int id, Newtonsoft.Json.Linq.JArr
 
             return GetValueAsArray<int>(jArray);
         }
-#pragma warning disable CA1801 // Parameter id of method ConvertValueIfNecessary is never used.Remove the parameter or use it in the method body.
 
+#pragma warning disable CA1801 // Parameter id of method ConvertValueIfNecessary is never used.Remove the parameter or use it in the method body.
```
jalkire, line 384: While you're fixing comments, this could use a space
```suggestion
#pragma warning disable CA1801 // Parameter id of method ConvertValueIfNecessary is never used. Remove the parameter or use it in the method body.
```

jalkire (approved):
Approved with one minor suggestion

PR #587, opened 2021-06-01
feat(custom-uia): Configuration file spec

codeofdusk:
#### Details
This PR contains an initial configuration file format for specifying custom UIA properties for which Axe Windows should register.

##### Motivation

##### Context
This PR includes no functionality, just documentation on a possible config format. Code changes to follow in subsequent PRs.

#### Pull request checklist

### DIFF
--- a/docs/CustomUIA.md
+++ b/docs/CustomUIA.md
@@ -34,47 +34,47 @@ This example file contains definitions for various [Excel](https://docs.microsof
 {
   "properties": [
     {
-      "guid": "{4BB56516-F354-44CF-A5AA-96B52E968CFD}",
+      "guid": "4BB56516-F354-44CF-A5AA-96B52E968CFD",
       "programmaticName": "AreGridlinesVisible",
       "uiaType": "bool"
     },
     {
-      "guid": "{E244641A-2785-41E9-A4A7-5BE5FE531507}",
+      "guid": "E244641A-2785-41E9-A4A7-5BE5FE531507",
       "programmaticName": "CellFormula",
       "uiaType": "string"
     },
     {
-      "guid": "{626CF4A0-A5AE-448B-A157-5EA4D1D057D7}",
+      "guid": "626CF4A0-A5AE-448B-A157-5EA4D1D057D7",
       "programmaticName": "CellNumberFormat",
       "uiaType": "string"
     },
     {
-      "guid": "{312F7536-259A-47C7-B192-AA16352522C4}",
+      "guid": "312F7536-259A-47C7-B192-AA16352522C4",
       "programmaticName": "CommentReplyCount",
       "uiaType": "int"
     },
     {
-      "guid": "{7AAEE221-E14D-4DA4-83FE-842AAF06A9B7}",
+      "guid": "7AAEE221-E14D-4DA4-83FE-842AAF06A9B7",
       "programmaticName": "DataValidationPrompt",
       "uiaType": "string"
     },
     {
-      "guid": "{DFEF6BBD-7A50-41BD-971F-B5D741569A2B}",
+      "guid": "DFEF6BBD-7A50-41BD-971F-B5D741569A2B",
       "programmaticName": "HasConditionalFormatting",
       "uiaType": "bool"
     },
     {
-      "guid": "{29F2E049-5DE9-4444-8338-6784C5D18ADF}",
+      "guid": "29F2E049-5DE9-4444-8338-6784C5D18ADF",
       "programmaticName": "HasDataValidation",
       "uiaType": "bool"
     },
     {
-      "guid": "{1B93A5CD-0956-46ED-9BBF-016C1B9FD75F}",
+      "guid": "1B93A5CD-0956-46ED-9BBF-016C1B9FD75F",
       "programmaticName": "HasDataValidationDropdown",
       "uiaType": "bool"
     },
     {
-      "guid": "{F065BAA7-2794-48B6-A927-193DA1540B84}",
+      "guid": "F065BAA7-2794-48B6-A927-193DA1540B84",
       "programmaticName": "ViewType",
       "uiaType": "enum",
       "values": {
--- a/docs/CustomUIA.md
+++ b/docs/CustomUIA.md
@@ -1,0 +1,125 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT Licence. -->
+
+## Custom UI Automation properties
+Axe Windows supports the inspection of [custom UI Automation properties](https://docs.microsoft.com/en-gb/windows/win32/winauto/uiauto-custompropertieseventscontrolpatterns). The custom properties for which Axe Windows should register are described in a configuration file that is loaded once per session and applies to all inspected UI Automation elements.
+
+### Configuration file format
+Custom UI Automation configuration data is specified in a [JSON](https://en.wikipedia.org/wiki/JSON#Syntax) formatted text file in the configuration directory. The configuration file consists of one object with the following possible attributes:
+
+#### metadata
+This attribute, if specified, can contain values to describe the file as determined by the user. It may be useful, for instance, to store version or authorship information. Example:
+
+```
+  "metadata": {
+    "version": "1.0",
+    "modified": "1970-01-01"
+  }
+```
+
+#### properties
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
+
+##### guid
+The globally unique identifier of this property.
+
+##### programmaticName
+A textual description of this property.
+
+##### uiaType
+The data type of this property's value, one of `"string"`, `"int"`, or `"bool"`.
+
+##### interpretation
+If specified, an object mapping values of a property to more friendly descriptions. This is useful, for instance, when the value of a property is an enumeration.
+
+##### Example
+An example of a `properties` object:
+``` json
+  "properties": [
+    {
+      "guid": "{F065BAA7-2794-48B6-A927-193DA1540B84}",
+      "programmaticName": "ViewType",
+      "uiaType": "int",
+      "interpretation": {
+        "1": "ViewSlide",
+        "2": "ViewSlideMaster",
+        "3": "ViewNotesPage",
+        "4": "ViewHandoutMaster",
+        "5": "ViewNotesMaster",
+        "6": "ViewOutline",
+        "7": "ViewSlideSorter",
+        "8": "ViewTitleMaster",
+        "9": "ViewNormal",
+        "10": "ViewPrintPreview",
+        "11": "ViewThumbnails",
+        "12": "ViewMasterThumbnails",
+        "13": "ViewNotesText",
+        "14": "ViewOutlineMaster",
+        "15": "ViewSlideShow",
+        "16": "ViewSlideShowFullScreen",
+        "17": "ViewSlideShowBrowse",
+        "18": "ViewPresenterSlide",
+        "19": "ViewPresenterNotes",
+        "20": "ViewPresenterNextStep",
+        "21": "ViewPresenterTitle",
+        "22": "ViewGridSections",
+        "23": "ViewGridThumbnails",
+        "24": "ViewGridSectionTitle",
+        "25": "ViewGridThumbnailZoom",
+        "26": "ViewGridBack",
+        "27": "ViewProtected",
+        "28": "ViewVisualBasic",
+        "29": "ViewNone",
+        "30": "ViewUnknown"
+      }
+    }
+  ]
+```
+
+#### Example of a complete custom UI Automation configuration file
+``` json
+{
+  "properties": [
+    {
+      "guid": "{4BB56516-F354-44CF-A5AA-96B52E968CFD}",
+      "programmaticName": "AreGridlinesVisible",
+      "uiaType": "bool"
+    },
+    {
+      "guid": "{E244641A-2785-41E9-A4A7-5BE5FE531507}",
+      "programmaticName": "CellFormula",
+      "uiaType": "string"
+    },
+    {
+      "guid": "{626CF4A0-A5AE-448B-A157-5EA4D1D057D7}",
+      "programmaticName": "CellNumberFormat",
+      "uiaType": "string"
+    },
+    {
+      "guid": "{312F7536-259A-47C7-B192-AA16352522C4}",
+      "programmaticName": "CommentReplyCount",
+      "uiaType": "int"
+    },
+    {
+      "guid": "{7AAEE221-E14D-4DA4-83FE-842AAF06A9B7}",
+      "programmaticName": "DataValidationPrompt",
+      "uiaType": "string"
+    },
+    {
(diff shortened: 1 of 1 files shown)


docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT Licence. -->
+
+## Custom UI Automation properties
+Axe Windows supports the inspection of [custom UI Automation properties](https://docs.microsoft.com/en-gb/windows/win32/winauto/uiauto-custompropertieseventscontrolpatterns). The custom properties for which Axe Windows should register are described in a configuration file that is loaded once per session and applies to all inspected UI Automation elements.
+
+### Configuration file format
+Custom UI Automation configuration data is specified in a [JSON](https://en.wikipedia.org/wiki/JSON#Syntax) formatted text file in the configuration directory. The configuration file consists of one object with the following possible attributes:
```
codeofdusk, line 8: Specifics on the config file location intentionally omitted. Exactly where should it be stored and what should it be called?

docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
...
+This attribute, if specified, can contain values to describe the file as determined by the user. It may be useful, for instance, to store version or authorship information. Example:
+
+```
+  "metadata": {
+    "version": "1.0",
+    "modified": "1970-01-01"
+  }
+```
+
+#### properties
```
codeofdusk, line 20: Config for "annotations" and "patterns" could be defined as siblings of this block, but intentionally left out of this version.

docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
...
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
+
```
codeofdusk, line 22: What's the best way to format these properties? For now I left them as headings with paragraphs of text. A definition list perhaps? (but I couldn't get it to render in GitHub-flavoured markdown).

docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT Licence. -->
+
+## Custom UI Automation properties
+Axe Windows supports the inspection of [custom UI Automation properties](https://docs.microsoft.com/en-gb/windows/win32/winauto/uiauto-custompropertieseventscontrolpatterns). The custom properties for which Axe Windows should register are described in a configuration file that is loaded once per session and applies to all inspected UI Automation elements.
```
DaveTryon, line 5: Suggestion: Omit the "once per session" part. That's an implementation detail of the initial version in AIWin, but from an Axe.Windows perspective, there's no reason that we shouldn't be able to add or remove custom properties at anytime during the session lifetim.

docs/CustomUIA.md
DaveTryon, line 8: Suggestion: Omit the file location. This is another artifact of the initial planned implementation in AIWin, but there's no need for Axe.Windows to have this constraint. In fact, if/when we enable this in the CLI, we'll probably want to allow the user to specify the exact path to the config file to use.

docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
...
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
+
+##### guid
+The globally unique identifier of this property.
+
+##### programmaticName
+A textual description of this property.
+
+##### uiaType
+The data type of this property's value, one of `"string"`, `"int"`, or `"bool"`.
```
DaveTryon, line 30: Excel and PowerPoint currently only use string, int, or bool, but the [docs](https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-designingcustompropseventpatterns) also allow support of double, element, or point types. We should support the full set of supported types

codeofdusk: How might we specify "element" or "point" as a string/in the serialized representation?

docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
...
+
+##### interpretation
+If specified, an object mapping values of a property to more friendly descriptions. This is useful, for instance, when the value of a property is an enumeration.
```
DaveTryon, line 33: Questions:
1. Do we disallow this for non-int types?
2. Should we call out that the property value is the key?
3. Should we call out the behavior if no matching key exists? For example, in the sample, if the input is the number 40, is the output the number 40 or the string "40" ?. It could simplify consumption to have a single type of returned object

docs/CustomUIA.md
```diff
@@ -0,0 +1,125 @@
...
+        "28": "ViewVisualBasic",
+        "29": "ViewNone",
+        "30": "ViewUnknown"
+      }
+    }
+  ]
+```
+
+#### Example of a complete custom UI Automation configuration file
+``` json
```
DaveTryon, line 80: Would it make sense to point to https://docs.microsoft.com/en-us/office/uia/excel/excelcustomproperties as the source of these values?

docs/CustomUIA.md
DaveTryon: They'd both need to be converted to a string, similar to how we handle a BoundingRectangle. For a point, we'd just report x and y. For an element, we'd probably report the element's `Glimpse` property, which is essentially a concatenation of the element type and the element name. It's admittedly a bit crude, but it should be adequate to validate that the custom property is working as expected

docs/CustomUIA.md
DaveTryon: Maybe use the style in https://github.com/microsoft/axe-windows/blob/main/docs/AutomationReference.md ?

docs/CustomUIA.md
codeofdusk: I've added a bit more context here, but not sure if it's well-explained.

DaveTryon (approved):

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+#### properties
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
+
+Attribute | Description
+--- | ---
+guid|The globally unique identifier of this property.
+programmaticName|A textual description of this property.
+uiaType|The data type of this property's value, one of `string`, `int`, `bool`, `point`, `element`, or `enum`.
```
dbjorge, line 27: Is `enum` accurate? The [UI Automation docs for custom properties](https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-designingcustompropseventpatterns#designing-custom-properties) suggests that there should be a `double` type and should not be an `enum` type.

dbjorge: Ah, sorry, hadn't read far enough down to see that it's using an int representation. But is it intentional for `double` to be omitted?

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
...
+    "modified": "1970-01-01"
+  }
+```
+
+#### properties
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
+
+Attribute | Description
+--- | ---
+guid|The globally unique identifier of this property.
```
dbjorge, line 25: It would be good for this to clarify the exact expected format of the GUID. If you don't have a reason to prefer any particular specific format, I'd reference [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122#section-4.1) (this matches what [the latest draft of the JSON Schema spec](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-02#section-7.3.5) does). If you do this, you'll want to drop the `{` and `}` characters in the example's GUIDs

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
...
+programmaticName|A textual description of this property.
```
dbjorge, line 26: It would be good for this to clarify any restrictions/format expectations, maybe explaining how the name will be used. All the examples use PascalCasedNamesWithoutSpaces - is this a requirement/recommendation?

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT Licence. -->
+
+## Custom UI Automation properties
+Axe Windows supports the inspection of [custom UI Automation properties](https://docs.microsoft.com/en-gb/windows/win32/winauto/uiauto-custompropertieseventscontrolpatterns). The custom properties for which Axe Windows should register are described in a user-supplied text file that is applied to all inspected UI Automation elements.
+
+### Configuration file format
+Custom UI Automation configuration data is specified in a [JSON](https://en.wikipedia.org/wiki/JSON#Syntax) formatted text file. The file consists of one object with the following possible attributes:
```
dbjorge, line 8: For JSON files intended for direct human consumption, I'd recommend including an allowance for comments in the file as part of our spec. `System.Text.Json` and `Newtonsoft.Json` both support reading files with comments so this shouldn't cause any extra implementation burden.

dbjorge: (GitHub and VS Code both understand "JSON with Comments" as a separate language using extension "jsonc", if you want to include comments in any of the examples as a demonstration)

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### properties
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
+
+Attribute | Description
+--- | ---
+guid|The globally unique identifier of this property.
+programmaticName|A textual description of this property.
+uiaType|The data type of this property's value, one of `string`, `int`, `bool`, `point`, `element`, or `enum`.
+values|If this property is an enumeration (`"uiaType": "enum"`), this attribute contains a mapping of integer members of the enumeration to human-readable descriptions. If the value of the property matches one of the attributes of this object, `"value from object (<raw value>)"` is presented as the value for this custom property. If not present, `"Unknown (<raw value>)"` is presented instead.
```
dbjorge, line 28: I worry that the phrase "a mapping of integer members of the enumeration to human-readable descriptions" might confuse folks about whether they should be writing the keys as `7` or `"7"`. The JSON spec demands the latter (which matches your examples), but I don't think all users will necessarily be aware of that; it might preempt some user mistakes to clarify here in the description that the keys should be string representations of integer members.

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT Licence. -->
+
+## Custom UI Automation properties
+Axe Windows supports the inspection of [custom UI Automation properties](https://docs.microsoft.com/en-gb/windows/win32/winauto/uiauto-custompropertieseventscontrolpatterns). The custom properties for which Axe Windows should register are described in a user-supplied text file that is applied to all inspected UI Automation elements.
+
+### Configuration file format
+Custom UI Automation configuration data is specified in a [JSON](https://en.wikipedia.org/wiki/JSON#Syntax) formatted text file. The file consists of one object with the following possible attributes:
+
+#### metadata
```
dbjorge, line 10: Do we have any plan right now for having axe-windows use this information directly? If not, it might be better to suggest users represent this as a comment instead, to avoid user confusion if we ever need to add a more strongly specified version field/etc.

docs/CustomUIA.md
```diff
@@ -0,0 +1,115 @@
...
+
+```
+  "metadata": {
+    "version": "1.0",
+    "modified": "1970-01-01"
+  }
+```
+
+#### properties
+This attribute contains an array of objects, one per defined custom UI Automation property. Each property object contains the following fields:
```
dbjorge, line 21: If these are intended to match exactly with the properties as-registered with UI Automation, it might be good to note that/link those docs, maybe something like:

```markdown
The values specified here should match the values in [the UIAutomationPropertyInfo structure](https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcore/ns-uiautomationcore-uiautomationpropertyinfo) used to [register the custom property](https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-regcustompropseventpatterns#registering-custom-properties-and-events).
```

dbjorge (changes requested):
Looks great overall! Suggested a few updates inline

docs/CustomUIA.md
DaveTryon: Thanks, @dbjorge! I didn't know about NewtonSoft supporting comments. Based on my experiments, it looks like we can use either `//` for single-line comments or `/*...*/` for multi-line comments

docs/CustomUIA.md
codeofdusk: Not intentional!

docs/CustomUIA.md
codeofdusk: These are the names used by the Office docs. I don't know of any restrictions other than how the names are implemented on the provider side?

docs/CustomUIA.md
codeofdusk: If we don't need to be strictly json compliant, how about [JSON5](https://json5.org/)?

docs/CustomUIA.md
codeofdusk: I added this as strict JSON has no support for comments.

docs/CustomUIA.md
codeofdusk: This is the format used both by NVDA and in the UIA docs for Excel and PowerPoint.

DaveTryon: As discussed, Newtonsoft can read guids with or without braces, but it writes guids without braces. Let's omit the braces in the spec.

docs/CustomUIA.md
DaveTryon: As discussed, let's remove this block in favor of comments

docs/CustomUIA.md
DaveTryon: json5 may not be completely supported by Newtonsoft or GitHub, so let's not commit to json5 at this time

docs/CustomUIA.md
dbjorge: It seems reasonable to me that the only restriction for us would be "match the provider's name"

docs/CustomUIA.md
dbjorge: This is an excellent question. I think my recollection from feature estimation was that we intended for Accessibility Insights for Windows to use a well-known file name/location, but for axe-windows to add "accepting that file" as part of API calls; if that's correct, I think it would be reasonable for axe-windows' doc here to omit the name/location and for that to be documented in accessibility-insights-windows.

I'm fine with deferring the specifics of filename/location/API changes for a later PR, though; it's okay for this one to just cover the format.

dbjorge (approved):
LGTM!

docs/CustomUIA.md
DaveTryon: The specific location will be part of changes in AIWin. The Axe.Windows CLI will eventually allow the user to specify a json file via the command line.

PR #588, opened 2021-06-03
feat(custom-uia): Add Interop.UIAutomationCore assembly

DaveTryon:
#### Details

The definitions for the classes that are needed for the custom UIA feature are not in the interop assembly that we already have. The assembly added here is taken from https://www.nuget.org/packages/TestStack.UiaCustomPattersManaged/, which comes from a project that has recently been licensed under MIT license (see https://github.com/TestStack/uia-custom-pattern-managed), so we can use it without running into licensing problems. That said, we don't want to just consume this package directly since it has a very long tail of transitive dependencies that come from other assemblies within the package.

##### Motivation

Unblock the custom uia feature

##### Context

I'm not a huge fan of having binaries in the repo. It might make sense to replace the checked-in assemblies with checked-in IL files (generated from **ildasm.exe**), then have a build step where we compile these files back into assemblies using **ilasm.exe**. This could be added fairly simply at a later time if we choose to take this approach and get binaries out of the repo.

We also considered writing our own interop layer, but there are enough layers in the API that I really wanted to build upon other people's work, instead of rolling our own.

#### Pull request checklist


dbjorge:
Bundling a 3rd party component like this requires:

* that we ensure it is included in NOTICE content (which must be distributed with both axe-windows and Accessibility Insights for Windows)
* that we include a copy of its LICENSE file alongside the binary

DaveTryon:
Closing this in favor of a new plan: The author of the interop assembly is going to publish the interop as a standalone, then we'll consume it that way.

PR #589, opened 2021-06-07
feat(custom-uia): Load and validate custom properties from disk

codeofdusk:
#### Details
This PR adds classes and unit tests for reading and validating custom UIA properties defined in JSON according to the specification defined in #587.

##### Motivation

##### Context
Later PRs to handle the "enum" type, custom property registration, and data rendering.

This PR assumes that a non-empty "properties" collection will *always* be defined. This might not be the case if custom annotations/patterns are later added, so this constraint may need to be relaxed later (and `EmptyConfigTest` possibly updated to reflect this).

#### Pull request checklist

### DIFF
--- a/src/Desktop/UIAutomation/CustomObjects/Config.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Config.cs
@@ -25,7 +25,8 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
         private void Validate()
         {
             if (this.Properties == null || this.Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
-            foreach (CustomProperty p in this.Properties) p.Validate();
+            foreach (CustomProperty p in this.Properties) 
+                p.Validate();
         }
     }
 }
--- a/src/Desktop/UIAutomation/CustomObjects/Config.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Config.cs
@@ -12,7 +12,7 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
         [JsonProperty("properties")]
         public CustomProperty[] Properties { get; set; }
 
-        public static Config FromText(string text)
+        public static Config ReadFromText(string text)
         {
             Config conf = JsonConvert.DeserializeObject<Config>(text);
             conf.Validate();
@@ -20,12 +20,12 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
             return conf;
         }
 
-        public static Config FromFile(string path) { return Config.FromText(File.ReadAllText(path)); }
+        public static Config ReadFromFile(string path) { return Config.ReadFromText(File.ReadAllText(path)); }
 
         private void Validate()
         {
-            if (this.Properties == null || this.Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
-            foreach (CustomProperty p in this.Properties) 
+            if (Properties == null || Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
+            foreach (CustomProperty p in Properties) 
                 p.Validate();
         }
     }
--- a/src/Desktop/UIAutomation/CustomObjects/Config.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Config.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+using System;
+using System.IO;
+
+namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
+{
+    class Config
+    {
+        [JsonProperty("properties")]
+        public CustomProperty[] Properties { get; set; }
+
+        public static Config FromText(string text)
+        {
+            Config conf = JsonConvert.DeserializeObject<Config>(text);
+            conf.Validate();
+            // We made it here, so config must be valid.
+            return conf;
+        }
+
+        public static Config FromFile(string path) { return Config.FromText(File.ReadAllText(path)); }
+
+        private void Validate()
+        {
+            if (this.Properties == null || this.Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
+            foreach (CustomProperty p in this.Properties) p.Validate();
+        }
+    }
+}
--- a/src/Desktop/UIAutomation/CustomObjects/Config.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Config.cs
@@ -25,7 +25,7 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
 
         private void Validate()
         {
-            if (Properties == null || !Properties.Any()) throw new ArgumentException("Empty or missing definition of custom properties.");
+            if (Properties == null || !Properties.Any()) throw new InvalidDataException("Empty or missing definition of custom properties.");
             foreach (CustomProperty p in Properties) 
                 p.Validate();
         }
--- a/src/Desktop/UIAutomation/CustomObjects/Config.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Config.cs
@@ -4,6 +4,7 @@
 using Newtonsoft.Json;
 using System;
 using System.IO;
+using System.Linq;
 
 namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
 {
@@ -24,7 +25,7 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
 
         private void Validate()
         {
-            if (Properties == null || Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
+            if (Properties == null || !Properties.Any()) throw new ArgumentException("Empty or missing definition of custom properties.");
             foreach (CustomProperty p in Properties) 
                 p.Validate();
         }
--- a/src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
@@ -10,9 +10,9 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
     {
         /// <summary>The RFC4122 globally unique identifier of this property.</summary>
         [JsonProperty("guid")]
-        #pragma warning disable CA1720 // Identifier contains type name: name from JSON
+#pragma warning disable CA1720 // Identifier contains type name: name from JSON
         public Guid Guid { get; set; }
-        #pragma warning restore CA1720 // Identifier contains type name: name from JSON
+#pragma warning restore CA1720 // Identifier contains type name: name from JSON
         /// <summary>A textual description of this property.</summary>
         [JsonProperty("programmaticName")]
         public string ProgrammaticName { get; set; }
@@ -20,7 +20,7 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
         /// <summary>The data type of this property's value as specified by the user, one of string, int, bool, double, point, or element.</summary>
         // TODO Bill: add enum (with values member)
         [JsonProperty("uiaType")]
-        public string UserType { get; set; }
+        public string DataType { get; set; }
 
         /// <summary>The dynamic ID assigned to this property by the system.</summary>
         [JsonIgnore]
@@ -30,8 +30,13 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
         {
             if (Guid == Guid.Empty) throw new ArgumentException("Missing GUID in custom property definition.");
             if (ProgrammaticName == null) throw new ArgumentException("Missing programmatic name in custom property definition.");
-            if (UserType == null) throw new ArgumentException("Missing type in custom property definition.");
-            switch (UserType)
+            if (DataType == null) throw new ArgumentException("Missing type in custom property definition.");
+            ValidateType();
+        }
+
+        internal void ValidateType()
+        {
+            switch (DataType)
             {
                 case "string":
                     // Valid type, further processing in a subsequent PR.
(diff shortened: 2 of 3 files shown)


src/Desktop/UIAutomation/CustomObjects/BoolTypeConverter.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
DaveTryon, line 1: How would you feel about moving all of the converters inside a Converters folder, and putting them into a Converters namespace? That would group them nicely together. As is, it seems a bit strange to have all the code intermixed

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        public static Config FromFile(string path) { return Config.FromText(File.ReadAllText(path)); }
+
+        private void Validate()
+        {
+            if (this.Properties is null || this.Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
+            foreach (CustomProperty p in this.Properties)
+            {
+                if (p.Guid is null) throw new ArgumentException("Missing GUID in custom property definition.");
```
DaveTryon, line 30: I thought we'd decided to use `==` instead of `is`, since the error messages seem to be more helpful?

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,59 @@
...
+                if (p.ProgrammaticName is null) throw new ArgumentException("Missing programmatic name in custom property definition.");
+                if (p.UserType is null) throw new ArgumentException("Missing type in custom property definition.");
+                switch (p.UserType)
```
DaveTryon, line 33: It feels odd to have the config update the CustomProperty. What if we moved the code to create the converters to CustomProperty.Validate? That brings the code together and allows the setter to be private.

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    case "string":
```
DaveTryon, line 35: Tests only cover the `bool` case, so coverage will have gaps. To keep the PR small(ish) now, maybe implement just the bool type and class, then have the other classes (and their tests) in a separate PR?

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.IO;
+
+namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
+{
+    class Config
+    {
+        [JsonProperty("properties")]
+        public CustomProperty[] Properties { get; set; }
+
+        public static Config FromText(string text)
```
RobGallo, line 15: Methods should begin with a verb. In this case "Create" would be appropriate because an object is being created and returned.

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,31 @@
...
+            // We made it here, so config must be valid.
+            return conf;
+        }
+
+        public static Config FromFile(string path) { return Config.FromText(File.ReadAllText(path)); }
+
+        private void Validate()
+        {
+            if (this.Properties == null || this.Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
+            foreach (CustomProperty p in this.Properties) p.Validate();
```
RobGallo, line 28: It would be preferable, to me anyway, if the for statement and the executed code were on separate lines. That makes it easier for me to separate the loop from the action taken within the loop. The same applies to most if conditions. The one place I like to have the condition and the execution on the same line is when doing parameter or returned data validation. Those statements are just boiler plate and having them on the same line means I know I can ignore them. New lines are a great tool for finding what is important and ignoring what isn't, or ignoring what can safely be assumed.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+
+namespace Axe.Windows.Desktop.UIAutomation.CustomObjects
+{
+    public class CustomProperty
+    {
+        /// <summary>The RFC4122 globally unique identifier of this property.</summary>
+        [JsonProperty("guid")]
+        #pragma warning disable CA1720 // Identifier contains type name: name from JSON
+        public Guid? Guid { get; set; }
```
RobGallo, line 14: Why is this property optional? Later we check to ensure it isn't null.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
```diff
@@ -0,0 +1,59 @@
...
+        #pragma warning restore CA1720 // Identifier contains type name: name from JSON
+        /// <summary>A textual description of this property.</summary>
+        [JsonProperty("programmaticName")]
+        public string ProgrammaticName { get; set; }
```
RobGallo, line 18: Let's don't overthink this: this property is just a "Name". "Programmatic" doesn't add distinguishing information.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>The data type of this property's value as specified by the user, one of string, int, bool, double, point, or element.</summary>
+        // TODO Bill: add enum (with values member)
+        [JsonProperty("uiaType")]
+        public string UserType { get; set; }
```
RobGallo, line 23: Again, let's don't overthink this. The name "Type" is sufficient to indicate the usage of this property within the context. And the property name within the class should match the json property name.

DaveTryon: The analyzer doesn't like `Type` as the name here

src/Desktop/UIAutomation/CustomObjects/Config.cs
DaveTryon: This pattern also appears in other places, like [A11yElement](https://github.com/microsoft/axe-windows/blob/main/src/Core/Bases/A11yElement.cs#L685). It might be worth revising them all if we want to have a consistent pattern

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
DaveTryon: Guid is a value type, so it can't be checked for null unless it's nullable (the `?` operator). We could use a non-nullable Guid and check for Guid.Empty in the validation step if that's preferable.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
codeofdusk: Also, we'll want to distinguish between the type given by the user and the underlying UIA representation, but the latter isn't yet present in this PR.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
codeofdusk: How much do we care about [being consistent with UIA](https://docs.microsoft.com/en-us/windows/win32/api/uiautomationcore/ns-uiautomationcore-uiautomationpropertyinfo) here? (CTRL+f for "programmatic").

src/Desktop/UIAutomation/CustomObjects/Config.cs
codeofdusk: I agree, but this breaks precedent from names in other classes. Do we care?

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>The dynamic ID assigned to this property by the system.</summary>
+        [JsonIgnore]
+        public int DynamicId { get; set; }
+
+        internal void Validate()
+        {
+            if (this.Guid == null) throw new ArgumentException("Missing GUID in custom property definition.");
+            if (this.ProgrammaticName == null) throw new ArgumentException("Missing programmatic name in custom property definition.");
+            if (this.UserType == null) throw new ArgumentException("Missing type in custom property definition.");
+            switch (this.UserType)
```
RobGallo, line 34: How about breaking this switch out into a separate function that either validates the type or returns the Type (note the capital T in "Type").

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,32 @@
...
+            conf.Validate();
+            // We made it here, so config must be valid.
+            return conf;
+        }
+
+        public static Config FromFile(string path) { return Config.FromText(File.ReadAllText(path)); }
+
+        private void Validate()
+        {
+            if (this.Properties == null || this.Properties.Length < 1) throw new ArgumentException("Empty or missing definition of custom properties.");
```
DaveTryon, line 27: Let's drop the superfluous `this.` prefixes. Some of the older code in Axe.Windows uses them, but newer code generally doesn't--it just hasn't been a priority to remove them from the older code.

DaveTryon, line 27: `!Properties.Any()` instead of `Properties.Length < 1`

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
RobGallo: How is the type given by the user going to be different than the type from UIA?

src/Desktop/UIAutomation/CustomObjects/Config.cs
RobGallo: I personally prefer that we use best practices in newly added code over maintaining consistency. Others may have different points of view.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
codeofdusk: The user type is a string (from JSON). The type needed for registration is a `UIAutomationType`.

codeofdusk: How about `JsonType`?

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
RobGallo: I like Guid.empty. It seems more straight forward to me.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
codeofdusk: Discussed offline with @DaveTryon, changing to `DataType`.

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
```diff
@@ -0,0 +1,64 @@
...
+                    // Valid type, further processing in a subsequent PR.
+                    break;
+                case "point":
+                    // Valid type, further processing in a subsequent PR.
+                    break;
+                case "element":
+                    // Valid type, further processing in a subsequent PR.
+                    break;
+                default:
+                    throw new ArgumentException($"Unknown type {this.DataType}.");
```
DaveTryon, line 60: Remove `this.`
```suggestion
                    throw new ArgumentException($"Unknown type {DataType}.");
```

src/DesktopTests/UIAutomation/CustomObjects/ConfigTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            CustomProperty prop = conf.Properties[0];
+            Assert.AreEqual(new Guid("4BB56516-F354-44CF-A5AA-96B52E968CFD"), prop.Guid);
+            Assert.AreEqual("AreGridlinesVisible", prop.ProgrammaticName);
+            Assert.AreEqual("bool", prop.DataType);
+        }
+
+        [TestMethod, Timeout(1000)]
+        public void EmptyConfigTest()
+        {
+            Assert.ThrowsException<ArgumentException>(() => Config.ReadFromText("{}"));
```
DaveTryon, line 28: Should we have tests for non-JSON strings, like null, string.Empty, etc?

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,33 @@
...
+            return conf;
+        }
+
+        public static Config ReadFromFile(string path) { return Config.ReadFromText(File.ReadAllText(path)); }
+
+        private void Validate()
+        {
+            if (Properties == null || !Properties.Any()) throw new ArgumentException("Empty or missing definition of custom properties.");
+            foreach (CustomProperty p in Properties) 
+                p.Validate();
```
DaveTryon, line 30: In a case like this, I'd actually put p.Validate inside braces, purely to avoid jumping 2 levels of indent between lines 30 and 31. Not required, though

src/Desktop/UIAutomation/CustomObjects/Config.cs
```diff
@@ -0,0 +1,33 @@
...
+            conf.Validate();
+            // We made it here, so config must be valid.
+            return conf;
+        }
+
+        public static Config ReadFromFile(string path) { return Config.ReadFromText(File.ReadAllText(path)); }
+
+        private void Validate()
+        {
+            if (Properties == null || !Properties.Any()) throw new ArgumentException("Empty or missing definition of custom properties.");
```
DaveTryon, line 28: As designed, ArgumentException is intended for use against a parameter, and the parameter name should be in the exception. In this case, since we don't have a parameter, let's throw [InvalidDataException](https://docs.microsoft.com/en-us/dotnet/api/system.io.invaliddataexception?view=net-5.0).

Same logic applies to CustomProperty.Validate() and CustomProperty.ValidateType(), neither of which has a parameter

DaveTryon (approved):

PR #593, opened 2021-06-10
feat(custom-uia): add "type converters" for rendering and registration

codeofdusk:
#### Details
Repeatedly checking the JSON type string (`CustomProperty.DataType`) at registration and runtime is expensive, and the `enum` type requires special handling. This PR therefore adds a new `iTypeConverter` interface and `TypeConverter` classes which encapsulate type information for custom properties and streamline rendering.

##### Motivation

##### Context
* The `enum` type is out-of-scope for this PR.
* Commit 1bf2a0c922192f65d4a45611042beec7e6f9a468 (removal of `UnderlyingUIAType` property used to aid registration) to be reverted after merge of #592 (in a subsequent PR).

#### Pull request checklist

### DIFF
--- a/src/Desktop/UIAutomation/CustomObjects/Converters/DoubleTypeConverter.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Converters/DoubleTypeConverter.cs
@@ -12,7 +12,6 @@ namespace Axe.Windows.Desktop.UIAutomation.CustomObjects.Converters
         {
             if (value == null) throw new ArgumentNullException(nameof(value));
             return ((double)value).ToString(CultureInfo.InvariantCulture);
-
         }
     }
 }
--- a/src/Desktop/UIAutomation/CustomObjects/Converters/DoubleTypeConverter.cs
+++ b/src/Desktop/UIAutomation/CustomObjects/Converters/DoubleTypeConverter.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Interop.UIAutomationCore;
+using System;
+using System.Globalization;
+
+namespace Axe.Windows.Desktop.UIAutomation.CustomObjects.Converters
+{
+    class DoubleTypeConverter : ITypeConverter
+    {
+        public UIAutomationType UnderlyingUiaType => UIAutomationType.UIAutomationType_Double;
+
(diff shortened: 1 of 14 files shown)


src/Desktop/UIAutomation/CustomObjects/Converters/DoubleTypeConverter.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Axe.Windows.Desktop.UIAutomation.CustomObjects.Converters
+{
+    class DoubleTypeConverter : ITypeConverter
+    {
+        public string Render(dynamic value)
+        {
+            if (value == null) throw new ArgumentNullException(nameof(value));
+            return ((double)value).ToString(CultureInfo.InvariantCulture);
+
```
DaveTryon, line 15: Remove the empty line on line 15?

src/Desktop/UIAutomation/CustomObjects/CustomProperty.cs
```diff
@@ -23,6 +24,10 @@ public class CustomProperty
         [JsonProperty("uiaType")]
         public string DataType { get; set; }
 
+        /// <summary>A type converter for this property, representing its underlying UIA type and providing string rendering.</summary>
```
DaveTryon, line 27: (nit): We're not yet representing the underlying UIA type

src/Desktop/UIAutomation/CustomObjects/Converters/PointTypeConverter.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Axe.Windows.Desktop.UIAutomation.CustomObjects.Converters
+{
+    class PointTypeConverter : ITypeConverter
+    {
+        public string Render(dynamic value)
+        {
+            if (value == null) throw new ArgumentNullException(nameof(value));
+            double[] arr = (double[])value;
+            return $"[x={arr[0]},y={arr[1]}]";
```
DaveTryon, line 14: Strictly speaking, this is returning an [interpolated string](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated), which is subject to the user's culture setting. This might lead to an issue in TSA (a build-time analysis database). It's been so long since we had one of these warnings (we fixed them all) that I'm not 100% sure that this still shows up as a warning. Let's try it for now, then we can add the trivial change to fix it later if we get a TSA warning.

Issue #590, opened 2021-06-08
[BUG] Allow RadioButton to implement TogglePattern


karanbirsingh:
## Describe the bug

The rule https://github.com/microsoft/axe-windows/blob/a23dda60ec3784ab5b53a7d6f3484a1836a815fa/src/Rules/Library/ControlShouldNotSupportTogglePattern.cs#L13 prohibits radio buttons from implementing the toggle pattern. This is in line with the UIA documentation [here](https://docs.microsoft.com/en-us/dotnet/framework/ui-automation/ui-automation-support-for-the-radiobutton-control-type#required-ui-automation-control-patterns). After some offline conversation with Rob+others, we feel it's more accurate to say that while it's rarely a good idea to implement the TogglePattern, doing so doesn't seem to impede the accessibility of the control. This bug is to relax or remove the rule so that radio buttons that do implement the toggle pattern aren't flagged by Axe.Windows. We are also reaching out to the docs owners to suggest accompanying updates there.

chingucoding:
Can I take a look at this issue?

DaveTryon:
@chingucoding, I'm not sure we know exactly what we want to do here. My guess is that we'll downgrade the rule from `ErrorCode.Error` to `ErrorCode.Warning`, although that's not 100% certain. Please feel free to create a PR against that assumption and I'll assign the issue for triage (we normally triage these issues on Mondays, so the PR might sit idle for a few days before we know for sure what our strategy will be on this).

chingucoding:
Ah I see, my "instinct" would have been to just ignore the RadioButton control from this rule and leave it as an error. Alternatively, what about forking this rule into two rules:
1. Controls that DEFINITELY should not support togglepattern -> Error
2. Radiobutton that should not support the toggle pattern but is also okay if it supports toggle pattern -> Warning

DaveTryon:
@chingucoding, that was my first thought, as well. Looking at this rule's [Condition](https://github.com/microsoft/axe-windows/blob/main/src/Rules/Library/ControlShouldNotSupportTogglePattern.cs#L32), though, only applies to `RadioButton` elements, so the decision tree collapses to just changing the [ErrorCode](https://github.com/microsoft/axe-windows/blob/main/src/Rules/Library/ControlShouldNotSupportTogglePattern.cs#L20) to a warning (and updating appropriate tests).

We'll have to wait and see what comes out of the triage discussion.

chingucoding:
[...]

PR #594, opened 2021-06-11
fix: Remove RadioButton should not support TogglePattern rule

chingucoding:
#### Details

The "RadioButton should not support TogglePattern rule" has now been removed.

##### Motivation

Addresses issue #590 

#### Pull request checklist

- [x] Addresses an existing issue: #590


DaveTryon:
@chingucoding, we discussed this today and decided that we want to remove this rule instead of making it a warning. I've flagged this PR as "do not merge" in its current form. Please feel free to repurpose this PR or create a new one if you still want to address this issue.

BTW, the PR looked really good had we gone with the "make it a warning" approach.

chingucoding:
@DaveTryon PR should be up to standard now, rule (and obsolete tests) are now removed.

DaveTryon:
You probably also need to delete https://github.com/microsoft/axe-windows/blob/main/src/Core/Enums/RuleId.cs#L113

DaveTryon (approved):

PR #595, opened 2021-06-11
fix: use StrongName-signed interop

DaveTryon:
#### Details

#592 broke the signed build because the assembly from the [Interop.UIAutomationCore](https://www.nuget.org/packages/Interop.UIAutomationCore/) package is not StrongName signed. This PR simply changes to the [Interop.UIAutomationCore.Signed](https://www.nuget.org/packages/Interop.UIAutomationCore.Signed/) package, which is a StrongName signed version of the same assembly. The assembly itself is still named Interop.UIAutomationCore, so the only required change is to update the package reference. I queued a [validation build](https://mseng.visualstudio.com/1ES/_build/results?buildId=15448326&view=results), which completed successfully.

##### Motivation

Broken builds are very bad

#### Pull request checklist

### DIFF
--- a/src/Desktop/Desktop.csproj
+++ b/src/Desktop/Desktop.csproj
@@ -9,7 +9,7 @@
   <Import Project="..\..\build\NetStandardRelease.targets" />
 
   <ItemGroup>
-    <PackageReference Include="Interop.UIAutomationCore" Version="10.19041.0" />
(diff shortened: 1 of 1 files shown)


codeofdusk (approved):
Broken builds are very bad indeed!

PR #597, opened 2021-06-13
feat(custom-uia): remove Validate() methods and perform field validation on property setters

codeofdusk:
#### Details
This PR removes the `Config.Validate()` and `CustomProprty.Validate()` methods, instead validating at the attribute level. It also relaxes the requirement for the JSON file to have custom properties, or any contents at all, defined as this may be useful for follow-up work on annotations/patterns. `DataType` has been renamed `ConfigType` (as it is the type from the user configuration) to distinguish it with a newly-added `Type` attribute containing an enum value kept in sync with the `ConfigType` property via setters. The new `Type` attribute should be used in calling code for faster performance, but `ConfigType` must be public for the benefit of Newtonsoft. Since Newtonsoft catches all exceptions thrown by calling code and re-throws its own, tests expecting our `InvalidDataException` were modified accordingly.

##### Motivation
Part of custom UIA extensions (internal access required to view).

##### Context
Further work to follow in subsequent PRs.

#### Pull request checklist

### DIFF
--- a/src/Core/CustomObjects/CustomProperty.cs
+++ b/src/Core/CustomObjects/CustomProperty.cs
@@ -10,54 +10,53 @@ namespace Axe.Windows.Core.CustomObjects
 {
     public class CustomProperty
     {
+#pragma warning disable CA1720 // Identifier contains type name: name from JSON
+        private Guid _Guid;
         /// <summary>The RFC4122 globally unique identifier of this property.</summary>
         [JsonProperty("guid")]
-#pragma warning disable CA1720 // Identifier contains type name: name from JSON
-        public Guid Guid { get; set; }
+        public Guid Guid {
+            get { return _Guid; }
+            set
+            {
+                if (value == Guid.Empty) throw new InvalidDataException("Missing GUID in custom property definition.");
+                _Guid = value;
+            }
+        }
 #pragma warning restore CA1720 // Identifier contains type name: name from JSON
+
+        private string _ProgrammaticName;
         /// <summary>A textual description of this property.</summary>
         [JsonProperty("programmaticName")]
-        public string ProgrammaticName { get; set; }
-
-        /// <summary>The data type of this property's value as specified by the user, one of string, int, bool, double, point, or element.</summary>
-        // TODO Bill: add enum (with values member)
-        [JsonProperty("uiaType")]
-        public string DataType { get; set; }
-
-        /// <summary>A type converter for this property, providing string rendering.</summary>
-        [JsonIgnore]
-        public ITypeConverter TypeConverter { get; private set; }
-
-        /// <summary>The dynamic ID assigned to this property by the system.</summary>
-        [JsonIgnore]
-        public int DynamicId { get; set; }
-
-        internal void Validate()
-        {
-            if (Guid == Guid.Empty) throw new InvalidDataException("Missing GUID in custom property definition.");
-            if (ProgrammaticName == null) throw new InvalidDataException("Missing programmatic name in custom property definition.");
-            if (DataType == null) throw new InvalidDataException("Missing type in custom property definition.");
-            TypeConverter = CreateTypeConverter();
+        public string ProgrammaticName {
+            get { return _ProgrammaticName; }
+            set
+            {
+                if (value == null) throw new InvalidDataException("Missing programmatic name in custom property definition.");
+                _ProgrammaticName = value;
+            }
         }
 
-        internal ITypeConverter CreateTypeConverter()
+        private string _DataType;
+        /// <summary>The data type of this property's value as specified by the user, one of string, int, bool, double, point, or element.</summary>
+        [JsonProperty("uiaType")]
+        public string DataType
         {
-            switch (DataType)
+            get { return _DataType; }
+            set
             {
-                case "string":
-                    return new StringTypeConverter();
-                case "int":
-                    return new IntTypeConverter();
-                case "bool":
-                    return new BoolTypeConverter();
-                case "double":
-                    return new DoubleTypeConverter();
-                case "point":
-                    return new PointTypeConverter();
-                case "element":
-                    return new ElementTypeConverter();
-                default:
-                    throw new InvalidDataException($"Unknown type {DataType}.");
+                switch (value)
+                {
+                    case "string":
+                    case "int":
+                    case "bool":
+                    case "double":
+                    case "point":
+                    case "element":
+                        _DataType = value;
+                        break;
+                    default:
+                        throw new InvalidDataException("Type in custom property definition is missing or invalid.");
+                }
             }
         }
     }
--- a/src/Core/CustomObjects/CustomProperty.cs
+++ b/src/Core/CustomObjects/CustomProperty.cs
@@ -52,7 +52,7 @@ namespace Axe.Windows.Core.CustomObjects
                         Type = CustomUIAPropertyType.Element;
                         break;
                     default:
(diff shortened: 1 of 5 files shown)


src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,67 @@ namespace Axe.Windows.Core.CustomObjects
 {
     public class CustomProperty
     {
+#pragma warning disable CA1720 // Identifier contains type name: name from JSON
+        private Guid _Guid;
```
DaveTryon, line 14: We follow the .NET pattern of starting property names with upper case, data members with lower case. As such, we should use `_guid`  (lower case 'g') instead of `_Guid` (upper case 'G')

src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,67 @@ namespace Axe.Windows.Core.CustomObjects
...
         /// <summary>The RFC4122 globally unique identifier of this property.</summary>
         [JsonProperty("guid")]
-#pragma warning disable CA1720 // Identifier contains type name: name from JSON
-        public Guid Guid { get; set; }
+        public Guid Guid {
+            get { return _Guid; }
+            set
+            {
+                if (value == Guid.Empty) throw new InvalidDataException("Missing GUID in custom property definition.");
```
DaveTryon, line 21: If the file doesn't specify the value, we'll have Guid.Empty, which suggests that we need to check this later on either way. With that in mind, is this validation really useful?

src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,67 @@ namespace Axe.Windows.Core.CustomObjects
...
+                _Guid = value;
+            }
+        }
 #pragma warning restore CA1720 // Identifier contains type name: name from JSON
+
+        private string _ProgrammaticName;
```
DaveTryon, line 27: Casing: power case 'P'

src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,67 @@ namespace Axe.Windows.Core.CustomObjects
...
+                if (value == null) throw new InvalidDataException("Missing programmatic name in custom property definition.");
+                _ProgrammaticName = value;
+            }
         }
 
-        internal ITypeConverter CreateTypeConverter()
+        ///  <summary>An internal representation of this property's type. For performance reasons, calling code should always use this property in place of the config representation.s</summary>
+        public CustomUIAPropertyType Type { get; private set; }
+
+        private string _ConfigType;
```
DaveTryon, line 42: Casing: lower case 'C'

src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,67 @@ namespace Axe.Windows.Core.CustomObjects
...
-                    return new BoolTypeConverter();
-                case "double":
-                    return new DoubleTypeConverter();
-                case "point":
-                    return new PointTypeConverter();
-                case "element":
-                    return new ElementTypeConverter();
-                default:
-                    throw new InvalidDataException($"Unknown type {DataType}.");
+                switch (value)
```
DaveTryon, line 50: We should include unit tests to pin the behavior of each of these possible branches

src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,67 @@ namespace Axe.Windows.Core.CustomObjects
...
-        {
-            if (Guid == Guid.Empty) throw new InvalidDataException("Missing GUID in custom property definition.");
-            if (ProgrammaticName == null) throw new InvalidDataException("Missing programmatic name in custom property definition.");
-            if (DataType == null) throw new InvalidDataException("Missing type in custom property definition.");
-            TypeConverter = CreateTypeConverter();
+        public string ProgrammaticName {
+            get { return _ProgrammaticName; }
+            set
+            {
+                if (value == null) throw new InvalidDataException("Missing programmatic name in custom property definition.");
```
DaveTryon, line 34: Similar to the guid, if this gets omitted from the JSON, the default value will be null. Since validation will be needed eventually, this check might be unnecessary

src/CoreTests/CustomObjects/ConfigTests.cs
```diff
@@ -67,28 +63,12 @@ public void BadGuidConfigTest()
...
-            const string MissingNameConfig = "{\"properties\": [{\"guid\": \"4BB56516-F354-44CF-A5AA-96B52E968CFD\", \"uiaType\": \"bool\"}]}";
-            Assert.ThrowsException<InvalidDataException>(() => Config.ReadFromText(MissingNameConfig));
-        }
-
-        [TestMethod, Timeout(1000)]
-        public void MissingTypeConfigTest()
-        {
-            const string MissingTypeConfig = "{\"properties\": [{\"guid\": \"4BB56516-F354-44CF-A5AA-96B52E968CFD\", \"programmaticName\": \"AreGridlinesVisible\"}]}";
-            Assert.ThrowsException<InvalidDataException>(() => Config.ReadFromText(MissingTypeConfig));
+            try
```
DaveTryon, line 66: Does this test still pass? I don't understand where it would throw...

src/CoreTests/CustomObjects/CustomPropertyTests.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        [TestMethod, Timeout(1000)]
+        public void ElementConfigTest()
+        {
+            CustomProperty prop = new CustomProperty();
+            prop.ConfigType = "element";
+            Assert.AreEqual(CustomUIAPropertyType.Element, prop.Type);
+        }
+    }
```
DaveTryon, line 62: Should we have a test that passes in an unsupported value and throws?

src/CoreTests/CustomObjects/CustomPropertyTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
+
+namespace Axe.Windows.CoreTests.CustomObjects
+{
+    [TestClass()]
+    public class CustomPropertyTests
+    {
+        [TestMethod, Timeout(1000)]
```
DaveTryon, line 15: Should we have a test that pins the default values?

src/CoreTests/CustomObjects/ConfigTests.cs
```diff
@@ -67,28 +63,12 @@ public void BadGuidConfigTest()
...
+            {
```
DaveTryon, line 67: Should this be in CustomPropertyTests?

src/Core/CustomObjects/CustomProperty.cs
```diff
@@ -10,54 +10,51 @@ namespace Axe.Windows.Core.CustomObjects
...
+                        Type = CustomUIAPropertyType.Double;
+                        break;
+                    case "point":
+                        Type = CustomUIAPropertyType.Point;
+                        break;
+                    case "element":
+                        Type = CustomUIAPropertyType.Element;
+                        break;
+                    default:
+                        throw new InvalidDataException("Type in custom property definition is missing or invalid.");
```
DaveTryon, line 55:
```suggestion
                        throw new ArgumentExeption($"'${value}' is not a supported type", nameof(value));
```

DaveTryon (approved):
Approved with a suggestion

DaveTryon (approved):

PR #603, opened 2021-06-25
fix: Delete Interop.UIAutomationCore.dll from CLI projects

DaveTryon:
#### Details

The signed binary check is unhappy about Interop.UIAutomationCore.dll. We _could_ sign it with our 3rd party cert and ship it with our binaries, but since we embed the assembly in the [Desktop assembly](https://github.com/microsoft/axe-windows/blob/main/src/Desktop/Desktop.csproj#L48), there's no need to ship it.

##### Motivation

Don't ship unsigned files

#### Pull request checklist

### DIFF
--- a/src/CLI/CLI.csproj
+++ b/src/CLI/CLI.csproj
@@ -79,4 +79,8 @@
     </None>
   </ItemGroup>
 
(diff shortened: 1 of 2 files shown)


codeofdusk (approved):
LGTM

PR #611, opened 2021-07-01
chore: Package Interop in dual formats

DaveTryon:
#### Details

We've tried packaging Interop.UIAutomationCore.dll as a purely embedded assembly and as a purely non-embedded assembly. Neither worked completely. This PR uses both strategies--it embeds the assembly into Axe.Windows.Desktop (for the NuGet package consumption) and carries along the external version as a build artifact to keep the CLI happy. This increases the CLI size slightly (about 50KB on disk), but solves all of our scenarios. We've confirmed that both the CLI and AIWin work correctly with this change.

##### Motivation

Being able to release is important

#### Pull request checklist


codeofdusk:
> 50MB on disk

Either there's an order of magnitude problem or that's the understatement of the century!

codeofdusk (commented):
Looks good pending change to description.

codeofdusk (approved):

Issue #636, opened 2021-09-23
[BUG] WinForms edit control does not support TextPattern


lisli1:
Please check whether the bug has [already been filed](https://github.com/Microsoft/axe-windows/issues).

## Describe the bug

On .NET Framework, WinForms edit (numeric up down) control does not support the TextPattern

**To Reproduce**

## Expected behavior

Running AI Win on an app with the edit control should not flag an accessibility issue regarding support for TextPattern.

## Actual behavior

Running AI Win on an app with the edit control flags an accessibility issue: "An element of the given ControlType must support the Text pattern."

## Desktop (please complete the following information):

 - OS: [e.g. Windows 10 1809] (Get the version by running `winver` from the command line) Windows 10.0.19043 Build 19043
 - Accessibility Insights for Windows Version: 1.1.1689.1
 - Target Application: [e.g. Chrome, Wildlife Manager] any app with WinForms Edit(50004) ControlType
 - Target Application Version: [e.g. 22]

## Additional context

Priority requested -

PR #637, opened 2021-09-23
fix: Exclude WinForms edit control from ControlShouldSupportTextPattern rule

DaveTryon:
#### Details

As reported in #636, WinForms Edit controls don't support TextPattern. It's a known framework issue that we don't really expect app developers to try to fix. It's possible that a future .NET WinForms version might support this pattern, but there's no guarantee of _if_ or _when_ this might occur. To minimize noise, we've decided to exclude WinForms edit controls from this rule until a less noisy solution becomes available. 

##### Motivation

Remove noise, address #636

#### Pull request checklist

- [x] Addresses an existing issue: #636

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -19,7 +19,7 @@ namespace Axe.Windows.AutomationTests
         // These values should change only in response to intentional rule modifications
         const int WildlifeManagerKnownErrorCount = 12;
         const int Win32ControlSamplerKnownErrorCount = 0;
-        const int WindowsFormsControlSamplerKnownErrorCount = 6;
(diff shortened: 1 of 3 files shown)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -19,7 +19,7 @@ public class AutomationIntegrationTests
         // These values should change only in response to intentional rule modifications
         const int WildlifeManagerKnownErrorCount = 12;
         const int Win32ControlSamplerKnownErrorCount = 0;
-        const int WindowsFormsControlSamplerKnownErrorCount = 6;
+        const int WindowsFormsControlSamplerKnownErrorCount = 4;
```
DaveTryon, line 22: Sampler app has both an edit control and a spinner control, each of which used to fail.

RobGallo (approved):

PR #650, opened 2021-11-09
fix: Be more resilient to rule execution errors

DaveTryon:
#### Details

When a rule execution error occurs, it becomes a rule failure. We know relatively little about the specifics of these errors, and it's possible that they could succeed on a simple retry. This PR does just that--it allows for 3 consecutive attempts of each rule before finally giving up and reverting to the current behavior.

##### Motivation

Hopefully make the scans more resilient to transient failures.

##### Context

My first thought was to insert a new method between `RuleRule` and `RunRuleWorker`, and to allow it to manage the retry logic. That worked until I wanted to report the intermediate exceptions, and the retry and the exception reporting just made sense being in the same method.

#### Pull request checklist

- [ ] Addresses an existing issue:

### DIFF
--- a/src/Rules/RuleRunner.cs
+++ b/src/Rules/RuleRunner.cs
@@ -33,7 +33,7 @@ namespace Axe.Windows.Rules
                 };
             }
 
-            var retVal = RunRule(rule, element);
+            var retVal = RunRuleWithRetries(rule, element);
 
             return retVal;
         }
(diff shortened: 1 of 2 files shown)


src/Rules/RuleRunner.cs
```diff
@@ -44,35 +44,43 @@ public IEnumerable<RunResult> RunAll(IA11yElement element)
...
+                        return new RunResult
+                        {
+                            RuleInfo = rule.Info,
+                            element = element,
+                            EvaluationCode = EvaluationCode.RuleExecutionError,
+                            ErrorMessage = ex.ToString()
+                        };
+                    }
+                } // catch
+            }
```
chingucoding, line 83: Could a `for` loop be sufficient for this? Also although it be a VERY niche case, what if the different runs return different exceptions? I think it would be fine to just pick one but none the less maybe other folks have different opinions on this?

DaveTryon: Thanks, @chingucoding. I explicitly didn't want a for loop here so that we only had 2 exit paths--either we succeeded or we threw. I considered aggregating the exceptions, but when I followed how we use RunResult.ErrorMessage, we just throw it away further upstream. Each separate exception would get reported in line 68, so the telemetry would be able to view them all.

DaveTryon:
This was opened to stimulate discussion, and based on that discussion, we're not going to implement this at this time

src/Rules/RuleRunner.cs
chingucoding: Ah I see, thank you for the clarification @DaveTryon !

Issue #648, opened 2021-11-08
[BUG] We should not apply name rules for controls that are children of a CheckBox


DaveTryon:
Please check whether the bug has [already been filed](https://github.com/Microsoft/axe-windows/issues).

In #572, we chose to ignore text checks on controls that are children of buttons, since AT's will use the `Name` property that the Button provides, and we don't need to know how it is computed. We have a similar issue with CheckBoxes. Consider the following XAML:

```
        <CheckBox Margin="20" AutomationProperties.Name="Cat">
            <CheckBox.ContentTemplate>
                <DataTemplate>
                    <TextBlock>
                        <AccessText>_Cat</AccessText>
                    </TextBlock>
                </DataTemplate>
            </CheckBox.ContentTemplate>
        </CheckBox>
```

In this sample, AT's will read the Checkbox as "Cat Checkbox Not Checked C", which makes sense. When tested with Axe.Windows, though, we throw an error on the `TextBlock` because it has no Name property. In this trivial sample, the `TextBlock` can safely be removed. In the real-world report that we received, the `TextBlock` exists only to provide a visual style.

Since the `TextBlock` adds no value for AT's does the same logic that #572 applied to buttons also apply to the text in a CheckBox? I think it might...

**To Reproduce**

1. Build an app with the above XAML
2. Read the screen with an AT (I used NVDA) & it seems to make sense
3. Test it in Axe.Windows

## Expected behavior

Since the AT is happy, one could argue that no accessibility bug exists.

## Actual behavior

Axe.Windows complains about the name of the `TextBlock` containing only spaces

## Screenshots or .GIF

![image](https://user-images.githubusercontent.com/45672944/140820962-ae1a38f2-6cfb-4c68-89d1-85f807be95d7.png)

## Desktop (please complete the following information):

 - OS: Win10 21H1
 - Axe-Version: 1.1.1
 - Target Application: Custom WPF app

## Additional context

Priority requested -

DaveTryon:
This has been fixed in [Axe.Windows 1.1.2](https://github.com/microsoft/axe-windows/releases/tag/v1.1.2)

PR #651, opened 2021-11-10
fix: Ignore names on WPF Text inside a CheckBox

DaveTryon:
#### Details

As called out in #648, we should skip name checking of items inside a WPF CheckBox, since AT's will simply read the Name property of the CheckBox, which will be set either by the framework or overridden by the app developer.

##### Motivation

Handle case identified in #648.

##### Context

This is similar to #572, but handled differently. #572 changed the condition on a single rule, but this effectively causes us to ignore _all_ name-related rules within the scope of the fix. Happy to discuss pros & cons of the options, if desired.

#### Pull request checklist

- [x] Addresses an existing issue: #648

### DIFF
--- a/src/Rules/PropertyConditions/ElementGroups.cs
+++ b/src/Rules/PropertyConditions/ElementGroups.cs
@@ -141,7 +141,7 @@ namespace Axe.Windows.Rules.PropertyConditions
                 | ToolTip | Tree | TreeItem | Window
                 | trueWhenElementHasSiblingsOfSameType)
                 & ~IsWPFCheckBoxText
-                &~isWPFPopupOrContextMenu;
+                & ~isWPFPopupOrContextMenu;
         }
 
         private static Condition CreateNameOptionalCondition()
--- a/src/Rules/PropertyConditions/ElementGroups.cs
+++ b/src/Rules/PropertyConditions/ElementGroups.cs
@@ -31,6 +31,8 @@ namespace Axe.Windows.Rules.PropertyConditions
         public static Condition ParentWPFDataItem = CreateParentWPFDataItemCondition();
         public static Condition WPFScrollBarPageButtons = CreateWPFScrollBarPageButtons();
         public static Condition NonFocusableSliderButtons = CreateNonFocusableSliderButtonsCondition();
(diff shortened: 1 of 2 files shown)


src/Rules/PropertyConditions/ElementGroups.cs
```diff
@@ -31,6 +31,8 @@ static class ElementGroups
         public static Condition ParentWPFDataItem = CreateParentWPFDataItemCondition();
         public static Condition WPFScrollBarPageButtons = CreateWPFScrollBarPageButtons();
         public static Condition NonFocusableSliderButtons = CreateNonFocusableSliderButtonsCondition();
+        public static Condition IsCheckBoxText = Text & AnyAncestor(CheckBox);
+        public static Condition IsWPFCheckBoxText = WPF & IsCheckBoxText;
         public static Condition NameRequired = CreateNameRequiredCondition();
```
DaveTryon, line 36: Order here is drive by the fact that `CreateNameRequiredCondition` depends on `IsWPFCheckBoxText`.

src/RulesTest/PropertyConditions/ElementGroupsTests.cs
```diff
@@ -280,5 +280,54 @@ public void IsButtonText_MatchExpected()
...
+                Assert.IsFalse(ElementGroups.IsCheckBoxText.Matches(e));
+
+                e.ControlTypeId = Text;
+                Assert.IsFalse(ElementGroups.IsCheckBoxText.Matches(e));
+
+                e.Parent = parent;
+                parent.ControlTypeId = CheckBox;
+                Assert.IsTrue(ElementGroups.IsCheckBoxText.Matches(e));
+
+                parent.ControlTypeId = Window;
```
RobGallo, line 302: I think we need to add a check for another type of child element under a checkbox parent -- that seems like a case that isn't covered. I'm also wondering if the change from Checkbox to Window isn't redundant with the checks above when e has no parent at all.

DaveTryon: OK, I've reduced this to 4 cases: Element is Button or Window, parent is Null or CheckBox.

RobGallo (approved):

PR #655, opened 2021-11-17
feat(compliance-fall-21): Exclude non-localizable fields from CLI help output

DaveTryon:
#### Details

When running the CLI, we use localized text for the fields that we control, but the fields for `--help` and `--version` are hardcoded:

```
AxeWindowsCLI 1.0.10
Copyright c 2020
  --processid                Process Id
  --processname              Process Name
  --outputdirectory          Output directory
  --scanid                   Scan ID
  --verbosity                Verbosity level (Quiet/Default/Verbose)
  --showthirdpartynotices    Display Third Party Notices (opens file in browser
                             without executing scan). If specified, all other
                             options will be ignored.
  --delayinseconds           How many seconds to delay before triggering the
                             scan. Valid range is 0 to 60 seconds, with a
                             default of 0.
  --customuia                The path to a configuration file specifying custom
                             UI Automation attributes
  --help                     Display this help screen.
  --version                  Display version information.
```

The --help and --version fields are hardcoded in `ComandLineParser` and frankly aren't very interesting for our purposes, so we'll just exclude then from the output. Here's the output with this change (same as before, but without the --help or --version options):

```
AxeWindowsCLI 1.1.10
Copyright c 2020
  --processid                Process Id
  --processname              Process Name
  --outputdirectory          Output directory
  --scanid                   Scan ID
  --verbosity                Verbosity level (Quiet/Default/Verbose)
  --showthirdpartynotices    Display Third Party Notices (opens file in browser
                             without executing scan). If specified, all other
                             options will be ignored.
  --delayinseconds           How many seconds to delay before triggering the
                             scan. Valid range is 0 to 60 seconds, with a
                             default of 0.
  --customuia                The path to a configuration file specifying custom
                             UI Automation attributes
```

##### Motivation

Part of global readiness compliance work. Make it simpler to localize the CLI if/when that ever happens.

##### Context

The code changes are based on the documented way to handle this at https://github.com/commandlineparser/commandline/wiki/How-To#q1

#### Pull request checklist

### DIFF
--- a/src/CLI/Program.cs
+++ b/src/CLI/Program.cs
@@ -3,6 +3,7 @@
 
 using Axe.Windows.Automation;
 using CommandLine;
+using CommandLine.Text;
 using System;
 using System.IO;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


jlperkins:
Seems the [CLI README](https://github.com/microsoft/axe-windows/blob/main/src/CLI/README.MD) should be updated to reflect this change as well.  Should that be part of this PR?

jlperkins (approved):
LGTM!

Issue #659, opened 2021-11-30
[False Positive] Win32 SplitButton creates error on inner element


DaveTryon:
## Describe the bug

The Win32 SplitButton control is implemented via nested SplitButton elements. The parent element supports the ExpandCollapse pattern, but the child element does not. We should modify the `ControlShouldSupportExpandCollapsePattern` to exclude the child SplitButton element on Win32.

**To Reproduce**

1. Start the [Win32ControlSampler](https://github.com/microsoft/axe-windows/blob/main/tools/Win32ControlSampler/Win32ControlSampler.exe) app
2. Scan it with the Axe.Windows CLI
3. Load the a11ytest file into Accessibility Insights for Windows
4. Expand to the SplitButton child, and see the following error: **An element of the following ControlType must support the ExpandCollapse pattern**

## Expected behavior
The ExpandCollapse pattern is supported on the parent; its absence on the child should not produce an error

## Actual behavior
The code incorrectly complains that the child has no ExpandCollapse pattern

## Desktop (please complete the following information):

 - OS: Windows 10 21H1
 - Accessibility Insights for Windows Version: 1.1.1789.1
 - Target Application: Win32ControlSampler.exe

## Additional context

This was created in response to investigation of #653

DaveTryon:
We discussed this in triage on 11/29

chingucoding:
What would be the best way to fix this? Check if we have two SplitButtons nested inside each other and skip the second one?

DaveTryon:
@chingucoding, it would be along the same lines as #575 or #651. In `ControlShouldSupportExpandCollapsePattern.Condition`, I'd probably create a local variable that specifically excludes a child SplitButton:
```
var eligibleSplitButton = SplitButton & ~Parent(SplitButton);
```
...then replace `SplitButton` with `eligibleSplitButton` in line 34.

This adds some complexity to the unit test since what used to be 2 test cases now becomes several.

DaveTryon:
Marking this as resolved until we release a new version of Axe.Windows

DaveTryon:
This has been fixed in [Axe.Windows 1.1.2](https://github.com/microsoft/axe-windows/releases/tag/v1.1.2)

PR #661, opened 2021-12-01
fix: Fix win32 SplitButton nesting leading to faulty expand collapse pattern rule errors

chingucoding:
#### Details

In win32 app, the SplitButton is implented using nested SplitButtons. Only the parent one implements the expand collapse pattern which is fine, however the child does not. This PR updates the rule to exclude direct SplitButton childrens of SplitButtons from the ExpandCollapsePatternRule.

##### Motivation

Addresses #659 

##### Context

N/A

#### Pull request checklist

- [x] Addresses an existing issue: #659

### DIFF
--- a/src/Rules/Library/ControlShouldSupportExpandCollapsePattern.cs
+++ b/src/Rules/Library/ControlShouldSupportExpandCollapsePattern.cs
@@ -30,8 +30,9 @@ namespace Axe.Windows.Rules.Library
 
         protected override Condition CreateCondition()
         {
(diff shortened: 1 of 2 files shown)


DaveTryon (approved):

DaveTryon:
Thanks, @chingucoding! Merged :)

chingucoding:
Thank you for merging @DaveTryon ! I noticed that the solution is still a VS 2019 solution and the test projects are .NET Core 3.1 projects. Are there plans to change this? Would this be better discussed/tracked in a separate issue?

PR #668, opened 2022-02-01
feat: Improve color detection

DaveTryon:
#### Details

Our current automated color detection is based on an algorithm that was primarily designed for portable devices, which have limited memory, limited CPU availability, and large text size relative to the screen size. None of these conditions apply to a windows development machine, and our automatic color detection suffers as a result. This is reported at https://github.com/microsoft/accessibility-insights-windows/issues/996.

This PR adds a new algorithm for automated color detection. The algorithm works as follows:
1. Each scanline in the bitmap is evaluated as a separate entity
2. For each scanline, the colors are recorded
3. For each scanline, the most frequently appearing color is assumed to the background color
4. For each scanline, the color with the highest contrast to the assumed background color is assumed to be the foreground color for that scanline
5. For each scanline, we count the number of times that we transition back to the assumed background color. We call this the transition count. This number will be larger for lines that contain actual text
6. Iterate through each scanline result. With each scanline counting as 1 vote, the background colors are added up. The most commonly occurring background color is assumed to the be the background color of the overall image
7. A confidence value is assigned to the background color, based on how many scanlines agreed on that value
8. Scanlines that match the background color are considered. The foreground colors are aggregated into similar colors (this is to help with antialiasing), so we end up with aggregated ballots. Each aggregated ballot receives as many votes as the sum of all transitions of all scanlines that are represented by that aggregated ballot. The highest contrast is assumed to the be the foreground color
9. A confidence value is assigned to the foreground color, based on how many aggregated ballots agreed on the color
10. The overall confidence is the lower value of the foreground and background color confidences

The confidence thresholds are user configurable. I replaced the old `DefaultColorContastConfig` class with a builder class. This will be a breaking change when we pick it up in AIWin, but the changes are outside of the scope of the Automation interface, so that's not a problem that we need to worry about.
 
The old code included some test images that weren't actually scanned because they were known to be problematic. These images are all scanned by the new code. There are also additional images of controls that were problematic.

##### Motivation

Improve automated color detection--this will make AIWin's color contrast analyzer more useful, and be a necessary first step if we want to include automatic detection of color blindness issues.

#### Pull request checklist

- [x] Addresses an existing issue: Part of addressing https://github.com/microsoft/accessibility-insights-windows/issues/996


src/Desktop/ColorContrastAnalyzer/ImageCollection.cs
```diff
@@ -58,13 +99,15 @@ public ColorContrastResult RunColorContrastCalculation()
...
-                    if (newResult.ConfidenceValue() == ColorContrastResult.Confidence.High)
+                    if (newResult.Confidence == Confidence.High)
                     {
                         result = newResult;
                         break;
                     }
-                    else if (newResult.ConfidenceValue() == ColorContrastResult.Confidence.Mid &&
-                      result.ConfidenceValue() == ColorContrastResult.Confidence.Low)
+                    else if (newResult.Confidence == Confidence.Mid &&
+#pragma warning disable CA1508 // Avoid dead conditional code
```
karanbirsingh, line 108: Can you share some context around this warning? It feels like we can remove line `109` because it will always be true. By this point in the code possible values for `result` include:
- `newResult` in the current iteration (then `newResult.Confidence` must equal `result.Confidence`
- `newResult` from a previous iteration (then `result.Confidence` must be high or mid because we don't replace `result` otherwise)

Otherwise I agree it's probably good to leave the old algorithm mostly alone to preserve its semantics. If so, it might be good to explicitly mention that near the warning.

src/Desktop/ColorContrastAnalyzer/RowResultV2Accumulator.cs
```diff
@@ -0,0 +1,223 @@
...
+                }
+            });
+
+            var sortedBackgroundColorBallots = backgroundColorBallots.OrderByDescending(x => x.Value);
+
+            return BuildColorVoteInfo(sortedBackgroundColorBallots, GetTotalVoteCountFromAllBallots);
+        }
+
+        /// <summary>
+        /// Counts up the ForegrounColor from each entry in _rowResults, then
```
karanbirsingh, line 80: (nit) small typo, ForegroundColor

src/Desktop/ColorContrastAnalyzer/AnalyzerVersion.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Axe.Windows.Desktop.ColorContrastAnalyzer
+{
+    public enum AnalyzerVersion
+    {
+        Default, // Use whatever the current default happens to be
+        V1,      // Decides based on a single scanline
```
karanbirsingh, line 9: It looks like `GetBinaryRowSearchIterator()` inside `ImageCollection.cs` looks at at least one scanline - its comments say: "A special iterator, that looks at the middle of an image first, followed by recursively looking at the upper half and lower half of the two pieces of image, until the given samples are some distance apart." If I'm reading that right, we may want to amend this description.

src/Desktop/ColorContrastAnalyzer/RowResultV2Accumulator.cs
```diff
@@ -0,0 +1,223 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
karanbirsingh, line 1: The new approach is more performance-intensive. Have you tried running it on larger bitmaps? I'm wondering what would happen if we focused on an element with a very large bounding rectangle from AI-Windows; it would be nice to understand how this scales with bitmap size to avoid app stutters.

src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+    class ColorContrastConfig : IColorContrastConfig
+    {
+        internal const int DefaultMaxTextThickness = 20;
+        internal const int DefaultMinColorTransitions = 4;
+        internal const int DefaultMinSpaceBetweenSamples = 12;
+        internal const int DefaultTransitionCountDominanceFactor = 2;
+        internal const AnalyzerVersion DefaultAnalyzerVersion = AnalyzerVersion.V1;
+        internal const double DefaultHighConfidenceThreshold = 0.1;
+        internal const double DefaultMidConfidenceThreshold = 0.3;
+        internal const double DefaultLowConfidenceThreshold = 0.5;
```
karanbirsingh, line 17: I flipped back and forth on whether it'd be more understandable to invert the confidence thresholds. At first I was confused that the HighConfidenceThreshold was lower than the MidConfidenceThreshold. But it made more sense as I dug deeper into V2's approach. If we want to change this, we could invert these and pass (1 - ConfidenceThreshold) to the underlying V2 calculator. I am fine either way.

src/Desktop/ColorContrastAnalyzer/ImageCollection.cs
DaveTryon: The Analyzer claims the condition is always false, but testing in the debugger proves otherwise. I've restructured it a bit (fixed a bug in the process) and updated the comments

src/Desktop/ColorContrastAnalyzer/RowResultV2Accumulator.cs
DaveTryon: Fixed

src/Desktop/ColorContrastAnalyzer/AnalyzerVersion.cs
DaveTryon: The existing code may scan multiple lines before making its decision, but the ultimate decision is based on exactly one scanline. I updated the comment a bit. If the V2 scanning works out, I'd eventually like to remove the V1 analyzer completely. If perf proves to be a problem, I could also see a world where we dynamically choose one analyzer or the other, possibly based on the size of the bitmap in question

src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
DaveTryon: I waffled on this, partly because it felt like I was exposing too much of the implementation in the names. After taking another look, intuitive seems good. I've update the constants and code so that High confidence defaults to 90%, Mid to 70%, and Low to 50%.

src/Desktop/ColorContrastAnalyzer/RowResultV2Accumulator.cs
DaveTryon: I don't yet have numbers, but in practice it's perhaps less of a difference than it might seem. For a large bitmap (an entire app, for instance), you're likely to have a lot of controls and not be able to get quickly to a High confidence number. If this happens, the old code will scan about 1/12 of the scanlines (by default) before giving up.

I definitely want to benchmark this (and possibly see if using the `System.Drawing.Color` _struct_ instead of the `ColorContrastAnalyzer.Color` _class_ will improve our perf), but this PR is already really large just getting the functionality in--perf measurements and tuning feel like they belong in a separate PR

src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+            HighConfidenceThreshold = highConfidenceThreshold.HasValue ?
+                highConfidenceThreshold.Value : DefaultHighConfidenceThreshold;
+            MidConfidenceThreshold = midConfidenceThreshold.HasValue ?
+                midConfidenceThreshold.Value : DefaultMidConfidenceThreshold;
+            LowConfidenceThreshold = lowConfidenceThreshold.HasValue ?
+                lowConfidenceThreshold.Value : DefaultLowConfidenceThreshold;
+
+            if (HighConfidenceThreshold <= MidConfidenceThreshold)
+            {
+                throw new ArgumentException("High Confidence threshold must be smaller than Mid confidence threshold");
```
karanbirsingh, line 63: Thanks for inverting the thresholds, this seems more intuitive. I think we need to update these strings to match (`smaller` should be `larger`).

src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            if (MidConfidenceThreshold <= LowConfidenceThreshold)
+            {
+                throw new ArgumentException("Mid Confidence threshold must be smaller than Low confidence threshold");
```
karanbirsingh, line 67: Thanks for inverting the thresholds, this seems more intuitive. I think we need to update these strings to match (smaller should be larger).

karanbirsingh (approved):

src/Desktop/ColorContrastAnalyzer/ColorContrastConfigBuilder.cs
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        public ColorContrastConfigBuilder WithAnalyzerVersion(AnalyzerVersion analyzerVersion)
+        {
+            _analyzerVersion = analyzerVersion;
+            return this;
+        }
+
+        /// <summary>
+        /// Sets the maximum threshold for high confidence. Default value is 0.1.
```
karanbirsingh, line 55: Default value comments in this file need to be updated after latest revision

src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
DaveTryon: Fixed

src/Desktop/ColorContrastAnalyzer/ColorContrastConfig.cs
DaveTryon: Fixed

src/Desktop/ColorContrastAnalyzer/ColorContrastConfigBuilder.cs
DaveTryon: Fixed

karanbirsingh (approved):

jalkire (approved):

PR #675, opened 2022-03-25
chore: Run automatic cleanup tool

DaveTryon:
#### Details

Microsoft publishes a Visual Studio extension called [Power Commands for Visual Studio](https://marketplace.visualstudio.com/items?itemName=VisualStudioPlatformTeam.PowerCommandsforVisualStudio) that can apply normalizations across an entire project from a single command. After creating [AIWin PR 1303](https://github.com/microsoft/accessibility-insights-windows/pull/1303), I ran the same tool on Axe.Windows. I ran the extension using the default set of fixers, which are:

- Format document (includes things like consistent spacing, braces, etc.)
- Remove unnecessary usings
- Sort usings
- Apply file header preferences

The last fixer did nothing to our code, since we have no configured file header preferences. The first 3 fixers, though, found and fixed issues across the entire solution.

This PR is fairly large (nearly 20% of our files), but the individual changes are trivial--the vast majority of them are tiny whitespace tweaks that just got missed in previous changes. I've reviewed them, but it would be good to get a second reviewer

##### Motivation

Code consistency

##### Context

The first commit here is straight from running the tool. Upon reviewing, I found an extra line that provided no value, so I went ahead and removed it manually--that's the second commit.

Newer versions of VS (which we get in the PR builds) support [EditorConfig files](https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2019), which can provide enforcement and even (I think) format on save support. Let's try to add EditorConfig in a future PR

#### Pull request checklist

### DIFF
--- a/src/CurrentFileVersionCompatibilityTests/Program.cs
+++ b/src/CurrentFileVersionCompatibilityTests/Program.cs
@@ -1,6 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Axe.Windows.Actions;
+using Axe.Windows.Actions.Contexts;
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
+using Axe.Windows.Core.Results;
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/src/CurrentFileVersionCompatibilityTests/Program.cs
+++ b/src/CurrentFileVersionCompatibilityTests/Program.cs
@@ -1,11 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Axe.Windows.Actions;
-using Axe.Windows.Actions.Contexts;
-using Axe.Windows.Core.Bases;
-using Axe.Windows.Core.Enums;
-using Axe.Windows.Core.Results;
 using System;
 using System.Collections.Generic;
 using System.Linq;
--- a/src/Desktop/Styles/StyleId.cs
+++ b/src/Desktop/Styles/StyleId.cs
@@ -20,21 +20,21 @@ namespace Axe.Windows.Desktop.Styles
         /// https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-style-identifiers
         /// </summary>
 #pragma warning disable CA1707 // Identifiers should not contain underscores
-        public const int StyleId_Custom       = 70000;
-        public const int StyleId_Heading1     = 70001;
-        public const int StyleId_Heading2     = 70002;
-        public const int StyleId_Heading3     = 70003;
-        public const int StyleId_Heading4     = 70004;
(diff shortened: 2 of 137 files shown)


src/Desktop/Styles/StyleId.cs
```diff
@@ -20,21 +20,21 @@ public class StyleId : TypeBase
         /// https://docs.microsoft.com/en-us/windows/win32/winauto/uiauto-style-identifiers
         /// </summary>
 #pragma warning disable CA1707 // Identifiers should not contain underscores
-        public const int StyleId_Custom       = 70000;
```
DaveTryon, line 23: I kind of like the alignment, but I'm willing to give it up to get the rest of the code consistently formatted

chingucoding: Totally with you on this but I'm not sure how easy it would be to work against the formatter in the long term.

chingucoding (approved):

src/Desktop/UIAutomation/CustomObjects/Registrar.cs
```diff
@@ -5,7 +5,6 @@
 using Axe.Windows.Core.CustomObjects;
 using Axe.Windows.Core.CustomObjects.Converters;
 using Axe.Windows.Core.Enums;
-using Interop.UIAutomationCore;
```
chingucoding, line 8: I think this needs to be added again.

src/DesktopTests/UIAutomation/CustomObjects/RegistrarTests.cs
```diff
@@ -4,7 +4,6 @@
 using Axe.Windows.Core.CustomObjects;
 using Axe.Windows.Core.CustomObjects.Converters;
 using Axe.Windows.Desktop.UIAutomation.CustomObjects;
-using Interop.UIAutomationCore;
```
chingucoding, line 7: Same here, needs to be re-added.

src/CurrentFileVersionCompatibilityTests/Program.cs
```diff
@@ -1,11 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using Axe.Windows.Actions;
-using Axe.Windows.Actions.Contexts;
-using Axe.Windows.Core.Bases;
-using Axe.Windows.Core.Enums;
-using Axe.Windows.Core.Results;
```
chingucoding, line 8: Those need to be included.

chingucoding (changes requested):
It seems that some using statements were removed while they shouldn't.

chingucoding (approved):

karanbirsingh (approved):
Looks good! Should we add a link to the extension in our dev docs? (ok in a follow-up PR)

codeofdusk:
Changes to `CustomObjects` files seem okay.

DaveTryon:
> Looks good! Should we add a link to the extension in our dev docs? (ok in a follow-up PR)

If possible, we'll set up a `.editorconfig` file to the extension will no longer be necessary.

Issue #665, opened 2022-01-11
[Feature Request] Enable automated scans of processes with multiple top-level windows


chingucoding:
As we worked on adding automated Axe tests for [Files](https://github.com/files-community/files) we stumbled upon the fact that Files only has one process ID but it can spawn multiple windows. After some short testing, we noticed that the most recently focused window was the one to be scanned, a11y issues on the other windows were ignored. My question would be, if that is the expected behavior or if there are better ways to test in those cases. If this is something deeper in the UIA library, i.e. Windows, feel free to close this issue.

chingucoding:
Hi @dbjorge , thank you for your detailed reply!

I think I would definitely use a feature that would allow me to scan all windows of a given process. My ideal solution would be having a configuration option when creating a scanner to specify whether to either scan the "first" window or scan all windows. It could look something like this:

```c#
 // Create config to specifically target a process
 var myConfigBuilder = Config.Builder.ForProcessId(1234);

myConfigBuilder.WithMultiWindowStrategy(MultiWindowStrategy.LastFocused | MultiWindowStrategy.All)

 var myConfig = myConfigBuilder.build();
 var scanner = ScannerFactory.CreateScanner(myConfig);
```

The default option for this would be `LastFocused` to not break the previous behavior. If there was an option to filter to scan a specific window, I would probably only use it to scan all windows open by that app.

chingucoding:
Also, if we decide on an API to use for this, I would be more than happy to contribute such a feature!

RobGallo:
@chingucoding , Sorry it's taken us a minute to get back to you. Thanks for your patience. The API we're considering at the moment is probably the simplest option: adding a method to the builder which takes a 0 based window index for the process (i.e., Config.Builder.WithWindowIndex(int index)). In this way, you would pass in the process id and the window index and the result returned would be for the specified top-level window. Would that solve the issue you experienced? And do you foresee any problems with that approach?

chingucoding:
[...]

PR #676, opened 2022-03-28
[feat(scan-multiple-windows)] Add code to scan multiple root-level windows

RobGallo:
#### Details

This change refactors TargetElementLocator so that it returns all root-level window elements for a given process id. The scanner itself still only scans the first window returned, which is the same as the existing behavior. 

##### Motivation

This commit partially addresses  #665.

##### Context

This change was kept as small as possible. But it should pave the way for a change where the user can specify that all root-level windows are scanned.

#### Pull request checklist

### DIFF
--- a/src/Automation/TargetElementLocator.cs
+++ b/src/Automation/TargetElementLocator.cs
@@ -1,28 +1,40 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using Axe.Windows.Actions.Contexts;
+using Axe.Windows.Automation.Resources;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Desktop.UIAutomation;
 using System;
+using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 
 namespace Axe.Windows.Automation
 {
     class TargetElementLocator : ITargetElementLocator
     {
-        public A11yElement LocateRootElement(int processId)
+        public IEnumerable<A11yElement> LocateRootElements(int processId)
         {
             try
             {
-                var element = A11yAutomation.ElementFromProcessId(processId);
-
-#pragma warning disable CA2000 // Call IDisposable.Dispose()
-                return new ElementContext(element).Element;
-#pragma warning restore CA2000
+                var desktopElements = A11yAutomation.ElementsFromProcessId(processId);
+                return GetA11YElementsFromDesktopElements(desktopElements);
             }
             catch (Exception ex)
             {
                 throw new AxeWindowsAutomationException(string.Format(CultureInfo.InvariantCulture, DisplayStrings.ErrorFailToGetRootElementOfProcess, processId, ex), ex);
+            }
+        }
+
+        private static IEnumerable<A11yElement> GetA11YElementsFromDesktopElements(IEnumerable<DesktopElement> desktopElements)
+            {
+            if (!desktopElements.Any()) throw new ArgumentException(ErrorMessages.NoDesktopElements, nameof(desktopElements));
+
+            foreach (var e in desktopElements)
+            {
+#pragma warning disable CA2000 // Call IDisposable.Dispose()
+                yield return new ElementContext(e).Element;
+#pragma warning restore CA2000
             }
         }
     }
--- a/src/Desktop/UIAutomation/A11yAutomation.cs
+++ b/src/Desktop/UIAutomation/A11yAutomation.cs
@@ -153,9 +153,8 @@ namespace Axe.Windows.Desktop.UIAutomation
             for (int i = 0; i < elementArray.Length; ++i)
             {
                 var uiaElement = elementArray.GetElement(i);
-                if (DesktopElement.IsFromCurrentProcess(uiaElement)) continue;
-
-                var e = new DesktopElement(uiaElement, true, false);
+                var e = ElementFromUIAElement(uiaElement);
+                if (e == null) continue;
 
                 e.PopulateMinimumPropertiesForSelection();
 
--- a/src/Desktop/UIAutomation/A11yAutomation.cs
+++ b/src/Desktop/UIAutomation/A11yAutomation.cs
@@ -87,6 +87,7 @@ namespace Axe.Windows.Desktop.UIAutomation
                 {
                     Marshal.ReleaseComObject(condition);
                 }
+
                 if (elementArray != null)
                 {
                     Marshal.ReleaseComObject(elementArray);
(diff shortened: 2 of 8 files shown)


src/Automation/TargetElementLocator.cs
```diff
@@ -1,29 +1,41 @@
...
+                return GetA11YElementsFromDesktopElements(desktopElements);
             }
             catch (Exception ex)
             {
                 throw new AxeWindowsAutomationException(string.Format(CultureInfo.InvariantCulture, DisplayStrings.ErrorFailToGetRootElementOfProcess, processId, ex), ex);
             }
         }
+
+        private static IEnumerable<A11yElement> GetA11YElementsFromDesktopElements(IEnumerable<DesktopElement> desktopElements)
+            {
```
DaveTryon, line 30: Extra tab
```suggestion
        {
```

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -40,20 +41,33 @@ private static IUIAutomation GetUIAutomationInterface()
...
+        {
+            var elements = ElementsFromProcessId(pid);
+            return elements != null ? elements.First() : null;
+        }
+
+        /// <summary>
+        /// Get DesktopElements based on Process Id.
+        /// </summary>
+        /// <param name="pid"></param>
+        /// <returns>return null if fail to get elements by process Id</returns>
```
DaveTryon, line 53: This feels like it's missing a word. Maybe:
```suggestion
        /// <returns>return null if we fail to get elements by process Id</returns>
```

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -73,8 +87,13 @@ public static DesktopElement ElementFromProcessId(int pid)
                 {
                     Marshal.ReleaseComObject(condition);
                 }
-            }
-            return element;
+                if (elementArray != null)
+                {
+                    Marshal.ReleaseComObject(elementArray);
+                }
+                    }
```
DaveTryon, line 94: Indent:
```suggestion
            }
```

src/Automation/TargetElementLocator.cs
```diff
@@ -1,29 +1,41 @@
...
-
-#pragma warning disable CA2000 // Call IDisposable.Dispose()
-                return new ElementContext(element).Element;
-#pragma warning restore CA2000
+                var desktopElements = A11yAutomation.ElementsFromProcessId(processId);
+                return GetA11YElementsFromDesktopElements(desktopElements);
             }
             catch (Exception ex)
             {
                 throw new AxeWindowsAutomationException(string.Format(CultureInfo.InvariantCulture, DisplayStrings.ErrorFailToGetRootElementOfProcess, processId, ex), ex);
```
DaveTryon, line 25: Should this error be renamed from `DisplayStrings.ErrorFailToGetRootElementOfProcess` to something that reflects the potential for multiple elements?

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -124,6 +143,26 @@ private static DesktopElement ElementFromUIAElement(IUIAutomationElement uia)
...
+        /// Get DesktopElements from UIAElements.
+        /// </summary>
+        /// <param name="uia"></param>
+        /// <returns>An IEnumerable of <see cref="DesktopElement"/></returns>
+        private static IEnumerable<DesktopElement> ElementsFromUIAElements(IUIAutomationElementArray elementArray)
+        {
+            for (int i = 0; i < elementArray.Length; ++i)
+            {
+                var uiaElement = elementArray.GetElement(i);
+                if (DesktopElement.IsFromCurrentProcess(uiaElement)) continue;
```
DaveTryon, line 156: Do we need this guard? The guard in `ElementFromUIAElement` exists because that code gets called in response to mouse hover and we explicitly exclude AIWin's process. I'm not sure it applies in the case that we're addressing here

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -124,6 +143,26 @@ private static DesktopElement ElementFromUIAElement(IUIAutomationElement uia)
...
+        /// <summary>
+        /// Get DesktopElements from UIAElements.
+        /// </summary>
+        /// <param name="uia"></param>
+        /// <returns>An IEnumerable of <see cref="DesktopElement"/></returns>
+        private static IEnumerable<DesktopElement> ElementsFromUIAElements(IUIAutomationElementArray elementArray)
+        {
+            for (int i = 0; i < elementArray.Length; ++i)
+            {
+                var uiaElement = elementArray.GetElement(i);
```
DaveTryon, line 155: 2 thoughts from comparing this code with what's in `ElementFromUIAElement`:
1. Do we need a call to `Marshal.ReleaseComObject` somewhere in this method, perhaps on `elementArray`?
2. If yes, can this loop just call `ElementFromUIAElement` for each item in the array?

RobGallo: Updating the code to call ElementFromUIAElement instead of duplicating the logic. The element array is released in the calling function ElementsFromProcessId.

src/Automation/TargetElementLocator.cs
RobGallo: Suitably renamed

src/Desktop/UIAutomation/A11yAutomation.cs
RobGallo: We may not need it. But it will now be checked by ElementFromUIAElement and I don't think it will hurt anything.

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -124,6 +144,36 @@ private static DesktopElement ElementFromUIAElement(IUIAutomationElement uia)
...
+        /// Get DesktopElements from UIAElements.
+        /// </summary>
+        /// <param name="uia"></param>
+        /// <returns>An IEnumerable of <see cref="DesktopElement"/></returns>
+        private static IEnumerable<DesktopElement> ElementsFromUIAElements(IUIAutomationElementArray elementArray)
+        {
+            if (elementArray == null) throw new ArgumentNullException(nameof(elementArray));
+
+            var count = elementArray.Length;
+            if (count <= 0) return null;
```
DaveTryon, line 157: Suggestion: Instead of returning `null` here, can we return either the empty `List` or `Enumerable.Empty<DesktopElement>` ? Looking upstream, if we return `null` here, then we use that parameter to call Linq's `Any()` method, and I think that's going to trigger a `NullReferenceException`. We could add a guard before calling `Any()`, but since it already checks for an empty set, that's one less case to cover

DaveTryon, line 157: Suggestion: Instead of returning `null` here, can we return either the empty `List` or `Enumerable.Empty<DesktopElement>` ? Looking upstream, if we return `null` here, then we use that parameter to call Linq's `Any()` method, and I think that's going to trigger a `NullReferenceException`. We could add a guard before calling `Any()`, but since it already checks for an empty set, that's one less case to cover

RobGallo: Updated

DaveTryon (approved):

PR #677, opened 2022-03-31
chore: Use actual size bitmap from Edge in test

DaveTryon:
#### Details

If #668, we added a bitmap based on https://github.com/microsoft/accessibility-insights-windows/issues/996. We added a bitmap that was greatly expanded, instead of the actual size. We're seeing that the scan frequently takes longer than the 2 second timeout when running unit tests in the build pipeline. We could increase the timeout, but it seemed better to just use the actual size image.

##### Motivation

Stabilize the CI build

#### Pull request checklist


codeofdusk:
> If #668, 

Guessing you meant "*in* #668" here in the description?

jalkire (approved):

PR #682, opened 2022-04-17
[feat] Add multi window scan support

chingucoding:
#### Details

This PR adds support for multi window scanning and is based off of https://github.com/DaveTryon/axe-windows/pull/35 (thank you for the help @DaveTryon !).

##### Motivation

Addresses issue #665 

##### Context

Open questions (also commented on the necessary lines of code):
1. The return type of the `Scanner.Scan` API changed and it is a breaking change. What is the best way to proceed with this?
2. Should single window scans output the regular file names while multi window scan outputs files with special suffixes or should the behavior be always the same regardless of the number of scanned windows?

#### Pull request checklist

- [x] Addresses an existing issue: #665


src/Automation/SnapshotCommand.cs
```diff
@@ -35,28 +36,46 @@ public static ScanResults Execute(Config config, IScanTools scanTools)
...
+            {
+                if (targetCount > 0)
+                {
+                    results.OutputFile = WriteOutputFiles(config.OutputFileFormat, scanTools, element, elementId, (name) => $"{name}_{targetIndex}_of_{targetCount}");
+                }
+                else 
+                {
+                    results.OutputFile = WriteOutputFiles(config.OutputFileFormat, scanTools, element, elementId, null);
+                }
+            }
```
chingucoding, line 73: @DaveTryon This is a modified version of your code. I am not sure if the target count query makes sense since this can never be smaller than 1. Should it be `targetCount > 1` to restore the existing behavior for single window scan file output?

src/Automation/Interfaces/IScanner.cs
```diff
@@ -24,7 +25,7 @@ public interface IScanner
         /// </remarks>
         /// <returns>Information about the scan and any issues detected</returns>
         /// <exception cref="AxeWindowsAutomationException"/>
-        ScanResults Scan();
+        IReadOnlyCollection<ScanResults> Scan();
```
chingucoding, line 28: This is a breaking change for the public API, is it fine to just change it or is there a better way to do this and prevent breaking customers code.

chingucoding: Also, I'm inclined to say that leaving this as read only collection is advantages since customers might want to verify that every window is scanned or want to iterate using for loops (though in that case, LINQ is probably also needed to achieve it with index based loops).

src/Automation/SnapshotCommand.cs
DaveTryon: It should be `targetCount > 1`

src/Automation/Interfaces/IScanner.cs
DaveTryon: Good catch on the breaking change--that's something we want to avoid. Maybe we add a new entry point, something like:
```
IReadOnlyCollection<ScanResult> Scan(string scanId, bool enableMultipleWindows)
```
then we refactor the existing Scan implementations to become wrappers for the new one (and the existing entry points pass in false for `enableMultipleWindows`. We could even add an `EnableMultipleWindows` property to the `Config` class and propagate the control of the behavior all the way to the CLI command line.

chingucoding: That sounds like a good idea, I'll add the second method than and also add a new field to the config class.

chingucoding: Changes are implemented now. I also reworded the documentation slightly, but of course I would love the teams input on this.

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -126,8 +126,8 @@ public void Execute_TargetElementLocatorReturnsNull_PassesNullToScan()
             _scanToolsMock.Setup(x => x.Actions).Returns(_actionsMock.Object);
             _scanToolsMock.Setup(x => x.NativeMethods).Returns(_nativeMethodsMock.Object);
 
-            _targetElementLocatorMock.Setup(x => x.LocateRootElements(It.IsAny<int>())).Returns<IEnumerable<A11yElement>>(null);
-            _actionsMock.Setup(x => x.Scan(null, It.IsNotNull<ScanActionCallback<ScanResults>>())).Returns<ScanResults>(null);
+            _targetElementLocatorMock.Setup(x => x.LocateRootElements(It.IsAny<int>())).Returns(CreateMockElementArray());
```
DaveTryon, line 129: This test is simulates the case where the `TargetElementLocator` returns `null`. That behavior shouldn't change

src/CLI/Resources/OptionsHelpText.resx
```diff
@@ -135,6 +135,9 @@
   <data name="ScanId" xml:space="preserve">
     <value>Scan ID</value>
   </data>
+  <data name="ScanMultipleWindows" xml:space="preserve">
+    <value>Determines if all windows should be scanned (true) or not (false).</value>
```
DaveTryon, line 139: This text feels confusing. We're not scanning all windows, since we've scoped to a single process. Maybe something like "Determines if multiple top-level windows of the target process (true) or just the first one (false)."

src/CLI/OptionsEvaluator.cs
```diff
@@ -55,6 +55,7 @@ public static IOptions ProcessInputs(Options rawInputs, IProcessHelper processHe
                 }
             }
 
+
```
DaveTryon, line 58: Remove extra newline?

src/Automation/SnapshotCommand.cs
```diff
@@ -20,8 +21,9 @@ static class SnapshotCommand
         /// <param name="config">A set of configuration options</param>
         /// <param name="scanTools">A set of tools for writing output files,
         /// creating the expected results format, and finding the target element to scan</param>
+        /// <param name="useMultipleWindows">Determines whether only the first window should be scanned or all windows. Default is <code>true</code></param>
         /// <returns>A SnapshotCommandResult that describes the result of the command</returns>
-        public static ScanResults Execute(Config config, IScanTools scanTools)
+        public static IReadOnlyCollection<ScanResults> Execute(Config config, IScanTools scanTools, bool useMultipleWindows = true)
```
DaveTryon, line 26: Rather than adding an extra parameter to the method, can we add a new property in the `Config`, then populate it based on the `IOptions` ?

src/CLI/IOptions.cs
```diff
@@ -12,5 +12,6 @@ internal interface IOptions
         VerbosityLevel VerbosityLevel { get; }
         int DelayInSeconds { get; }
         string CustomUia { get; }
+        bool ScanMultipleWindows { get; }
```
DaveTryon, line 15: Is it possible to make this name more descriptive, especially since it's publicly exposed? Maybe EnableMultipleScanRoots or something along those lines?

DaveTryon:
Question from above: Should single window scans output the regular file names while multi window scan outputs files with special suffixes or should the behavior be always the same regardless of the number of scanned windows?

We want to leave the current behavior intact for apps with a single top-level window, so 
1. No suffixes
2. Only create the .a11ytest file if errors exist.

Apps with multiple top-level windows are slightly different:
1. Suffixes to distinguish between the various files
2. Always create the .a11ytest files, regardless of whether or not errors exist.

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -343,22 +390,31 @@ public void Execute_WithErrors_CreatesSnapshotAndA11yTestFile()
...
                 .ForProcessId(-1)
                 .WithOutputFileFormat(OutputFileFormat.A11yTest)
                 .Build();
 
             var actualResults = SnapshotCommand.Execute(config, _scanToolsMock.Object);
-            Assert.AreEqual(75, actualResults.ErrorCount);
-            Assert.AreEqual(expectedPath, actualResults.OutputFile.A11yTest);
+            var expectedPaths = new string[]
+            {
+                "Test.File_1_of_3",
```
DaveTryon, line 408: This output seems wrong. I'd expect `Test_1_of_3.File`

DaveTryon (changes requested):
@chingucoding, this is coming together nicely. I added some comments in and tried to answer your open questions

src/Automation/SnapshotCommand.cs
chingucoding: The only reason this new parameter is here is that we know have functions on the IScanner/Scanner that can change this property based on the parameter being passed into the scan method. Should that parameter be ignorable then or should the new `Config.MultipleScanRoots` property be mutable in contrast to the existing properties? I'm inclined to say this should be done based on the config, however I think we should re-evaluate the scanner API again since the properties might not make that much sense. Or we decide that passing the parameter into the Scan function would subsequently change the underlying config so that we can get around introducing a new parameter.

src/AutomationTests/SnapshotCommandUnitTests.cs
chingucoding: Very good point, I think I fixed the test correctly now but of course you are more knowledgeable here.

src/CLI/Resources/OptionsHelpText.resx
chingucoding: Very good point, I've switched to your wording now and updated the parameter name everywhere applicable.

src/CLI/IOptions.cs
chingucoding: Went with EnableMultipleScanRoots, I think it's a very fitting name.

src/AutomationTests/SnapshotCommandUnitTests.cs
chingucoding: Yeah that is just the test mock being wired in such a way that "Test.File" would be the regular file name.

src/Automation/SnapshotCommand.cs
chingucoding: @DaveTryon What are your thoughts on this?

DaveTryon: Sorry, I didn't notice that you were waiting for my input on this.

The current design is for the `Config` object to be immutable, and we should continue that design--I think I may have given you a nudge in the wrong direction previously. I propose the adding a `Builder.WithMultipleScanRootsEnabled()` method, then `ConfigBuilder.Build()` sets the `Config.MultipleScanRootsEnabled` property based on that method being called, and all of the downstream code inspects that new property. That also sets us up to have the default behavior remain unchanged, then we only enable the multiple windows when asked to do so. It's tempting to make the new behavior the default, but I don't think we want to go in that direction.

src/Automation/Interfaces/IScanner.cs
```diff
@@ -39,10 +60,28 @@ public interface IScanner
...
+        /// An exception may be thrown if the value of <see cref="Config.ProcessId"/> is invalid
+        /// or if the directory provided in <see cref="Config.OutputDirectory"/> cannot be created or accessed.
+        /// All exceptions are wrapped in <see cref="AxeWindowsAutomationException"/>.
+        /// If the exception was not thrown by AxeWindows automation, the <see cref="Exception.InnerException"/> property
+        /// will contain the exception.
+        /// If <paramref name="enableMultipleWindows"/> is set to true, all windows will be scanned. If it is set to false, only the first window will be scanned.
+        /// </remarks>
+        /// <returns>Information about the scan and any issues detected</returns>
+        /// <exception cref="AxeWindowsAutomationException"/>
+        IReadOnlyCollection<ScanResults> Scan(string scanId, bool enableMultipleWindows);
```
DaveTryon, line 85: The original logic for making `scanId` a parameter to `IScanner.Scan` was that someone might create a single `IScanner`, then trigger multiple scans using that object. I think we'd be hard pressed to argue that the same test app would trigger some scans with multiple roots enabled and other scans without. For that reason, I don't think we want to have this mehod

src/Automation/Interfaces/IScanner.cs
```diff
@@ -21,11 +22,31 @@ public interface IScanner
...
+        /// An exception may be thrown if the value of <see cref="Config.ProcessId"/> is invalid
+        /// or if the directory provided in <see cref="Config.OutputDirectory"/> cannot be created or accessed.
+        /// All exceptions are wrapped in <see cref="AxeWindowsAutomationException"/>.
+        /// If the exception was not thrown by AxeWindows automation, the <see cref="Exception.InnerException"/> property
+        /// will contain the exception.
+        /// If <paramref name="enableMultipleWindows"/> is set to true, all windows will be scanned. If it is set to false, only the first window will be scanned.
+        /// </remarks>
+        /// <returns>Information about the scan and any issues detected</returns>
+        /// <exception cref="AxeWindowsAutomationException"/>
+        IReadOnlyCollection<ScanResults> Scan(bool enableMultipleWindows);
```
DaveTryon, line 47: I think we only add 1 new method to this interface:
```suggestion
        IReadOnlyCollection<ScanResults> Scan(string scanId);
```

src/Automation/Interfaces/IScanner.cs
```diff
@@ -39,10 +60,28 @@ public interface IScanner
         /// All exceptions are wrapped in <see cref="AxeWindowsAutomationException"/>.
         /// If the exception was not thrown by AxeWindows automation, the <see cref="Exception.InnerException"/> property
         /// will contain the exception.
+        /// Calling this function is equivalent to passing false to <see cref="IScanner.Scan(string, bool)"/> and passing false for <c>enableMultipleWindows</c>.
         /// </remarks>
         /// <returns>Information about the scan and any issues detected</returns>
         /// <exception cref="AxeWindowsAutomationException"/>
         ScanResults Scan(string scanId);
```
DaveTryon, line 67: If someone calls this version with a `Config` that enables multiple roots, we should probably throw an error code telling them to call the other method. Otherwise we don't have a good way to return the data that they need.

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -328,37 +412,88 @@ public void Execute_NullOutputFilePath_ThrowsException()
         }
 
         [TestMethod]
-        [Timeout(1000)]
-        public void Execute_WithErrors_CreatesSnapshotAndA11yTestFile()
+        [Timeout(100000)]
```
DaveTryon, line 415: Wow that's a long timeout! Is it necessary?

src/Automation/SnapshotCommand.cs
chingucoding: I see. I went with the parameter now and removed the optional boolean parameter. Instead, there are two new functions on Scanner that are called `ScanAll()` and `ScanAll(string scanId)`. The `Scan` functions throw an exception if the config has EnableMultipleRoots set to true.

PS: Resolved all other comments that touch the same area so we can have this thread to discuss the general API changes for this.

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -94,54 +98,106 @@ public void Scan_Integration_WindowsFormsControlSampler()
...
                 .WithOutputFileFormat(OutputFileFormat.A11yTest)
+                .WithMultipleScanRootsEnabled()
                 .Build();
+            var scanner = ScannerFactory.CreateScanner(config);
+
+            var action = new Action(() => scanner.Scan("TestIDForThrow"));
+            Assert.ThrowsException<InvalidOperationException>(action);
+        }
+
+        private ScanResults Scan_Integration_Core(string testAppPath, int expectedErrorCount, bool enableMultipleScanRoots = true)
```
DaveTryon, line 154: Should this return `IReadOnlyCollection<ScanResults>` so that upstream tests can validate all roots? Maybe non-multiple root tests call a wrapper that sets enableMultipleScanRoots to false, asserts that only 1 ScanResult was returned, then returns the first element

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -185,20 +182,53 @@ public void Execute_ActionsScan_IsCalledWithExpectedElement()
...
+        public void Execute_ReturnsExpectedResults_MultiRoot()
         {
             _scanToolsMock.Setup(x => x.TargetElementLocator).Returns(_targetElementLocatorMock.Object);
             _scanToolsMock.Setup(x => x.Actions).Returns(_actionsMock.Object);
             _scanToolsMock.Setup(x => x.NativeMethods).Returns(_nativeMethodsMock.Object);
             _scanToolsMock.Setup(x => x.ResultsAssembler).Returns(_resultsAssemblerMock.Object);
+            _scanToolsMock.Setup(x => x.OutputFileHelper).Returns(_outputFileHelperMock.Object);
 
             _targetElementLocatorMock.Setup(x => x.LocateRootElements(It.IsAny<int>())).Returns(CreateMockElementArray());
+            //_outputFileHelperMock.Setup(m => m.EnsureOutputDirectoryExists());
```
DaveTryon, line 221: Should these be removed?

src/Automation/Interfaces/IScanner.cs
```diff
@@ -26,6 +27,23 @@ public interface IScanner
         /// <exception cref="AxeWindowsAutomationException"/>
         ScanResults Scan();
 
+        /// <summary>
+        /// Run AxeWindows automated tests on all windows.
+        /// </summary>
+        /// <remarks>
+        /// If a value was provided in <see cref="Config.OutputDirectory"/>,
+        /// output files may be written to the specified directory.
+        /// (Note: no output files will be written if no errors were found.)
```
DaveTryon, line 36: The comment here is correct for the single-root case, but the behavior is different for multi-root. That seems like it will just increase confusion. Would it be better for ScanAll to always writes the output files, even if it's a single-root case with no errors?

src/Automation/Interfaces/IScanner.cs
```diff
@@ -44,5 +62,21 @@ public interface IScanner
         /// <exception cref="AxeWindowsAutomationException"/>
         ScanResults Scan(string scanId);
 
+        /// <summary>
+        /// Run AxeWindows automated tests on all windows.
+        /// </summary>
+        /// <remarks>
+        /// If a value was provided in <see cref="Config.OutputDirectory"/>,
+        /// output files may be written to the specified directory.
+        /// (Note: no output files will be written if no errors were found.)
```
DaveTryon, line 71: Same comment as previous ScanAll call

src/CLI/Options.cs
```diff
@@ -32,6 +32,9 @@ public class Options : IOptions
         [Option(Required = false, HelpText = "CustomUia", ResourceType = typeof(Resources.OptionsHelpText))]
         public string CustomUia { get; set; }
 
+        [Option(Required = false, HelpText = "EnableMultipleScanRoots", ResourceType = typeof(Resources.OptionsHelpText), Default = true)]
```
DaveTryon, line 35: For back compat, we should default to false

src/CLI/Options.cs
```diff
@@ -32,6 +32,9 @@ public class Options : IOptions
...
+        public bool EnableMultipleScanRoots { get; set; } = true;
```
DaveTryon, line 36: For back compat, these should both default to false

src/Automation/SnapshotCommand.cs
```diff
@@ -35,28 +36,57 @@ public static ScanResults Execute(Config config, IScanTools scanTools)
...
+            List<ScanResults> resultList = new List<ScanResults>();
+
             var rootElements = scanTools.TargetElementLocator.LocateRootElements(config.ProcessId);
 
-            return scanTools.Actions.Scan(rootElements?.First(), (element, elementId) =>
+            if (rootElements is null || !rootElements.Any())
+            {
+                return resultList;
+            }
+
```
DaveTryon, line 47: In the case where an app has multiple roots but doesn't specify the flag, we will fail silently, which seems non-ideal. I'm wondering if we should defend this more aggressively and throw an Exception if a multi-root app is scanned without the multi-app flag. That's technically a breaking change for anyone currently using the `IScanner` interface on a multi-root app, but I think we can defend it since multi-root apps are not common (and those that are using the current tool are failing to report all results). As a side note, that also removes the need for the early exit (lines 57-61)

src/AutomationTests/SnapshotCommandUnitTests.cs
```diff
@@ -185,20 +182,53 @@ public void Execute_ActionsScan_IsCalledWithExpectedElement()
...
+            //_outputFileHelperMock.Setup(x => x.GetNewA11yTestFilePath(It.IsAny<Func<string, string>>())).Returns((Func<string, string> decorator) => decorator("Test.File"));
```
DaveTryon, line 222: Should these lines be removed?

src/CLI/Program.cs
```diff
@@ -19,7 +20,7 @@ class Program
         private readonly IBrowserAbstraction _browserAbstraction;
         private readonly IScanDelay _scanDelay;
 
-        private ScanResults _scanResults;
+        private IReadOnlyCollection<ScanResults> _scanResults;
```
DaveTryon, line 23: For consistent naming, perhaps `_scanResultsCollection` ? This could be a separate PR if that makes sense

src/Automation/SnapshotCommand.cs
```diff
@@ -35,28 +36,57 @@ public static ScanResults Execute(Config config, IScanTools scanTools)
...
+            int targetIndex = 1;
+
+            foreach (var rootElement in rootElements)
             {
-                return ProcessResults(element, elementId, config, scanTools);
-            });
+                resultList.Add(scanTools.Actions.Scan(rootElement, (element, elementId) =>
+                {
+                    return ProcessResults(element, elementId, config, scanTools, targetIndex++, rootElements.Count());
```
DaveTryon, line 54: Hypothetical case: The app has 3 roots, and an Exception is thrown while scanning the last one. The current code would write the output for the 2 successful scans, then the Exception comes much later in output. As a customer, would you prefer to have the "this test failed to complete" message at the top or bottom of the output?

DaveTryon:
@chingucoding, I added some comments now that I'm thinking about this from the point of view of how the various cases could impact someone who is already using the tools in some sort of test environment. I think our first duty is to ensure that their current environment works unchanged in the single-root cases.

src/AutomationTests/AutomationIntegrationTests.cs
chingucoding: Currently, the result from this function is completely unused. I've updated the function to take an additional parameter that verifies how many outputs there are. Moving part of the verification outside the method and inside the actual tests would probably make the tests quite more complex and/or result in duplicate code. But if there is a need for this that with the updated behavior is not covered let me know.

src/Automation/Interfaces/IScanner.cs
chingucoding: That is a very good point. I have updated the behavior and now if EnableMultipleScanRoots option is enabled, we will write the output regardless if there are multiple outputs or not. One thought that just came to mind: What should happen if you don't enable multiple scan roots and still call ScanAll? Should that throw an exception like Scan does if EnableMultipleScanRoots are enabled? Or should the behavior in that case be that output files are written if there were issues with the first window?

src/AutomationTests/SnapshotCommandUnitTests.cs
chingucoding: Fair point, now they are actually used.

src/Automation/SnapshotCommand.cs
chingucoding: This is a tough one. On the one hand, I would argue we should break existing usage as little as possible, however notifying people that they are possibly missing out on accessibility issues is also a fair point. I'm a bit worried that for some folks, calling Scan might get a bit annoying if they have to encounter this exception; I would prefer having people use Scan like they used to regardless if there are multiple windows or not.

src/Automation/SnapshotCommand.cs
chingucoding: I have to be honest that I mostly used Axe.Windows in a interaction test context where I am using the C# API. In that case, I would just want to have the exception be thrown since then I know that something went wrong when it shouldn't. A possible problem I see when interacting with the exceptions (setting them as the first item,etc.) that this could make it possibly harder to see the issue. Given that the exceptions could be because of anything, I think just have the test execution be "natural" is fine.

src/Automation/Data/Config.cs
```diff
@@ -39,6 +39,11 @@ public class Config
         ///  <summary>The path to a file containing configuration instructing Axe Windows how to interpret custom UI Automation data.</summary>
         public string CustomUIAConfigPath { get; private set; }
 
+        /// <summary>
+        /// Determines whether multiple UIA roots will be scanned or just the first one.
+        /// </summary>
+        public bool EnableMultipleScanRoots { get; private set; }
```
DaveTryon, line 45: The name here suggests an action, not a state. Maybe `AreMultipleScanRootsEnabled` or something similar?

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -94,54 +98,107 @@ public void Scan_Integration_WindowsFormsControlSampler()
...
                 .WithOutputFileFormat(OutputFileFormat.A11yTest)
+                .WithMultipleScanRootsEnabled()
                 .Build();
+            var scanner = ScannerFactory.CreateScanner(config);
+
+            var action = new Action(() => scanner.Scan("TestIDForThrow"));
+            Assert.ThrowsException<InvalidOperationException>(action);
+        }
+
+        private ScanResults Scan_Integration_Core(string testAppPath, int expectedErrorCount, bool enableMultipleScanRoots = true, int expectedWindowCount = 1)
```
DaveTryon, line 154: These defaults feel wrong, since the callers each override one value. It seems simpler to make `enableMultipleScanRoots` default to `false`, so that single-root tests override neither value and multi-root tests override both

src/CLI/Options.cs
```diff
@@ -32,6 +32,9 @@ public class Options : IOptions
         [Option(Required = false, HelpText = "CustomUia", ResourceType = typeof(Resources.OptionsHelpText))]
         public string CustomUia { get; set; }
 
+        [Option(Required = false, HelpText = "EnableMultipleScanRoots", ResourceType = typeof(Resources.OptionsHelpText))]
+        public bool EnableMultipleScanRoots { get; set; }
```
DaveTryon, line 36: The name here suggests an action, not a state. Maybe `AreMultipleScanRootsEnabled` or something similar?

src/CLI/Resources/OptionsHelpText.Designer.cs
```diff
@@ -19,7 +19,7 @@ namespace AxeWindowsCLI.Resources {
     // class via a tool like ResGen or Visual Studio.
     // To add or remove a member, edit your .ResX file then rerun ResGen
     // with the /str option, or rebuild your VS project.
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "17.0.0.0")]
```
DaveTryon, line 22: Our officially supported IDE is still VS2019, so maybe we should revert the version change in this line.

src/CLITests/OptionsEvaluatorUnitTests.cs
```diff
@@ -31,14 +31,15 @@ private void VerifyAllMocks()
...
-            VerbosityLevel verbosityLevel = VerbosityLevel.Default, int delayInSeconds = 0)
+            VerbosityLevel verbosityLevel = VerbosityLevel.Default, int delayInSeconds = 0, bool enableMultipleScanRoots = false)
         {
             Assert.AreEqual(processName, options.ProcessName);
             Assert.AreEqual(processId, options.ProcessId);
             Assert.AreEqual(scanId, options.ScanId);
             Assert.AreEqual(outputDirectory, options.OutputDirectory);
             Assert.AreEqual(verbosityLevel, options.VerbosityLevel);
             Assert.AreEqual(delayInSeconds, options.DelayInSeconds);
+            Assert.AreEqual(enableMultipleScanRoots, options.EnableMultipleScanRoots);
```
DaveTryon, line 42: If we rename `IOptions.EnableMultipleScanRoots`, then this parameter should also be renamed to match

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Properties/Resources.Designer.cs
```diff
@@ -0,0 +1,63 @@
...
+    
+    
+    /// <summary>
+    ///   A strongly-typed resource class, for looking up localized strings, etc.
+    /// </summary>
+    // This class was auto-generated by the StronglyTypedResourceBuilder
+    // class via a tool like ResGen or Visual Studio.
+    // To add or remove a member, edit your .ResX file then rerun ResGen
+    // with the /str option, or rebuild your VS project.
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "17.0.0.0")]
```
DaveTryon, line 22: Another VS2019/VS2022 line
```suggestion
    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
```

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Properties/Resources.resx
```diff
@@ -0,0 +1,117 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
DaveTryon, line 1: This was saved from VS2022, while our "standard" is still VS2019. Saving from VS2019 changes the file in unexpected ways. Would you please re-save it from VS2019? When we jump to VS2022, we'll revise all of the files in a single PR.

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Properties/Resources.Designer.cs
DaveTryon, line 22: This was saved from VS2022, while our "standard" is still VS2019. Saving from VS2019 changes the version to 16.0.0.0

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Form1.resx
```diff
@@ -0,0 +1,123 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
DaveTryon, line 1: This was saved from VS2022, while our "standard" is still VS2019. Saving from VS2019 changes the file in unexpected ways. Would you please re-save it from VS2019? When we jump to VS2022, we'll revise all of the files in a single PR.

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Form1.Designer.cs
```diff
@@ -0,0 +1,269 @@
+﻿namespace WindowsFormsMultiWindowSample
```
DaveTryon, line 1: This was saved from VS2022, while our "standard" is still VS2019. Saving from VS2019 changes the file in unexpected ways. Would you please re-save it from VS2019? When we jump to VS2022, we'll revise all of the files in a single PR.

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Properties/Settings.Designer.cs
```diff
@@ -0,0 +1,26 @@
...
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+
+namespace WindowsFormsMultiWindowSample.Properties {
+    
+    
+    [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "17.1.0.0")]
```
DaveTryon, line 15: VS2019 writes this with a different version:
```suggestion
    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "16.10.0.0")]
```

tools/WindowsFormsMultiWindowSample/src/WindowsFormsMultiWindowSample/Properties/Resources.Designer.cs
DaveTryon, line 22: Since our official platform is still VS2019 and to keep things consistent:

```suggestion
    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("System.Resources.Tools.StronglyTypedResourceBuilder", "16.0.0.0")]
```

chingucoding:
I think I addressed all your feedback @DaveTryon , sorry that there are so many VS 2022 changes introduced.

src/Automation/SnapshotCommand.cs
DaveTryon: I discussed this with @RobGallo, and let's keep the backward compatibility. As he put it, "They should know they have windows that aren't being scanned."

DaveTryon:
> I think I addressed all your feedback @DaveTryon , sorry that there are so many VS 2022 changes introduced.

Thanks for persevering! This may have been bigger than any of us quite realized going into this. Approved 😃

DaveTryon (approved):
(Closed by mistake, reopening and approving)

chingucoding:
> Thanks for persevering! This may have been bigger than any of us quite realized going into this. Approved 😃

It's definitely been a journey but I am very happy how the API turned out in the end and I definitely learned a lot about how AXE works! And it was much fun working on this!

PR #685, opened 2022-04-28
chore: Make some analyzer-suggested changes

DaveTryon:
#### Details

When opening the project in a recent version of Visual Studio, the analyzer suggests some changes. These changes were made by accepting the suggested changes.

##### Motivation

Reduce noise when opening project in Visual Studio

#### Pull request checklist

### DIFF
--- a/src/Actions/Contexts/ElementContext.cs
+++ b/src/Actions/Contexts/ElementContext.cs
@@ -39,9 +39,8 @@ namespace Axe.Windows.Actions.Contexts
         /// <param name="element"></param>
         public ElementContext(A11yElement element)
         {
-            if (element == null) throw new ArgumentNullException(nameof(element));
+            this.Element = element ?? throw new ArgumentNullException(nameof(element));
 
-            this.Element = element;
             if (this.Element.PlatformObject == null)
             {
                 this.SelectType = SelectType.Loaded;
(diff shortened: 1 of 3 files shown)


src/Actions/Contexts/ElementContext.cs
```diff
@@ -39,9 +39,8 @@ public class ElementContext : IDisposable
         /// <param name="element"></param>
         public ElementContext(A11yElement element)
         {
-            if (element == null) throw new ArgumentNullException(nameof(element));
+            this.Element = element ?? throw new ArgumentNullException(nameof(element));
```
DaveTryon, line 42: There are lots other places (mainly constructors) where this pattern occurs, but since this is the only one that VS flagged, it's the only one fixed in this PR.

src/Actions/Contexts/ElementContext.cs
```diff
@@ -39,9 +39,8 @@ public class ElementContext : IDisposable
...
 
-            this.Element = element;
```
DaveTryon, line 44: There are lots other places (mainly constructors) where this pattern occurs, but since this is the only one that VS flagged, it's the only one fixed in this PR.

jalkire (approved):

PR #687, opened 2022-05-03
feat(framework-issues-logic): Replace Edge-specific rules with 1 rule that deprecates Edge

DaveTryon:
#### Details

Microsoft has [announced](https://docs.microsoft.com/en-us/lifecycle/announcements/edge-legacy-eos-details) that the non-Chromium version of Edge has reached end of support. We have multiple rules that are sensitive to Edge. This PR combines all of them into a single rule that explains that Edge is no longer a supported product and should not be used.

When reviewing, I recommend going commit by commit, as it helps to expose the onion peeling. The general principles for this rule cleanup are:
1. Assume that [`Framework.Edge`](https://github.com/microsoft/axe-windows/blob/main/src/Rules/PropertyConditions/Framework.cs#L10) always returns [`Condition.False`](https://github.com/microsoft/axe-windows/blob/main/src/Rules/Conditions/Condition.cs#L76)
2. Remove any code that would no longer be run
3. Update tests as needed

##### Motivation

Part of feature work for known framework issues

##### Context

This doesn't include any work to add a link to a page for additional framework information.

#### Pull request checklist


katydecorah (approved):
This pull request appears to clean up and deprecate Edge as described. 

Looks good to me!

@DaveTryon Do you think this pull request would be a good opportunity to also update `docs/RuleDescriptions.md` to reflect these changes? I'm assuming that RuleDescriptions.md is a list of active rules descriptions, but please correct me if I'm wrong! (If yes, it seems like we could remove: `ControlShouldSupportTablePatternInEdge`, `LandmarkOneMain`, and `ControlShouldSupportTablePatternInEdge`; and add: `EdgeBrowserHasBeenDeprecated`.)

DaveTryon:
@katydecorah, it looks like we've been negligent in keeping RulesDescription up to date. I've opened https://github.com/microsoft/axe-windows/pull/695 to bring it up to date before this PR, then I'll revise this PR after that merges to main. That will help to keep the file history a little cleaner

PR #694, opened 2022-05-17
feat(framework-issues-logic): Infrastructure for framework issue links

DaveTryon:
#### Details

We're adding `RuleInfo.FrameworkIssueLink` to identify rules with known framework problems. If a rule sets this field, that serves as an indication that _every_ element that meets the condition of the rule is potentially caused by that known framework issue. The link will go to a documentation page that outlines how to determine if the app is using a framework that has this problem. Where possible, that page will also include links to issues in public GitHub repos, so teams already have an external bug to reference and possibly upvote.

##### Motivation

Needed for framework issue logic feature

##### Context

This PR is purely infrastructure and changes no rules. The unit test simulate rules with known issues. I also tested it locally by adding some links to rules and confirming that they appear in the a11ytest file and in the console output.

#### Pull request checklist

### DIFF
--- a/src/CLI/OutputGenerator.cs
+++ b/src/CLI/OutputGenerator.cs
@@ -153,9 +153,18 @@ namespace AxeWindowsCLI
             foreach (ScanResult scanResult in scanResults.Errors)
             {
                 _writer.WriteLine(DisplayStrings.ScanResultDetailHeader, ++errorCount, scanResult.Rule.Description);
+                WriteFrameworkLink(scanResult);
                 WriteProperties(scanResult);
                 WritePatterns(scanResult);
                 _writer.WriteLine(DisplayStrings.ScanResultDetailFooter);
+            }
+        }
+
+        private void WriteFrameworkLink(ScanResult scanResult)
+        {
+            if (scanResult.Rule.FrameworkIssueLink != null)
+            {
+                _writer.WriteLine(DisplayStrings.FrameworkIssueLink, scanResult.Rule.FrameworkIssueLink);
             }
         }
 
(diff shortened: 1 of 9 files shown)


katydecorah (approved):
@shanisebarona and I pair reviewed and it looks good!

One question:

> I also tested it locally by adding some links to rules and confirming that they appear in the a11ytest file and in the console output.

When you have time, could you share steps on how you did this?

DaveTryon:
> @shanisebarona and I pair reviewed and it looks good!
> 
> One question:
> 
> > I also tested it locally by adding some links to rules and confirming that they appear in the a11ytest file and in the console output.
> 
> When you have time, could you share steps on how you did this?

@katydecorah and @shanisebarona, No problem
1. I ran the CLI against an app (I used notepad++) to find a rule that fails. I chose the `NameNotNull` rule.
2. I added the following inside the ctor for the `NameNotNull` rule:
```
            this.Info.FrameworkIssueLink = "https://www.xyzzy.com";
```
3. I rebuilt the project
4. I ran the CLI in verbose mode against notepad++
```
AxeWindowsCLI.exe --processName notepad++ --scanId test --verbosity verbose
```
5. In the console output, the framework is is shown:
```
Error 3: The Name property of a focusable element must not be null.
  Potential Framework Issue: https://www.xyzzy.com
```
6. An `a11ytest` file is a customized `zip` file, so I renamed `test.a11ytest` (created when running the CLI) to `test.zip` and extracted its contents
7. I opened `el.snapshot` from the extracted contents and searched for `FrameworkIssueLink`. This is null for most of the rules. but is set to "https://www.xyzzy.com" for results that come from this rule (as determined by the matching description).

PR #698, opened 2022-05-18
feat(framework-issues-logic): Remove selected rules

DaveTryon:
#### Details

The spec for this feature calls for the removal of the following rules:
- BoundingRectangleOnUWPMenuBar
- BoundingRectangleOnUWPMenuItem
- BoundingRectangleOnWPFTextParent

The first 3 commits of this PR remove these rules (one rule cleaned up per commit). The 4th commit tidies up obsolete instructions about not removing values from the `RuleId` enum. Those instructions date back to a much earlier version of Axe.Windows, where the file format saved the serialized RuleID to disk, then assumed that the Axe.Windows version in use would still contain that enum value. This changed a long time ago but we never updated the comment. The 4th commit also cleans up some `RuleId` values that had been previously marked as obsolete or deprecated.

##### Motivation

Part of framework issues feature

##### Context

I noticed that some `RuleId` enum values have no corresponding rules. These probably represent cases where the rule was removed but the `RuleId` was left in without being marked as obsolete or deprecated. I left those for a future change.

#### Pull request checklist

- [x] Addresses an existing issue: [1951271](https://mseng.visualstudio.com/1ES/_workitems/edit/1951271)


src/Core/Enums/RuleId.cs
```diff
@@ -40,15 +37,11 @@ public enum RuleId
         ButtonInvokeAndTogglePatterns, // Button should not have Invoke and Toggle patterns together.
         ButtonInvokeAndExpandCollapsePatterns, // Button may have Invoke and ExpandCollapse patterns together. (warning)
         ButtonToggleAndExpandCollapsePatterns, // Button should have have Toggle and ExpandCollapse patterns together.
-        ButtonExpandCollapsePattern,   // obsolete but don't remove since removal will break backward compatibility.
-        ButtonPatterns, // obsolete. keep it to avoid backward compat issue.
```
katydecorah, line 47: Could this pull request also remove: `src\RulesTest\Library\ButtonPatterns.cs`?

src/Core/Enums/RuleId.cs
```diff
@@ -128,9 +121,6 @@ public enum RuleId
         // Axe.Windows.Rules
         EditSupportsIncorrectRangeValuePattern,
 
-        [Obsolete("This rule has been deprecated", true)]
-        EditSupportsOnlyValuePattern,
```
katydecorah, line 135: Is it expected that this rule ID still has corresponding data in: `HowToFix.resx`, `HowToFix.Designer.cs`, `Descriptions.resx`, and `Descriptions.Designer.cs`?

katydecorah (commented):
Overall, this looks good! I left comments regarding more files that could potentially be cleaned up as well.

src/Core/Enums/RuleId.cs
DaveTryon: Long ago, the `ButtonPatterns` rule was split into multiple pieces. The test file should be renamed to `ButtonShouldHavePatterns`, but it's still needed. I've made a note to make a holistic pass through the `RulesTest` project--some tests append "Tests" to the rule name, while others just use the rule name for the test file.

katydecorah (approved):
Looks good to me!

src/Core/Enums/RuleId.cs
DaveTryon: These must have been missed when the rule was removed 😦 

Now I have to wonder how many other "orphans" exist...

PR #699, opened 2022-05-20
chore: Remove unused RuleId values

DaveTryon:
#### Details

This PR removes unused RuleId values, including some that were artifacts of the legacy scan engine that was removed in 2019. It also corrects some comments that refer to the old scan engine. Finally, since all of the RuleId values now deal with `Axe.Windows.Rules`, there is no need to have comments that were originally intended to help to categorize the rules.

##### Motivation

Code hygiene

#### Pull request checklist

### DIFF
--- a/src/Core/Enums/RuleId.cs
+++ b/src/Core/Enums/RuleId.cs
@@ -4,12 +4,7 @@
 namespace Axe.Windows.Core.Enums
 {
     /// <summary>
-    /// Strong rule ids. Previously rules were only identifiable through
-    ///     their string description values
-    /// Rules are grouped by scan via whitespace when relevant
-    /// If writing a ScanException, generally the first rule within a scan grouping
-    ///     is appropriate to use in the constructor of GetRuleResultInstance(...)
-    ///
+    /// Strong rule ids. Each rule must have a unique ID.
     /// Unneeded RuleId values should be removed from this enum when the rule is removed.
     /// </summary>
     public enum RuleId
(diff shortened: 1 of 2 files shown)


src/Core/Enums/RuleId.cs
```diff
@@ -140,30 +119,18 @@ public enum RuleId
...
 
-        // Axe.Windows.Rules
         IsContentElementPropertyExists,
         IsContentElementFalseOptional,
         IsContentElementTrueOptional,
         IsControlElementTrueOptional,
         IsControlElementTrueRequired,
 
-        // Scans
-        IsKeyboardFocusable,
```
shanisebarona, line 163: I searched in the repo for `IsKeyboardFocusable` and saw a few instances of a property by the same name, are these two separate things?

src/Core/Enums/RuleId.cs
```diff
@@ -174,19 +141,10 @@ public enum RuleId
...
-
-        // Axe.Windows.Rules
         ItemTypeRecommended,
 
-        // Axe.Windows.Rules
-        LocalizedControlTypeExists,
         LocalizedControlTypeReasonable,
 
-        // scans
-        NameNonEmpty,
```
shanisebarona, line 187: `NameNonEmpty` is being referenced in a code comment as an example in file `src\Actions\Misc\ExtensionMethods.cs`. Do you think it's worth updating the example?

shanisebarona (approved):
Katy and I reviewed and found a few non-blocking items. Otherwise looks good!

src/Core/Enums/RuleId.cs
DaveTryon: Thanks for catching this. I've updated it to NameNotEmpty, which is a rule that still exists.

src/Core/Enums/RuleId.cs
DaveTryon: Yes, they're different things. UIA exposes the `IsKeyboardFocusable` property, and the `IsKeyboardFocusable` scan was originally intended to check the validity of the property. The scan has since been split into individual rules such as `IsKeyboardFocusableShouldBeTrue`

PR #702, opened 2022-05-20
feat(framework-issues-logic): BoundingRectangleNotNull*

DaveTryon:
#### Details

This PR splits the `BoundingRectangleNotNull` into 3 rules:
1. `BoundingRectangleNotNullListViewXAML` scans on-screen XAML List elements. The docs will eventually reference https://github.com/microsoft/microsoft-ui-xaml/issues/1979 and https://github.com/microsoft/microsoft-ui-xaml/issues/2036
2. `BoundingRectangleNotNullTextBlockXAML` scans on-screen XAML Hyperlink elements hosted in a Text element. The docs will eventually reference https://github.com/microsoft/microsoft-ui-xaml/issues/5666
3. The `BoundingRectangleNotNull` continues as before, although it excludes controls covered by the new rules

The new rule names are geared toward the XAML control names (`ListView` instead of `List` and `TextBlock` instead of `Text`). Failures of `BoundingRectangleNotNullTextBlockXAML` will be reported against the hyperlink element, but the bug actually exists inside the `TextBlock` class, which is why the name references TextBlock. We can change the names if we want--if we do, we should probably also take it into account for the contents of docs.microsoft.com.

##### Motivation

Part of framework issue feature

#### Pull request checklist

- [x] Addresses an existing issue: [1951108](https://mseng.visualstudio.com/1ES/_workitems/edit/1951108) and [1951110](https://mseng.visualstudio.com/1ES/_workitems/edit/1951110)

### DIFF
--- a/src/Rules/Library/BoundingRectangleNotNullListViewXAML.cs
+++ b/src/Rules/Library/BoundingRectangleNotNullListViewXAML.cs
@@ -1,0 +1,37 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
+using Axe.Windows.Rules.PropertyConditions;
+using Axe.Windows.Rules.Resources;
+using static Axe.Windows.Rules.PropertyConditions.BoolProperties;
+using static Axe.Windows.Rules.PropertyConditions.ElementGroups;
+
+namespace Axe.Windows.Rules.Library
+{
+    [RuleInfo(ID = RuleId.BoundingRectangleNotNullListViewXAML)]
+    class BoundingRectangleNotNullListViewXAML : Rule
+    {
+        public BoundingRectangleNotNullListViewXAML()
+        {
+            this.Info.Standard = A11yCriteriaId.ObjectInformation;
+            this.Info.PropertyID = Axe.Windows.Core.Types.PropertyType.UIA_BoundingRectanglePropertyId;
+            this.Info.Description = Descriptions.BoundingRectangleNotNull;
+            this.Info.HowToFix = HowToFix.BoundingRectangleNotNull;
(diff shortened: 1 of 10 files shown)


src/Rules/Library/BoundingRectangleNotNullListViewXAML.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Axe.Windows.Rules.Library
+{
+    [RuleInfo(ID = RuleId.BoundingRectangleNotNullListViewXAML)]
+    class BoundingRectangleNotNullListViewXAML : Rule
+    {
+        public BoundingRectangleNotNullListViewXAML()
+        {
+            this.Info.Standard = A11yCriteriaId.ObjectInformation;
+            this.Info.PropertyID = Axe.Windows.Core.Types.PropertyType.UIA_BoundingRectanglePropertyId;
+            this.Info.Description = Descriptions.BoundingRectangleNotNull;
```
shanisebarona, line 19: Curiosity question: I see that this uses the same description as `BoundingRectangleNotNull`, is this expected?

DaveTryon: Yes. The description is intended to explain what the symptom of the problem is, and since we're still checking the same thing in different frameworks, the same description applies. We've followed this pattern in other cases where we created more targeted versions of an existing rule. It has a downside, in that it violates the concept of a 1:1 relationship between description and rule. As a result, if we receive an issue containing just the description, we also need to look at the framework information to determine which rule is in use. In actual practice, we'd have to do that anyway, so it's not a significant penalty.

shanisebarona (approved):
LGTM!

src/Rules/Library/BoundingRectangleNotNullListViewXAML.cs
shanisebarona: Makes sense, thank you!

PR #705, opened 2022-05-23
feat(framework-issues-logic): Add rule `ControlShouldSupportTextPatternEditWinform`

katydecorah:
#### Details

This rule `ControlShouldSupportTextPatternEditWinform` splits off from [ControlShouldSupportTextPattern](https://github.com/microsoft/axe-windows/blob/main/src/Rules/Library/ControlShouldSupportTextPattern.cs#L40) when Framework is `WinForm` and `ControlType` is `Edit`. 

We found that the `ControlShouldSupportTextPattern` already excludes the conditions (WinForm & Edit) and did not feel the original rule nor its tests required updates to prevent overlap:

https://github.com/microsoft/axe-windows/blob/d269eb404592121dc797f42dc1f694dc932ecd7e/src/Rules/Library/ControlShouldSupportTextPattern.cs#L32-L43

#### Pull request checklist

- [x] Addresses an existing issue: [1951115](https://mseng.visualstudio.com/1ES/_workitems/edit/1951115)

### DIFF
--- a/src/Rules/Library/ControlShouldSupportTextPatternEditWinform.cs
+++ b/src/Rules/Library/ControlShouldSupportTextPatternEditWinform.cs
@@ -5,8 +5,7 @@ using Axe.Windows.Core.Enums;
 using Axe.Windows.Rules.PropertyConditions;
 using Axe.Windows.Rules.Resources;
 using System;
-using static Axe.Windows.Rules.PropertyConditions.ControlType;
-using static Axe.Windows.Rules.PropertyConditions.Framework;
+using static Axe.Windows.Rules.PropertyConditions.ElementGroups;
 
 namespace Axe.Windows.Rules.Library
 {
@@ -31,7 +30,7 @@ namespace Axe.Windows.Rules.Library
 
         protected override Condition CreateCondition()
         {
-            return Edit & WinForms;
+            return WinFormsEdit;
         }
     } // class
 } // namespace
--- a/src/Rules/Library/ControlShouldSupportTextPatternEditWinform.cs
+++ b/src/Rules/Library/ControlShouldSupportTextPatternEditWinform.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Core.Bases;
+using Axe.Windows.Core.Enums;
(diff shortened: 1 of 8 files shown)


src/Rules/Library/ControlShouldSupportTextPatternEditWinform.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Axe.Windows.Rules.Library
+{
+    [RuleInfo(ID = RuleId.ControlShouldSupportTextPatternEditWinform)]
+    class ControlShouldSupportTextPatternEditWinform : Rule
+    {
+        public ControlShouldSupportTextPatternEditWinform()
+        {
+            this.Info.Description = Descriptions.ControlShouldSupportTextPattern;
+            this.Info.HowToFix = HowToFix.ControlShouldSupportTextPattern;
```
katydecorah, line 19: Does it make sense to use the same values from `ControlShouldSupportTextPattern` or would you recommend new values for this rule?

DaveTryon: We've been using the same description and HowToFix for these. When we make the AIWin changes, if both HowToFix and FrameworkIssueLink are set, we'll only display the FrameworkIssueLink. That provides us with downlevel compatibility if an older client opens a newer a11ytest file.

src/Rules/Library/ControlShouldSupportTextPatternEditWinform.cs
```diff
@@ -0,0 +1,37 @@
...
+        public override bool PassesTest(IA11yElement e)
+        {
+            if (e == null) throw new ArgumentNullException(nameof(e));
+
+            return Patterns.Text.Matches(e);
+        }
+
+        protected override Condition CreateCondition()
+        {
+            return Edit & WinForms;
```
DaveTryon, line 34: Can we move this check into a new Condition of `ElementGroups.WinFormsEdit`, then use it from both rules? It's a tiny thing, but if they both point to a shared `Condition`, it's less likely that some future change will accidentally get them out of sync.

katydecorah: Sounds good! Done in https://github.com/microsoft/axe-windows/pull/705/commits/d4c4ae2f8d0ac0760a422225725a369f1792014d

PR #706, opened 2022-05-23
feat(framework-issues-logic): Split IsControlElementTrueRequired into 3 rules

DaveTryon:
#### Details

This breaks the `IsControlElementTrueRequired` rule into 3 rules:
1. `IsControlElementTrueRequiredButtonWPF` is restricted to WPF Buttons. We don't have a GitHub issue for the rule, but the steps are documented at https://docs.microsoft.com/en-us/accessibility-tools-docs/items/wpf/control_iscontrolelement#suggested-fix
2. `IsControlElementTrueRequiredTextInEditXAML` is restricted to XAML Text elements inside an edit control. The docs will eventually point to https://github.com/microsoft/microsoft-ui-xaml/issues/6480, which is the issue tracking this behavior in the XAML framework
3. `IsControlElementTrueRequired` is updated to exclude controls covered by the 2 previous rules.

Unit tests have been added (there were none for `IsControlElementTrueRequired` 😦 ), and the docs have been updated.

##### Motivation

Framework issue feature

##### Context

I changed the 2nd rule name slightly from the spec to better match its behavior. As a result, the link and the docs are slightly different from the rule name. We may want to make a pass through everything at the end of the feature to align everything.

#### Pull request checklist

- [z] Addresses an existing issue: [1951118](https://mseng.visualstudio.com/1ES/_workitems/edit/1951118), [1951119](https://mseng.visualstudio.com/1ES/_workitems/edit/1951119)


src/RulesTest/MonsterTest.cs
```diff
@@ -102,7 +102,7 @@ public void MonsterButtonTest()
             Assert.AreEqual(EvaluationCode.NotApplicable, results[RuleId.IsContentElementTrueOptional]);
             Assert.AreEqual(EvaluationCode.Pass, results[RuleId.IsControlElementPropertyExists]);
             Assert.AreEqual(EvaluationCode.NotApplicable, results[RuleId.IsControlElementTrueOptional]);
-            Assert.AreEqual(EvaluationCode.Pass, results[RuleId.IsControlElementTrueRequired]);
+            Assert.AreEqual(EvaluationCode.NotApplicable, results[RuleId.IsControlElementTrueRequired]);
```
katydecorah, line 105: @DaveTryon Would you mind sharing a little bit about this change?

katydecorah (approved):
Looks good! Left one question to help clarify the change to the MonsterTest.

src/RulesTest/MonsterTest.cs
DaveTryon: Since this is a WPF button, the new rule applies instead of the old one. I'll add a line to capture the new rule

PR #707, opened 2022-05-24
chore: Refactor MonsterTest for better coverage & less churn

DaveTryon:
#### Details

`MonsterTest.cs` is an old test that hasn't really been maintained. It predates the rule-specific unit tests and may no longer be needed. It seems like it's time to either get update it or get rid of it. This is an attempt to update it and streamline future maintenance. It keeps exactly the same input files, but it refactors the tests as follows:
1. To reduce churn and noise, rules are assumed to be `EvaluationCode.NotApplicable` unless explicitly specified. This makes the tests significantly smaller.
2. If new rules are applicable to one of the tests, the test will fail until the expected results have been updated. This prevents new rules from being added without updating the test.
3. Tests will fail if an expected rule is specified more than once in a test.
4. The checks for failures have been moved into shared code.

If we choose to keep the test, we might also want to consider data files for XAML and UWP. That would potentially make the tests more reliable in the build pipeline, where the UI tests sometimes run into trouble. That said, we _want_ to run the CLI on a regular basis, in case we accidentally break something, and the UI tests provide a way to regularly exercise the CLI.

##### Motivation

Old tests were not current. Let's decide how to move forward.

#### Pull request checklist

### DIFF
--- a/src/RulesTest/MonsterTest.cs
+++ b/src/RulesTest/MonsterTest.cs
@@ -20,972 +20,190 @@ namespace Axe.Windows.RulesTests
             A11yElement e = Utility.LoadA11yElementsFromJSON("Snapshots/MonsterButton.snapshot");
             var results = GetTestResultsAsDictionary(e);
 
-            Assert.AreEqual(EvaluationCode.NotApplicable, results[RuleId.BoundingRectangleCompletelyObscuresContainer]);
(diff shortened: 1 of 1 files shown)


peterdur (commented):
LGTM! Messed with expected list, and changes caused tests to fail as expected.

peterdur (approved):

PR #708, opened 2022-05-24
chore: Normalize test names

DaveTryon:
#### Details

Our documented test [standard](https://github.com/microsoft/axe-windows/blob/main/docs/UnitTestBarAndStandards.md#standards) specifies that for a given class `ExplodingWidget`, its unit tests should be in a class named `ExplodingWidgetUnitTests`. In actual practice, we have all of the following variations::
- `ExplodingWidget`
- `ExplodingWidgetTest`
- `ExplodingWidgetTests`
- `ExplodingWidgetUnitTests`

Since the largest section of tests uses the 3rd option (`ExplodingWidgetTests`), that will become our standard. Code changes in this PR were done by renaming files (and in 1 file, renaming classes) using the F2 key in Visual Studio. I also found and fixed a few cases where the test class name was slightly different than the file name.

The following classes used the correct class name but had mismatched file names, so they appear as renames with no edits:
- `ClickablePointOffScreenTests`
- `ClickablePointOnScreenTests`
- `ProgressBarRangeValueTests`
- `SelectionItemPatternSingleSelectionTests`
- `SelectionItemPatternTests`
- `SelectionPatternTests`

##### Motivation

Code consistency, make it easy to distinguish between test classes and production classes. I've already discussed this change with @RobGallo and he's on board with it.

#### Pull request checklist


katydecorah (approved):
The new convention looks good!

Do you think these two files should have a plural `Tests` as well (for the filename and class name)?
- src\RulesTest\Conditions\ConditionTest.cs
- src\RulesTest\Library\Structure\ControlView\ScrollbarTest.cs

DaveTryon:
> Do you think these two files should have a plural `Tests` as well (for the filename and class name)?
> 
> * src\RulesTest\Conditions\ConditionTest.cs
> * src\RulesTest\Library\Structure\ControlView\ScrollbarTest.cs

Yes, I missed these cases. The code in `ScrollbarTest.cs` is testing the `ControlViewScrollbarStructure` class, so part of me wants to make this the `ControlViewScrollbarStructureTests` class. That's really wordy and is inconsistent with the way the rules classes themselves are named, so I've gone with `ScrollbarTests`. I moved it to the correct namespace to minimize confusion and potential collision.

PR #709, opened 2022-05-26
chore:  Add code sample links

DaveTryon:
#### Details

After discussion with @RobGallo, we've decided that the [links.json](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/AccessibilityInsights/Resources/links.json) file from Accessibility Insights for Windows should be a part of Axe.Windows, since there's a much closer affinity and since that's where the code lives that wraps the file. This PR adds the file in its current form and adds some documentation around it. We'll have a corresponding change on the AIWin side to consume the Axe.Windows version of the file.

##### Motivation

Put related code together.

#### Pull request checklist

- [x] Addresses an existing issue: [1951122](https://mseng.visualstudio.com/1ES/_workitems/edit/1951122)


src/Desktop/Utility/StandardLinksHelper.cs
```diff
@@ -25,7 +25,7 @@ public class StandardLinksHelper
         StandardLinksHelper()
         {
             // get the path of dictionary file.
-            var json = File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "links.json"));
+            var json = File.ReadAllText(Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Axe.Windows.CodeSampleLinks.json"));
```
DaveTryon, line 28: Changing the file name here creates a runtime break in AIWin's MSI-based install because we'll need to update [Product.wxs](https://github.com/microsoft/accessibility-insights-windows/blob/main/src/MSI/Product.wxs) to get the new file instead of the old one. That's a 1 line change that we'll make when we consume the updated package.

katydecorah: The rename makes sense to me! 

@DaveTryon Curiousity question for whenever you have a moment: I'm trying to understand is the value of `Assembly.GetExecutingAssembly().Location`. Since we are replacing `links.json` which was imported from Accessibility Insights for Windows, I'm not piecing together why this location remains the same. My best guess is that during build (or related process) `links.json` and now `Axe.Windows.CodeSampleLinks.json` are moved to the location defined by `Assembly.GetExecutingAssembly().Location` -- is that accurate? Just trying to put the pieces together :)

katydecorah (approved):
Looks good to me! I found one small link path fix and added a curiosity question if you have time.

src/Desktop/Utility/StandardLinksHelper.cs
DaveTryon: When AIWin consumes the NuGet package, the contents of the package all move together. Since we put the json file in the same folder as the DLL's, it's in the correct location by default. Please ping me if this doesn't answer your question.

PR #711, opened 2022-05-27
fix: Correct broken link

DaveTryon:
#### Details

Fix broken link. This was a PR comment from #709 but I didn't catch it until after the merge.

#### Pull request checklist

- [n/a] Addresses an existing issue:

PR #712, opened 2022-05-27
fix: Revert sample links

DaveTryon:
#### Details

This PR reverts #709 and #711. The problem was that while the json file containing the links is included in the NuGet package, it doesn't get copied as expected to the final destination. We'll need to figure this out eventually, but since we don't want to gate the release on this, we'll back it out and consider it for a future release.

##### Motivation

Unblock the release.

#### Pull request checklist


katydecorah (approved):

DaveTryon:
This should have gone in before #710 but I botched the order :(. Please note that #709 and #711 were _not_ in release 1.1.4

PR #716, opened 2022-06-08
fix: CLI is throwing an exception, displaying help twice

DaveTryon:
#### Details

The CLI has recently regressed in 2 ways:
1. It is throwing an Exception when no scan is run (for example, just display the help). This occurs as the process exits, so scans complete as expected but the return code indicates failure. This regressed in #682 when a null check was dropped.
2. When displaying the help, it displays twice. This may have come from a recent change in CommandLineParser, but I haven't tried to pin it down.

These are both recent regressions that we didn't catch in the last validation pass.

##### Motivation

CLI usability

##### Context

It would be awesome if the CLI tests could have detected the Exception, but I'm unsure of the best way to add that ability.

#### Pull request checklist

### DIFF
--- a/src/CLI/ReturnValueChooser.cs
+++ b/src/CLI/ReturnValueChooser.cs
@@ -16,9 +16,9 @@ namespace AxeWindowsCLI
         public const int ThirdPartyNoticesDisplayed = 3;  // Used outside this class
         public const int BadInputParameters = 255;
 
-        public static int GetReturnValue(IReadOnlyCollection<ScanResults> scanResults, Exception caughtException)
+        public static int GetReturnValue(bool parserError, IReadOnlyCollection<ScanResults> scanResults, Exception caughtException)
         {
-            if (caughtException as ParameterException != null)
+            if (parserError || caughtException is ParameterException)
                 return BadInputParameters;
 
 #pragma warning disable CA1508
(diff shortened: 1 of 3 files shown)


src/CLI/ReturnValueChooser.cs
```diff
@@ -22,7 +22,7 @@ public static int GetReturnValue(IReadOnlyCollection<ScanResults> scanResults, E
                 return BadInputParameters;
 
 #pragma warning disable CA1508
-            if (caughtException != null || scanResults.Any(result => result == null))
+            if (caughtException != null || scanResults == null || scanResults.Any(result => result == null))
```
peterdur, line 25: If `scanResults == null` we crash before we get to this check, at Program.cs@87, if `options` is not `null`. 
Can reproduce this by running `AxeWindowsCLI.exe --processname`
Adding a guard on `_scanResultsCollection` at Program.cs@85 results in no error message being displayed in this scenario, but the correct value `BadInputParameters` is returned.

DaveTryon: Thanks, @peterdur! I added a line of code to handle this scenario and still generate an appropriate output message

peterdur (approved):
Looks good! Verified error message appears instead of exception in the case noted earlier.

PR #719, opened 2022-06-23
chore: Add SourceLink info to PDB files

DaveTryon:
#### Details

[SourceLink](https://github.com/dotnet/sourcelink) is a mechanism that allows Visual Studio to streamline source-level debugging. It embeds a link from the symbols file (in PDB format) to the repo and commit that were used to create the corresponding assembly. Visual Studio can then use this information to create a local cache of the code at that commit so that the symbols and the code being debugged stay in sync.

This PR enables SourceLink when building the assemblies that are included in the Axe.Windows [NuGet package](https://www.nuget.org/packages/Axe.Windows), and documents the steps needed for consumers of this package to enable the feature inside VisualStudio.

Tested by running these changes through our build pipeline, then debugging from AIWin into Axe.Windows on a machine where cached information about Axe.Windows had been purged. The documentation covers the steps that were needed to get the pieces to work as expected.

##### Motivation

Simplify debugging of Axe.Windows code in other projects (something that I've often wanted in AIWin)

##### Context

This PR has some intentional limitations:
1. It does not publish symbols to any of the Microsoft symbol servers. This was an old security requirement that we dropped some time ago. It _is_ possible to publish symbols to an ADO instance, but this would only benefit people with access to our ADO instance, so the ROI is very small. A future PR could potentially publish symbols to more broadly available symbol servers. We've reached out to the OpenSource experts for their recommended best practices in this area.
2. It does not include symbols for the CLI projects. These packages aren't consumed via the NuGet package, so there's no point in adding SourceLink to these PDB's if the PDB's themselves aren't broadly available.

#### Pull request checklist

### DIFF
--- a/build/NetStandardRelease.targets
+++ b/build/NetStandardRelease.targets
@@ -6,7 +6,14 @@
     <!-- used by Microsoft.CodeAnalysis.NetAnalyzers -->
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
   </PropertyGroup>
-  
+
+  <PropertyGroup>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+  </PropertyGroup>
+
   <Import Project="..\src\props\version.props" />
 
   <ItemGroup>
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -233,6 +233,7 @@ jobs:
       vsVersion: 16.0
       platform: '$(BuildPlatform)'
       configuration: release
+      msbuildArgs: '/p:SourceLinkCreate=true'
 
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@3
     displayName: 'Run BinSkim'
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -233,7 +233,6 @@ jobs:
       vsVersion: 16.0
       platform: '$(BuildPlatform)'
       configuration: release
-      msbuildArgs: '/p:SourceLinkCreate=true'
 
   - task: securedevelopmentteam.vss-secure-development-tools.build-task-binskim.BinSkim@3
     displayName: 'Run BinSkim'
--- a/docs/AutomationReference.md
+++ b/docs/AutomationReference.md
@@ -218,8 +218,8 @@ There may be occasions where you want to debug assemblies within the Axe.Windows
 1. In Visual Studio, go to **Tools** > **Options** > **Debugging** > **General** and uncheck "Enable Just My Code". This enables debugging of code from NuGet packages
 2. In Visual Studio, go to **Tools** > **Options** > **Debugging** > **Symbols** and check "NuGet.org Symbol Server". This tells Visual Studio to attempt to download symbols for NuGet packages. (Note that this download will occur on the first app boot after the setting is enabled or after a version changes, so you may experience a delay in loading the app after making this change)
 3. Run and debug as usual.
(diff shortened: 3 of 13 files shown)


docs/AutomationReference.md
```diff
@@ -212,7 +212,14 @@ example below):
...
+2. In Visual Studio, go to **Tools** > **Options** > **Debugging** > **Symbols** and check "NuGet.org Symbol Server". This tells Visual Studio to attempt to download symbols for NuGet packages. (Note that this download will occur on the first app boot after the setting is enabled or after a version changes, so you may experience a delay in loading the app after making this change)
+3. Run and debug as usual.
+4. For versions 1.1.5 and earlier, you will need to have a local copy of the sources at the same commit as the release (one easy to do this is to download the source code zip file from the corresponding [release](https://github.com/microsoft/axe-windows/releases)). When Visual Studio needs the sources, it will prompt for the location of the source code.
+5. For versions 1.1.6 and later, Visual Studio, can optionally fetch the correct release sources automatically. In Visual Studio, go to **Tools** > **Options** > **Debugging** > **General** and check "Enable Source Link support". When Visual Studio needs the sources, it will ask for permission to retireve them, then it will cache a local copy for future use.
```
RobGallo, line 222:
```suggestion
5. For versions 1.1.6 and later, Visual Studio can optionally fetch the correct release sources automatically. In Visual Studio, go to **Tools** > **Options** > **Debugging** > **General** and check "Enable Source Link support". When Visual Studio needs the sources, it will ask for permission to retrieve them, then it will cache a local copy for future use.
```

build/signedbuild.yml
```diff
@@ -233,6 +233,7 @@ jobs:
       vsVersion: 16.0
       platform: '$(BuildPlatform)'
       configuration: release
+      msbuildArgs: '/p:SourceLinkCreate=true'
```
pharring, line 236: I don't see where this property is used in this PR.
I would expect to see a Condition in one of your .csproj or .targets files referencing it.
If there's no Condition, perhaps it's not necessary to include it.

DaveTryon: This is from the SourceLink docs at https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/symbols?view=azure-devops#set-up-the-build-task. I would suspect that it gets consumed by the SourceLink assemblies, although that's just a theory.

pharring: @tmat FYI

Oh, I see. Thanks. I'd missed that in the docs. I need to chase down where that comes from because, as far as I can tell, it has no effect on the Microsoft.SourceLink.* packages or the rest of the build. I find a few references to it in Blazor and one in dotnet/orleans, but it's not a standard property recognized by the Microsoft.SourceLink NuGets.

For the Microsoft.SourceLink set of NuGet packages, the user-facing MSBuild property is **EnableSourceLink**. You can use it to override the default behavior. See: https://github.com/dotnet/sourcelink/blob/9dd9f81089f875d2452db9d6c8397a8e0b19d688/src/SourceLink.Common/build/Microsoft.SourceLink.Common.props#L19

tmat: I don't know why the docs specify this property. Shouldn't be needed.

build/NetStandardRelease.targets
```diff
@@ -6,7 +6,14 @@
...
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
   </PropertyGroup>
-  
+
+  <PropertyGroup>
+    <PublishRepositoryUrl>true</PublishRepositoryUrl>
+    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <IncludeSymbols>true</IncludeSymbols>
+    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
+  </PropertyGroup>
```
tmat, line 15: I'd add the Source Link package reference here rather then to each project as it is easier to manage.

build/signedbuild.yml
pharring: No, but I entered https://github.com/MicrosoftDocs/azure-devops-docs/issues/12436 to get the docs fixed. I see you responded there already. Thanks.

RobGallo (approved):

Issue #722, opened 2022-06-30
[BUG] Deadlock in TreeWalkerForLive.GetTreeHierarchy


Alexey42:
Deadlock in LINQ AsParallel.ForAll loop when calling TreeWalkerForLive.GetTreeHierarchy with desktop element for the second time. If i use foreach instead of ForAll everything is ok. 

You can reproduce it by calling TreeWalkerForLive.GetTreeHierarchy(A11yAutomation.ElementFromDesktop(), TreeViewMode.Control) twice. 

I expected i can update children this way but it just locks my app. The fact that usual foreach works fine is confusing me.

sfoslund:
Hi @Alexey42, thanks for using axe windows! Can you please provide a little more information on how to reproduce this issue? I've attached a simple console app I used to try to repro but I wasn't able to reproduce the deadlock. If you could provide more repro instructions or a repro application that would be very helpful, thanks!

[DeadlockRepro.zip](https://github.com/microsoft/axe-windows/files/9021273/DeadlockRepro.zip)

Alexey42:
Hi. It should be WPF application and second call is in ui event. Example code (https://github.com/Alexey42/testApp):

public partial class MainWindow : Window
    {
        A11yElement element;
        TreeWalkerForLive tw;

        public MainWindow()
        {
            InitializeComponent();

            element = A11yAutomation.ElementFromDesktop();
            tw = new TreeWalkerForLive();
            tw.GetTreeHierarchy(element, Axe.Windows.Core.Enums.TreeViewMode.Control);
        }

        private void Button_Click(object sender, RoutedEventArgs e)
        {
            tw.GetTreeHierarchy(element, Axe.Windows.Core.Enums.TreeViewMode.Control);
        }
    }

sfoslund:
@Alexey42 thanks for the repro project, I was able to reproduce this behavior with the test app you provided.

PR #724, opened 2022-07-08
chore: Tighten type visibilities

DaveTryon:
#### Details

This PR marks several types that are intended strictly for Axe.Windows internal use as `internal`. This is consistent with the stated intent in the [automation documentation](https://github.com/microsoft/axe-windows/blob/main/docs/AutomationReference.md#axewindows---automation-reference), and will potentially reduce issues like #722, where an internal class was used outside of its intended context.

With the exception of the `Logger` class, the visibilities are changed at the type level. The `Logger` class is static and needs to be accessible by AIWin, so visibilities are changed at the method level. One previously internal method is now private.

Validated by building a nupkg, upgrading a local AIWin to use that package, then building and running AIWin. No problems were found.

##### Motivation

Reduce potential confusion

#### Pull request checklist


src/Desktop/UIAutomation/TreeWalkers/TreeWalkerForLive.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
```
jalkire, line 1: Not sure why GH is flagging a few copyright headers as changed 🤷‍♂️

PR #736, opened 2022-08-11
chore: Apply analyzer-suggested changes based on the .editorconfig file

DaveTryon:
#### Details

We recently added the .editorconfig file to pin down some styles and code recommendations. This is the follow-up PR that applies those styles and recommendations throughout the solution. Going through the process, we pick up 1 change from AIWin [#1402](https://github.com/microsoft/accessibility-insights-windows/pull/1402), so we stay in sync. Most of the changes here were machine-created. The notable exceptions were:
1. The changes to .editorconfig
1. Remove some dead code that the analyzer flagged
1. Add a local `.editorconfig` file inside the `RulesTests` project. This suppresses 2 suggestions inside that project that would have touched virtually every file within the project. We can approach that separately if we want in a future date.

The good news is that we're now analyzer clean, so people know that suggestions are related to new changes.
##### Motivation

Reduce the noise level from analyzer messages, and move towards a point where we can eventually automate code cleanup and styling.

#### Pull request checklist


src/AxeWindows.sln
```diff
@@ -24,6 +24,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Desktop", "Desktop\Desktop.
 	EndProjectSection
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "DesktopTests", "DesktopTests\DesktopTests.csproj", "{AB9E59B6-54CE-4F5D-8055-8AD6745F3B84}"
+	ProjectSection(ProjectDependencies) = postProject
```
DaveTryon, line 27: Build order is InteropDummy, Desktop, then DesktopTests, but the analyzer can't figure that out the transitive dependency. This makes the dependency explicit to satisfy the analyzer, but has no impact on the actual build order.

src/Desktop/Utility/ExtensionMethods.cs
```diff
@@ -114,7 +114,7 @@ public static Bitmap CaptureBitmap(this A11yElement e)
         public static A11yElement GetParentWindow(this A11yElement el)
         {
             A11yElement prev = null;
-            A11yElement curr = el;
+            A11yElement curr = el ?? throw new ArgumentNullException(nameof(el));
```
DaveTryon, line 117: The analyzer pointed out that `el` isn't validated for null. If `el` is null, the following will happen:
1. `curr` gets set to `null` in line 117
2. The `while` loop in line 118 gets skipped
3. A `NullReferenceException` gets thrown in line 124 when we try to read `curr.ControlTypeId`

This throws an `ArgumentNullException` for clarity

src/Actions/Actions/SaveAction.cs
```diff
@@ -24,15 +24,15 @@ public static class SaveAction
         /// <summary>
         /// Names for files inside of zipped snapshot
         /// </summary>
-        public const string elementFileName = "el.snapshot";
-        public const string screenshotFileName = "scshot.png";
-        public const string metadataFileName = "metadata.json";
-        public const string customPropsFileName = "CustomProperties.json";
+        public const string ElementFileName = "el.snapshot";
```
dbjorge, line 27: This is a breaking change, no? Perhaps it would be better to accept a few `[SuppressMessage]` attributes here for the sake of backcompat

dbjorge: (I didn't see any other instances of name changes for public things in the PR)

dbjorge (changes requested):
LGTM modulo the one question in `SaveAction.cs`!

src/Actions/Actions/SaveAction.cs
DaveTryon: These are only intended for `internal` use. Let me update the code accordingly...

DaveTryon: @dbjorge, I've pulled these `const` strings into a separate `internal` class that then gets used by both `LoadAction` and `SaveAction`

dbjorge (approved):

PR #738, opened 2022-08-12
chore: Fix various spelling and typographical errors in source code comments and strings (Core namespace)

codeofdusk:
#### Details
This PR fixes various spelling and typographical errors found in strings and comments in the `Core` namespace.

##### Motivation
Motivated by errors discovered while working on microsoft/accessibility-insights-windows#1323.

##### Context
One of a series of PRs.

#### Pull request checklist

### DIFF
--- a/src/Core/Enums/RuleId.cs
+++ b/src/Core/Enums/RuleId.cs
@@ -28,7 +28,7 @@ namespace Axe.Windows.Core.Enums
         ButtonShouldHavePatterns, // check whether button has at least one of three patterns(Invoke,Toggle,ExpandCollapse)
         ButtonInvokeAndTogglePatterns, // Button should not have Invoke and Toggle patterns together.
         ButtonInvokeAndExpandCollapsePatterns, // Button may have Invoke and ExpandCollapse patterns together. (warning)
-        ButtonToggleAndExpandCollapsePatterns, // Button should have Toggle and ExpandCollapse patterns together.
+        ButtonToggleAndExpandCollapsePatterns, // Button should not have Toggle and ExpandCollapse patterns together.
 
         SiblingUniqueAndFocusable,
         SiblingUniqueAndNotFocusable,
--- a/src/Core/Enums/RuleId.cs
+++ b/src/Core/Enums/RuleId.cs
@@ -28,7 +28,7 @@ namespace Axe.Windows.Core.Enums
         ButtonShouldHavePatterns, // check whether button has at least one of three patterns(Invoke,Toggle,ExpandCollapse)
         ButtonInvokeAndTogglePatterns, // Button should not have Invoke and Toggle patterns together.
         ButtonInvokeAndExpandCollapsePatterns, // Button may have Invoke and ExpandCollapse patterns together. (warning)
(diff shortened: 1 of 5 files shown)


src/Core/Enums/RuleId.cs
```diff
@@ -28,7 +28,7 @@ public enum RuleId
         ButtonShouldHavePatterns, // check whether button has at least one of three patterns(Invoke,Toggle,ExpandCollapse)
         ButtonInvokeAndTogglePatterns, // Button should not have Invoke and Toggle patterns together.
         ButtonInvokeAndExpandCollapsePatterns, // Button may have Invoke and ExpandCollapse patterns together. (warning)
-        ButtonToggleAndExpandCollapsePatterns, // Button should have have Toggle and ExpandCollapse patterns together.
+        ButtonToggleAndExpandCollapsePatterns, // Button should have Toggle and ExpandCollapse patterns together.
```
codeofdusk, line 31: Should this be **should not**?

```suggestion
        ButtonToggleAndExpandCollapsePatterns, // Button should not have Toggle and ExpandCollapse patterns together.
```

src/Core/Misc/BoundedCounter.cs
```diff
@@ -49,7 +49,7 @@ public BoundedCounter(int upperBound)
         /// <summary>
         /// Attempt to increment Count
         /// </summary>
-        /// <returns>true iff we can increment without exceeding UpperBound</returns>
+        /// <returns>true if and only if we can increment without exceeding UpperBound</returns>
```
codeofdusk, line 52: Is [iff](https://en.wikipedia.org/wiki/If_and_only_if) the intent here? If so, I think expanding it out is clearer for TTS users at least...

src/Core/Types/TypeBase.cs
```diff
@@ -64,7 +64,7 @@ protected virtual string GetNameInProperFormat(string name, int id)
         }
 
         /// <summary>
-        /// check whether the contant should be part of List
+        /// check whether the content should be part of List
```
codeofdusk, line 67: Is this content or constant?

src/Core/Enums/RuleId.cs
DaveTryon: "should not" is correct. The user-displayed message for this is "A button must not support both the Toggle and ExpandCollapse patterns"

src/Core/Misc/BoundedCounter.cs
DaveTryon: Reasonable. There are also 3 instances of `iff` in the Win32 project

src/Core/Types/TypeBase.cs
DaveTryon: Content seems to make more sense. The id is just a key into a collection

DaveTryon (approved):

src/Core/Misc/BoundedCounter.cs
codeofdusk: There are indeed, just splitting up the changes by namespace to ease review (see follow-up #741). Thanks for the quick reviews!

Issue #730, opened 2022-07-30
[BUG] Scanning UWP apps crashes since UIAutomation does not find any child elements


chingucoding:
## Describe the bug

When scanning a UWP app, calling `Scan(all)` crashes with an exception originating from TargetElementLocator.cs:

```
Axe.Windows.Automation.AxeWindowsAutomationException: ERROR Automation003: Unhandled Exception System.ArgumentException: The argument must be a non-empty IEnumerable (Parameter 'desktopElements')
   at Axe.Windows.Automation.TargetElementLocator.GetA11YElementsFromDesktopElements(IEnumerable`1 desktopElements)+MoveNext() in D:\a\_work\1\s\src\Automation\TargetElementLocator.cs:line 31
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source)
   at Axe.Windows.Automation.SnapshotCommand.Execute(Config config, IScanTools scanTools) in D:\a\_work\1\s\src\Automation\SnapshotCommand.cs:line 43
   at Axe.Windows.Automation.Scanner.<>c__DisplayClass7_0.<ExecuteScan>b__0() in D:\a\_work\1\s\src\Automation\Scanner.cs:line 78
   at Axe.Windows.Automation.ExecutionWrapper.ExecuteCommand[T](Func`1 command) in D:\a\_work\1\s\src\Automation\ExecutionWrapper.cs:line 23 ---> System.ArgumentException: The argument must be a non-empty IEnumerable (Parameter 'desktopElements')
```

The issue originates from the fact that calling `UIAutomation.FindAll` in `A11yAutomation.cs`, line 69, returns no elements. This is a regression introduced by #676 since behavior was changed to filter elements using `TreeScope.TreeScope_Children` instead of the previous scope `TreeScope.TreeScope_Descendants`. This regression only seems to occur with UWP apps, WPF, Winform and WinUI 3 Desktop apps do not seem to exhibit this behavior and can be scanned just fine.

**To Reproduce**

1. Download WinUI 2 Gallery from Store
2. Run the following test:
```
var processes = Process.GetProcessesByName("AppUIBasics");
Assert.IsTrue(processes.Length > 0);

var config = Config.Builder.ForProcessId(processes[0].Id).WithMultipleScanRootsEnabled().Build();
var scanner = ScannerFactory.CreateScanner(config);
var testResult = scanner.ScanAll();
```

## Expected behavior
The scan should succeed normally.

## Actual behavior

[...]

PR #744, opened 2022-08-16
fix: Fix process UIA scanning to also work on UWP apps by walking UIA tree manually

chingucoding:
#### Details
As outlined in #730 , scanning UWP apps was broken when we switched how we scan children nodes in #676. This PR based on https://github.com/DaveTryon/axe-windows/pull/45 fixes this by manually walking the UIA tree. In my personal testing, this change did not seem to create a big performance difference (if one at all) when running the interaction tests.

##### Motivation
Addresses issue #730 

##### Context

Ideally, we would want a test with a UWP being scanned, however due to the complexity of installing the app on the remote machine and then running it, I'm fairly sure this would not be that easy. This issue of testing a UWP app was also discussed in another issue (can't remember which :/) with the same outcome.

#### Pull request checklist

- [x] Addresses an existing issue: #730

### DIFF
--- a/src/Desktop/UIAutomation/A11yAutomation.cs
+++ b/src/Desktop/UIAutomation/A11yAutomation.cs
@@ -72,7 +72,6 @@ namespace Axe.Windows.Desktop.UIAutomation
             {
                 if (FindProcessMatchingChildren(root, walker, pid, matchingElements, nonMatchingElements))
                 {
-                    ReleaseElements(walker, nonMatchingElements);
                     return matchingElements;
                 }
 
@@ -80,7 +79,6 @@ namespace Axe.Windows.Desktop.UIAutomation
                 {
                     if (FindProcessMatchingChildren(nonMatchingElement, walker, pid, matchingElements, nonMatchingElementsSecondLevel))
                     {
-                        ReleaseElements(walker, nonMatchingElements, nonMatchingElementsSecondLevel);
                         return matchingElements;
                     }
                 }
--- a/src/Desktop/UIAutomation/A11yAutomation.cs
+++ b/src/Desktop/UIAutomation/A11yAutomation.cs
@@ -1,9 +1,8 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
 using Axe.Windows.Core.Misc;
-using Axe.Windows.Core.Types;
 using Axe.Windows.Desktop.UIAutomation.TreeWalkers;
 using Axe.Windows.Telemetry;
 using Axe.Windows.Win32;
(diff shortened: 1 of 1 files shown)


src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,63 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+            List<IUIAutomationElement> nonMatchingElements = new List<IUIAutomationElement>();
+            List<IUIAutomationElement> nonMatchingElementsSecondLevel = new List<IUIAutomationElement>();
+
+            if (FindProcessMatchingChildren(root, walker, pid, matchingElements, nonMatchingElements))
+            {
+                CleanUpNonMatchingElements();
+                return matchingElements;
+            }
+
+            foreach (var nonMatchingElement in nonMatchingElements)
```
DaveTryon, line 78: @chingucoding, can we move the call to `CleanUpNonMatchingElements` into a finally block, so that it also gets triggered if an exception gets thrown?

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,63 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+                if (FindProcessMatchingChildren(nonMatchingElement, walker, pid, matchingElements, nonMatchingElementsSecondLevel))
+                {
+                    CleanUpNonMatchingElements();
+                    return matchingElements;
+                }
+            }
+            CleanUpNonMatchingElements();
+            return matchingElements; // Will always be empty
+
+            void CleanUpNonMatchingElements()
```
DaveTryon, line 89: We don't use nested functions anywhere else in the code, and I believe that we want to keep it that way. Can we please refactor this?

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,63 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+            {
+                foreach (var nonMatchingElementToBeCleaned in nonMatchingElements)
```
DaveTryon, line 91: name suggestion: `nonMatchingElement`

DaveTryon, line 91: name suggestion: `nonMatchingElement`

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,63 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+                {
+                    Marshal.ReleaseComObject(nonMatchingElementToBeCleaned);
+                }
+                nonMatchingElements = null;
+                foreach (var nonMatchingElementToBeCleaned in nonMatchingElementsSecondLevel)
```
DaveTryon, line 96: name suggestion: either `nonMatchingElement` or `nonMatchingElementSecondLevel`

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,63 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+                {
+                    Marshal.ReleaseComObject(nonMatchingElementToBeCleaned);
+                }
+                nonMatchingElementsSecondLevel = null;
```
DaveTryon, line 100: There's probably no need to set the lists to null, especially if we drop the nested function

DaveTryon (changes requested):
@chingucoding, please see my comments, especially the one about removing the nested method.

src/Desktop/UIAutomation/A11yAutomation.cs
chingucoding: Sure thing, moved this into a static function.

src/Desktop/UIAutomation/A11yAutomation.cs
chingucoding: Good idea, done!

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,65 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+        {
+            IUIAutomationTreeWalker walker = GetTreeWalker(TreeViewMode.Control);
+            List<IUIAutomationElement> matchingElements = new List<IUIAutomationElement>();
+            List<IUIAutomationElement> nonMatchingElements = new List<IUIAutomationElement>();
+            List<IUIAutomationElement> nonMatchingElementsSecondLevel = new List<IUIAutomationElement>();
+            try
+            {
+                if (FindProcessMatchingChildren(root, walker, pid, matchingElements, nonMatchingElements))
+                {
+                    ReleaseElements(walker, nonMatchingElements);
```
DaveTryon, line 75: This will get handled by the finally block
```suggestion
```

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -46,6 +45,65 @@ public static DesktopElement ElementFromProcessId(int pid)
...
+                    return matchingElements;
+                }
+
+                foreach (var nonMatchingElement in nonMatchingElements)
+                {
+                    if (FindProcessMatchingChildren(nonMatchingElement, walker, pid, matchingElements, nonMatchingElementsSecondLevel))
+                    {
+                        ReleaseElements(walker, nonMatchingElements, nonMatchingElementsSecondLevel);
```
DaveTryon, line 83: This will get handled by the finally block
```suggestion
```

DaveTryon (approved):
Thanks, @chingucoding! I'll merge this once the PR succeeds (it's being a little bit grumpy today)

PR #747, opened 2022-08-19
chore: Refresh README that installs with CLI

DaveTryon:
#### Details

I noticed that the CLI's readme.md file was not as complete as it ought to be (it was last updated with version 1.0.8). Changes included:
- Include the help output from the current version (line breaks are different because of the presence of the `areMultipleScanRootsEnabled` parameter
- Bump the version so that it's in sync with the build version (see note under context)
- Remove superfluous `|` characters at start and stop of lines in the table that documents command line parameters
- Add documentation for the `areMultipleScanRootsEnabled` option
- Expand the explanation of the `customUia` note

##### Motivation

The readme that ships in the product should be in sync with what the user encounters when running the product.

##### Context

Q: Why update the build version to be in sync with `version.props`?
A: Ideally, the version in the readme should stay in sync with the version in `version.props`. In the future there should never be an edit to `version.props` without a corresponding edit to the readme. There may be cases (new options or improved documentation, for example) where the readme changes independently of the `version.props` file.

#### Pull request checklist

### DIFF
--- a/src/CLI/README.MD
+++ b/src/CLI/README.MD
@@ -16,7 +16,7 @@ You invoke the scanner by running `AxeWindowsCLI.exe`. By default, this tool get
 If you run the tool without parameters, you'll be presented with the help screen. An example follows:
 
 ```
-AxeWindowsCLI 1.1.7
+AxeWindowsCLI 1.1.8
 Copyright c 2020
 
   --processid                      Process Id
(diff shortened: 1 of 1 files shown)


codeofdusk (approved):
Custom UIA change looks reasonable.

jalkire (approved):

Issue #733, opened 2022-08-09
[Feature Request] NativeMethods.SetProcessDPIAware() should be reset after scan


chabiss:
## Describe the bug
The call to SetProcessDPIAware()
https://github.com/microsoft/axe-windows/blob/4a09d8728ed72fa9d307052a52a1dd4d9cee1258/src/Automation/SnapshotCommand.cs

Will affect any host application integrating axe-window into their product and it's not undone once the scan is completed.  Either remove the call or bring the host application state to where it was before.

spadapet:
I think a good alternative to SetProcessDPIAware() is SetThreadDpiAwarenessContext(), which just changes the current thread and also allows you to switch the thread's awareness back to the way it was:
https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setthreaddpiawarenesscontext

ferBonnin:
Thanks for the feature request, @chabiss, marking this as needs investigation to analyze on our side.

DaveTryon:
This has been merged into main and will be included in an upcoming release of Axe.Windows. We'll close it out once we release the updated package to GitHub

DaveTryon:
@chabiss, This is included in version 2.0.0. Release details are at https://github.com/microsoft/axe-windows/releases/tag/v2.0.0. Closing this issue

PR #751, opened 2022-09-09
feat: Allow callers to override default DPI awareness handling

DaveTryon:
#### Details

We need to be DPI-aware when running an Axe.Windows scan. We currently call SetProcessDPIAware then never reset the state, partially because there's no way to obtain the original state and partially because we never needed to revert to a non-DPI aware state. #733 asks for a way to scope this and/or reset the original state. We want to enable this, but since different scenarios could require different handling, we are moving the DPI awareness into an interface and allowing the caller to provide their own implementation. The new interface is defined as follows:

```
    public interface IDPIAwareness
    {
        /// <summary>
        /// Enable DPI awareness for the scan
        /// </summary>
        /// <returns>An object that will be passed as a parameter <see cref="IDPIAwareness.Restore"/></returns>
        object Enable();

        /// <summary>
        /// Restore DPI awareness to its original state
        /// </summary>
        /// <param name="dataFromEnable">The object returned from the call to <see cref="IDPIAwareness.Enable"/></param>
        void Restore(object dataFromEnable);
    }
```

Calls to `Enable` and `Restore` will always be paired. When we add async support, we'll ensure that both calls will occur on the thread where the scanning will occur. We'll need to work out the details for ScanAllAsync, but if we set this up to have the scans run concurrently, then a thread will be created for each scan and each scan will have a pair of `Enable`/`Restore` calls.

A user's implementation of `IDPIAwareness` is injected by calling `Config.Builder.WithDPIAwareness` when setting up the scanner. Apps that don't call `ConfigBuilder.WithDPIAwareness` will continue to use the existing behavior, so backward compatibility is preserved.

##### Motivation

Address #733 

##### Context

We _could_ have just set the DPI awareness on the thread, then reset it as suggested in a comment on #733. That would potentially break backward compatibility, and it opens the door of possibly trying to provide one-off solutions as customer needs change. The simpler long-term option is to allow customers to implement what works for their scenario. This may become even more important as we think about adding async versions of the scanning methods. Making assumptions about the calling code is a great way to create a bug pump.

#### Pull request checklist


DaveTryon:
@chabiss and @spadapet, please let us know if this approach will unblock the scenarios that you have in mind. Thanks!

src/Automation/SnapshotCommand.cs
```diff
@@ -28,40 +28,46 @@ public static IReadOnlyCollection<ScanResults> Execute(Config config, IScanTools
...
-
-            if (config.CustomUIAConfigPath != null)
-                scanTools.Actions.RegisterCustomUIAPropertiesFromConfig(config.CustomUIAConfigPath);
-
-            List<ScanResults> resultList = new List<ScanResults>();
-
-            var rootElements = scanTools.TargetElementLocator.LocateRootElements(config.ProcessId);
-
-            if (rootElements is null || !rootElements.Any())
+            object dpiAwarenessObject = scanTools.DpiAwareness.Enable();
```
spadapet, line 34: If the call below to scanTools.Actions.Scan() will create parallel tasks, I wonder if IDPIAwareness.Enable() needs to be called in each task since they run in their own threads. That depends where window coordinates are used. My implementation of IDPIAwareness would set the current thread's DPI awareness.

DaveTryon: I hadn't thought about the multi-threaded case, but it makes total sense. I'll move the DPI calls closer to the code that will eventually be moved into the ThreadProc when async calls are added

spadapet: Thanks!

spadapet (approved):

src/Automation/Data/Config.cs
```diff
@@ -84,6 +93,17 @@ public Builder WithOutputDirectory(string directory)
                 return this;
             }
 
+            /// <summary>
+            /// Specifies a custom mechansm to
```
sfoslund, line 97: nit: `mechansm` -> `mechanism`, and I think the rest of this comment got cut off.

DaveTryon: Fixed now :)

karanbirsingh (approved):

PR #754, opened 2022-10-05
feat(async-api): Improve DataManager

DaveTryon:
#### Details

This improves the `DataManager` class as follows as we prepare for an async API:
1. `DataManager.GetDefaultInstance` used a non-concurrent `Dictionary` to allow for side-by-side instances of `DataManager`, even though the code _never_ created anything but the default instance. This removes the `Dictionary`. Future changes will create separate `DataManager` objects that will be explicitly passed into API calls.
2. `DataManager.Dispose` was never implemented beyond the boilerplate, so it was essentially a NOP. The `Dispose` method will be a important part of our data cleanup in the async world, so we need a proper implementation. We'll be cleaning up UIA objects, so the finalizer is needed (the boilerplate added it commented)
3. `DataManager.RemoveDataContext` had a documented parameter whose code had been removed. This PR bring back this parameter since we'll use it in the `Dispose` method
4. `DataManager.Clear` has been renamed to `DataManager.ClearDefaultInstance`, since that's what it does
5. `DataManager.GetDefaultInstance` and `DataManager.ClearDefaultInstance` were not thread safe, so I added locking using the check/lock/check pattern. This confuses the analyzer, so suppress the CA1508 warning
6. Add `DataManager.CreateInstance` for creating a new `DataManager` object. We could just expose the constructor, but other changes in this feature will also use the `CreateInstance` pattern, so this sets us up for consistency

##### Motivation

Part of [1996306](https://mseng.visualstudio.com/1ES/_workitems/edit/1996306) to enable async API calls not to collide

##### Context

This is the first of a series of PR's. I'm trying to keep them as manageable as possible and this was an easily detachable piece.

#### Pull request checklist

- [x] Contributes to [1996306](https://mseng.visualstudio.com/1ES/_workitems/edit/1996306)

### DIFF
--- a/src/Actions/Actions/DataManager.cs
+++ b/src/Actions/Actions/DataManager.cs
@@ -1,5 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using Axe.Windows.Actions.Attributes;
 using Axe.Windows.Actions.Contexts;
 using Axe.Windows.Actions.Enums;
@@ -32,6 +33,8 @@ namespace Axe.Windows.Actions
         /// keep the record of all element context added by AddElementContext
         /// </summary>
         readonly Dictionary<Guid, ElementContext> ElementContexts = new Dictionary<Guid, ElementContext>();
+
+        static readonly Object LockObject = new Object();
 
         /// <summary>
         /// Get A11yPattern from an indicated element/elementcontext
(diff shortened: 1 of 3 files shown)


src/Actions/Actions/DataManager.cs
```diff
@@ -228,18 +228,25 @@ protected virtual void Dispose(bool disposing)
             {
                 if (disposing)
                 {
-                    // TODO: dispose managed state (managed objects).
+                    // We need an immutble copy of the ecId values for cleanup
```
sfoslund, line 231: nit: immutble -> immutable

sfoslund (approved):

src/Actions/Actions/DataManager.cs
DaveTryon: Thanks for catching this! Fixed

PR #755, opened 2022-10-06
feat(async-api): Add cancellation token for TreeWalker and RuleRunner

sfoslund:
#### Details

Adding cancellation logic to two of the most time intensive operations performed during scanning, the tree walker and rule runner. 

Based on docs/ discussion online, it appears that the norm is to use `ThrowIfCancellationRequested` to check the cancellation token throughout the async call. This throws a `OperationCanceledException`, which will be wrapped in an `AggregateException` once these methods are wrapped in a task.

Also added bare minimum test coverage for cancellation in TreeWalkerForTest, which previously did not have any unit tests. 

##### Motivation

Part of [1986365](https://mseng.visualstudio.com/1ES/_workitems/edit/1986365) to enable cancellation

##### Context

This is the first in what will be multiple PRs to add cancellation tokens. In the future we will need to add the logic to pass the cancellation token to the methods changed in this PR, but that is blocked until we add a top level async scan API. .

#### Pull request checklist

- [x] Contributes to [1986365](https://mseng.visualstudio.com/1ES/_workitems/edit/1986365)

### DIFF
--- a/src/Actions/Actions/CaptureAction.cs
+++ b/src/Actions/Actions/CaptureAction.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Axe.Windows.Actions.Attributes;
@@ -12,6 +12,7 @@ using Axe.Windows.Desktop.UIAutomation.TreeWalkers;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace Axe.Windows.Actions
 {
@@ -117,12 +118,13 @@ namespace Axe.Windows.Actions
         {
             dc.TreeMode = tm;
             dc.Mode = dcMode;
+            var cancellationToken = CancellationToken.None; // In the future this will be passed in from the top level API
 
             switch (dcMode)
             {
                 case DataContextMode.Test:
                     var stw = NewTreeWalkerForTest(dc.Element, dc.ElementCounter);
-                    stw.RefreshTreeData(tm);
+                    stw.RefreshTreeData(tm, cancellationToken);
                     dc.Elements = stw.Elements.ToDictionary(l => l.UniqueId);
                     dc.RootElment = stw.TopMostElement;
                     break;
--- a/src/Actions/Actions/CaptureAction.cs
+++ b/src/Actions/Actions/CaptureAction.cs
@@ -118,13 +118,13 @@ namespace Axe.Windows.Actions
         {
             dc.TreeMode = tm;
             dc.Mode = dcMode;
-            var cancellationToken = CancellationToken.None; // In the future this will be passed in from the top level API
+            var dataContext = new TreeWalkerDataContext();
 
             switch (dcMode)
             {
                 case DataContextMode.Test:
                     var stw = NewTreeWalkerForTest(dc.Element, dc.ElementCounter);
-                    stw.RefreshTreeData(tm, cancellationToken);
+                    stw.RefreshTreeData(tm, dataContext);
                     dc.Elements = stw.Elements.ToDictionary(l => l.UniqueId);
                     dc.RootElment = stw.TopMostElement;
                     break;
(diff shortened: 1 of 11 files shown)


src/Desktop/UIAutomation/TreeWalkers/TreeWalkerForTest.cs
```diff
@@ -57,15 +58,17 @@ public TreeWalkerForTest(A11yElement element, BoundedCounter elementCounter)
         /// Refresh tree node data with all children at once.
         /// <param name="mode">indicate the mode</param>
         /// </summary>
-        public void RefreshTreeData(TreeViewMode mode)
+        public void RefreshTreeData(TreeViewMode mode, CancellationToken? cancellationToken = null)
```
DaveTryon, line 61: How do you feel about using `CancellationToken.None` instead of a nullable object?

sfoslund: Good idea! I'll make that change.

src/DesktopTests/UIAutomation/TreeWalkerForTestTests.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
DaveTryon, line 1: For consistency with the production code, can we move this into the DesktopTests/UIAutomation/TreeWalkers folder and the Axe.Windows.DesktopTests.UIAutomation.TreeWalkers namespace?

src/DesktopTests/UIAutomation/TreeWalkerForTestTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        [TestMethod]
+        [Timeout(1000)]
+        public void RefreshTreeData_ThrowsWhenCancelled()
+        {
+            var cancellationToken = new CancellationTokenSource();
+            cancellationToken.Cancel();
+            var dataContext = new TreeWalkerDataContext(cancellationToken.Token);
+            var exception = Assert.ThrowsException<OperationCanceledException>(() => _treeWalker.RefreshTreeData(TreeViewMode.Raw, dataContext));
```
DaveTryon, line 38: We can simplify this if we aren't actually checking anything on the Exception:
```suggestion
            Assert.ThrowsException<OperationCanceledException>(() => _treeWalker.RefreshTreeData(TreeViewMode.Raw, dataContext));
```

src/Actions/Actions/CaptureAction.cs
```diff
@@ -117,12 +118,13 @@ internal static void PopulateData(ElementDataContext dc, DataContextMode dcMode,
         {
             dc.TreeMode = tm;
             dc.Mode = dcMode;
+            var dataContext = new TreeWalkerDataContext();
```
DaveTryon, line 121: Maybe a TODO to help the future integration?
```suggestion
            var dataContext = new TreeWalkerDataContext(); // TODO: Initialize from parameters
```

src/RulesTest/RuleRunnerTests.cs
```diff
@@ -343,6 +345,20 @@ public void RunAll_ReturnsExpectedResults()
...
+        {
+            var e = new MockA11yElement();
+            var ruleMock = new Mock<IRule>(MockBehavior.Strict);
+            var providerMock = new Mock<IRuleProvider>(MockBehavior.Strict);
+            providerMock.Setup(m => m.All).Returns(() => new IRule[] { ruleMock.Object }).Verifiable();
+
+            var runner = new RuleRunner(providerMock.Object);
+            var cancellationToken = new CancellationTokenSource();
+            cancellationToken.Cancel();
+            var exception = Assert.ThrowsException<OperationCanceledException>(() => runner.RunAll(e, cancellationToken.Token));
```
DaveTryon, line 359: We can simplify this if we don't need to use the exception:
```suggestion
            Assert.ThrowsException<OperationCanceledException>(() => runner.RunAll(e, cancellationToken.Token));
```

DaveTryon (approved):

DaveTryon (approved):
LGTM

Issue #732, opened 2022-08-09
[Feature Request] Need Task<ScanResults> IScanner.ScanAsync()


chabiss:
## Is your feature request related to a problem? Please describe.
Currently the call to ScanResults IScanner.Scan() will result in a blocking call that will blocked the host making the call and bloking the UI thread, which is not desirable. 

## Describe the solution you'd like
We need Task<ScanResults> IScanner.ScanAsync() which allows the scanner to run asynchronously in a background thread

## Describe alternatives you've considered
No other alternative considered. 

## Additional context
N/A

DaveTryon:
In offline conversations, having proper cancellation is part of the ask, even though it's not called out in the description.

ferBonnin:
Thanks for the feature request, @chabiss, marking this as needs investigation to analyze on our side.

DaveTryon:
We're actively working on this feature

DaveTryon:
@chabiss, This is included in version 2.0.0. Release details are at https://github.com/microsoft/axe-windows/releases/tag/v2.0.0. Closing this issue

PR #756, opened 2022-10-07
feat(async-api): Introduce the new Axe.Windows 2.0 API

codeofdusk:
##### Summary of the issue
* In #732, @chabiss requested the ability to run scans from asynchronous code with support for cancellation.
* Adding additional options to tune an individual scan becomes difficult, as new methods (`Scan`, `ScanAll`, and so on) and/or method overloads (the aforementioned methods with and without `scanId`) are required to avoid breaking the API.
* Returning additional data about an entire scan (i.e. not just additional data about the results of scanning one top-level window) is impossible without breaking the API.

###### new `IScanner` interface
The current `IScanner` interface has been replaced with a new design. The below methods accept one argument, a `ScanOptions` instance or `null` for default behaviour (see below for details) and return a `ScanOutput` (see below for details).

Method | Description
--- | ---
`ScanAsync` | Run a scan asynchronously. This method should be `await`ed.
`Scan` | Run a scan synchronously. This method blocks until the scan is complete.

###### `ScanOptions` class
Calls to methods on `IScanner` take a `ScanOptions` instance as argument. This allows for new options to be added without changing the `IScanner` interface or breaking calling code. Passing `null` in place of a `ScanOptions` instance will start a scan with all options set to default. At the moment, the `ScanOptions` class only has a `ScanId` field which works as before.

###### `ScanOutput` class
`ScanOutput` is the return type of the `IScanner` methods. Similar to `ScanOptions`, it allows new fields to be added without breaking calling code. At the moment, it contains one field: `WindowScanOutputs`, an `IReadOnlyCollection` of `WindowScanOutput` objects. `WindowScanOutput` is a new name for the `ScanResults` class.

Note that, even when multi-root scanning is disabled in config, `ScanOutput.WindowScanOutputs` remains a collection (simply containing one `WindowScanOutput` object).

###### Exceptions
`IScanner` methods no longer wrap all system-generated exceptions in `AxeWindowsAutomationException` to reduce indirection (both for asynchronous code, which wraps exceptions in `AggregateException`, and to prevent some fatal exceptions from being inadvertantly hidden). `AxeWindowsAutomationException` remains only for cases in which Axe.Windows can add meaningful error reporting.

##### Testing strategy
* Verified that unit tests pass.
* Verified that the Axe.Windows CLI remains functional (scanned Outlook with `--verbosity verbose` and various failures were displayed).
* In a follow-up PR, update system/end-to-end tests for asynchronous cancellation.

##### Known issues with pull request
* While the goal of the Axe.Windows project is not to break client code, given various code changes and increasing maintenance costs it is necessary to restructure the API. Therefore, this PR bumps the Axe.Windows version to 2.0.0.
* The documentation has not yet been updated to match the new API, but as this PR diff is already quite large, I'd like to make those changes in a follow-up.

##### References
* [Task-based Asynchronous Pattern documentation](https://learn.microsoft.com/en-gb/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap)
* [C# asynchronous programming cheat sheet](https://gist.github.com/jonlabelle/841146854b23b305b50fa5542f84b20c)

##### Motivation
Part of feature [1986365](https://mseng.visualstudio.com/1ES/_workitems/edit/1986365) to add an asynchronous API (internal access required to view).

##### Context
One of a series of PRs.

#### Pull request checklist

- [x] Addresses an existing issue: #732


src/Automation/SnapshotCommand.cs
```diff
@@ -7,29 +7,73 @@
...
+        }
+
+        private static Task<ScanResults> TriggerScanAsync(Config config, IScanTools scanTools, CancellationToken cancellationToken)
+        {
+            return Task.Run<ScanResults>(() => GetScanAllResults(config, scanTools, cancellationToken).FirstOrDefault());
+        }
+
+        private static Task<IReadOnlyCollection<ScanResults>> TriggerScanAllAsync(Config config, IScanTools scanTools, CancellationToken cancellationToken)
+        {
+            return Task.Run<IReadOnlyCollection<ScanResults>>(() => GetScanAllResults(config, scanTools, cancellationToken));
```
sfoslund, line 72: nit: personal preference but IMO adding `TriggerScanAllAsync` and `TriggerScanAsync` seems unnecessary, since they both only have one line. I would prefer to call `Task.Run` directly in `ExecuteScanAsync` and `ExecuteScanAll`.

src/Automation/Scanner.cs
```diff
@@ -72,8 +74,25 @@ private IReadOnlyCollection<ScanResults> ExecuteScan(string scanId)
...
+            });
+        }
 
-                return SnapshotCommand.Execute(_config, _scanTools);
+        public Task<ScanResults> ScanAsync(string scanId, CancellationToken cancellationToken)
+        {
+            return ExecutionWrapper.ExecuteCommand(() =>
+            {
+                _scanTools.OutputFileHelper.SetScanId(scanId);
+                return SnapshotCommand.ExecuteScanAsync(_config, _scanTools, cancellationToken);
```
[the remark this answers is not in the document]
codeofdusk: > Concretely, can we remove ExecuteScanAsync and instead use return SnapshotCommand.ExecuteScanAllAsync(_config, _scanTools, cancellationToken).First();?

I think this was added for symmetry with the sync API.

src/Automation/SnapshotCommand.cs
```diff
@@ -7,29 +7,73 @@
...
 using System.Globalization;
 using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Axe.Windows.Automation
 {
     /// <summary>
-    /// Class to take a snapshot (via SnapshotCommand.Execute).
+    /// Class to take a snapshot (via SnapshotCommand.Execute*).
```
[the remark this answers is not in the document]
DaveTryon: > I think the intent here was to refer to all methods starting with `Execute` (i.e. a shell wildcard).

Correct

src/Automation/Scanner.cs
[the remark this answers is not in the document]
sfoslund: > There's a mismatch in the types. Task<ScanResults> and Task<IReadOnlyCollection<ScanResults>> are incompatible.

Could we transform the type with `ContinueWith`? Something like:

```
ScanAllAsync(scanId, cancellationToken).ContinueWith((task) => task.Result.FirstOrDefault(), TaskScheduler.Default);
```

Alternatively, returning a `ScanResultsCollection` instead of the existing return types sounds reasonable to me.

sfoslund (approved):
Looks good to me!

src/Automation/Data/AsyncScan.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+
+namespace Axe.Windows.Automation.Data
+{
+    /// <summary>
+    /// Contains information, such as ScanResults, related to a particular invocation of the ScanAsync method.
+    /// This object currently contains ScanResults.
+    /// However, additional fields may be added at a future time to add extra information without breaking existing users of the IAsyncScanner API.
+    /// </summary>
+    public class AsyncScan
```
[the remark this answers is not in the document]
DaveTryon: > nit: Would `AsyncScanResults` be a more accurate name?

+1

src/Automation/SnapshotCommand.cs
```diff
@@ -2,35 +2,57 @@
...
-        /// Execute the scan
+        /// Execute the scan synchronously
         /// </summary>
         /// <param name="config">A set of configuration options</param>
         /// <param name="scanTools">A set of tools for writing output files,
         /// creating the expected results format, and finding the target element to scan</param>
-        /// <returns>A SnapshotCommandResult that describes the result of the command</returns>
-        public static IReadOnlyCollection<ScanResults> Execute(Config config, IScanTools scanTools)
+        /// <returns>A set of ScanResults objects that describes the result of the command</returns>
+        public static IReadOnlyCollection<ScanResults> ExecuteScanAll(Config config, IScanTools scanTools)
```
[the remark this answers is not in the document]
DaveTryon: +1

src/Automation/Data/AsyncScan.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// The ScanResults object associated with this scan
+        /// </summary>
+        public ScanResults ScanResults { get; private set; }
```
DaveTryon, line 18: This can be simplified:
```suggestion
        public ScanResults ScanResults { get; }
```

This makes the setter both private and accessible only in the constructor

src/Automation/SnapshotCommand.cs
```diff
@@ -2,35 +2,57 @@
...
+            ValidateScanParameters(config, scanTools);
+
+            return GetScanAllResults(config, scanTools, CancellationToken.None);
+        }
+
+        public static Task<AsyncScan> ExecuteScanAsync(Config config, IScanTools scanTools, CancellationToken cancellationToken)
+        {
+            ValidateScanParameters(config, scanTools);
+
+            return Task.Run<AsyncScan>(() => new AsyncScan(GetScanAllResults(config, scanTools, cancellationToken).FirstOrDefault()));
```
DaveTryon, line 40: Creating the `AsyncScan` here works fine as long as `AsyncScan`'s constructor only takes 1 input parameter. We'll have to change this if/when we add additional input parameters, which seems likely to occur at some point. Would it make sense for `GetScanAllResults` to return a fully-formed `AsyncScan` object? The sync wrappers would then just need to return the pieces that match the existing API

sfoslund, line 40: Experimenting with some e2e tests, I'm noticing that the cancellation is only properly handled when we call Task.Run in certain ways. Specifically, when running this code as-is (including some small changes to get it building/ pipe the cancellation token through), if we cancel it halfway through we see that the `isCanceled` property on the task is false, when we expect it to be true. If we want to pass the correct task metadata through to the user it looks like we have to change this slightly to be something like: 

```
return await Task.Run<AsyncScan>(() => new AsyncScan(GetScanAllResults(config, scanTools, cancellationToken).FirstOrDefault())).ConfigureAwait(true);
```

For context, this is the commit where I made these changes: https://github.com/sfoslund/axe-windows/commit/63d98d6dcb56db3701f80feddd831feb6cdabe52

DaveTryon: If we want to avoid adding the `async` keyword, we could use Task.Result to make the API call synchronous.

codeofdusk: > we could use `Task.Result`

This [looks not to be recommended](https://gist.github.com/jonlabelle/841146854b23b305b50fa5542f84b20c) (see the section that starts with heading "the async way of doing things").

sfoslund: > Why should the async keyword (and async/await pattern) be avoided here? Is there some C#-specific reason not to use it?

This stack overflow answer explains why it's generally a good idea in general to prefer returning a task directly over using await: https://stackoverflow.com/a/38017167/12323206. However, in this specific instance I think we might have to, since as far as I can tell, marking this function async is the only way to expose the task to the caller in such a way that the `isCanceled` property is set correctly.

src/Automation/Data/AsyncScan.cs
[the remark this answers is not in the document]
sfoslund: IMO I think it's more confusing as is, I would prefer including the redundant `results` for clarity

src/Automation/Scanner.cs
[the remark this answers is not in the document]
sfoslund: Looks like this point is now moot given the other changes to the API, resolving this thread :)

src/Automation/Scanner.cs
```diff
@@ -19,6 +22,7 @@ internal Scanner(Config config, IScanTools scanTools)
         {
             _config = config ?? throw new ArgumentNullException(nameof(config));
             _scanTools = scanTools ?? throw new ArgumentNullException(nameof(scanTools));
+            ScanOptions _defaultScanOptions = new ScanOptions(null);
```
sfoslund, line 25: looks like this isn't used anywhere, so it can be removed

src/Automation/SnapshotCommand.cs
```diff
@@ -2,43 +2,62 @@
...
+        /// Execute the scan synchronously
         /// </summary>
         /// <param name="config">A set of configuration options</param>
         /// <param name="scanTools">A set of tools for writing output files,
         /// creating the expected results format, and finding the target element to scan</param>
-        /// <returns>A SnapshotCommandResult that describes the result of the command</returns>
-        public static IReadOnlyCollection<ScanResults> Execute(Config config, IScanTools scanTools)
+        /// <param name="cancellationToken">A Cancellation token</param>
+        /// <returns>A set of ScanResults objects that describes the result of the command</returns>
+        public static IReadOnlyCollection<ScanResults> Execute(Config config, IScanTools scanTools, CancellationToken cancellationToken)
```
sfoslund, line 30: Can we remove the CancellationToken parameter here, since it will only ever be CancellationToken.None, and just pass that directly to GetScanAllResults on line 34? That way we could also revert the changes in `SnapshotCommandTests`.

DaveTryon: The CancellationToken here will be provided by the caller of the Async method, so we need to plumb it through

DaveTryon:
@codeofdusk, we've been greenlighted to proceed with the changes that we discussed Tuesday afternoon

src/Automation/Scanner.cs
```diff
@@ -23,58 +27,22 @@ internal Scanner(Config config, IScanTools scanTools)
...
-        /// </summary>
-        /// <returns></returns>
-        public IReadOnlyCollection<ScanResults> ScanAll(string scanId)
-        {
-            return ExecuteScan(scanId);
-        }
-
-        private IReadOnlyCollection<ScanResults> ExecuteScan(string scanId)
-        {
-            return ExecutionWrapper.ExecuteCommand(() =>
```
sfoslund, line 72: Are we okay with no longer having the error handling that the execution wrapper provided?

sfoslund (commented):
This is looking good to me other than a few small nits!

src/Automation/Interfaces/IScanner.cs
```diff
@@ -11,70 +15,32 @@ namespace Axe.Windows.Automation
...
-        /// will contain the exception.
         /// </remarks>
         /// <returns>Information about the scan and any issues detected</returns>
-        /// <exception cref="AxeWindowsAutomationException"/>
-        IReadOnlyCollection<ScanResults> ScanAll();
+        Task<ScanOutput> ScanAsync(ScanOptions scanOptions, CancellationToken cancellationToken);
 
         /// <summary>
-        /// Run AxeWindows automated tests
+        /// synchronously run AxeWindows automated tests
```
DaveTryon, line 33: Start with a capital letter?
```suggestion
        /// Synchronously run AxeWindows automated tests
```

src/Automation/Scanner.cs
```diff
@@ -1,19 +1,23 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Axe.Windows.Automation.Data;
```
DaveTryon, line 3: nit: Can we add a newline between the comments and the using statements?

src/Automation/Scanner.cs
DaveTryon: Yes

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -119,39 +120,7 @@ public void Scan_Integration_WpfControlSampler()
             Scan_Integration_Core(WpfControlSamplerAppPath, WpfControlSamplerKnownErrorCount);
         }
 
-        [TestMethod]
-        [Timeout(30000)]
-        public void SingleWindowScan_MultipleRootsEnabledThrows()
```
DaveTryon, line 124: Do we have another place where we're scanning multiple roots? We should have this somewhere...

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -119,39 +120,7 @@ public void Scan_Integration_WpfControlSampler()
...
-                .Build();
-            var scanner = ScannerFactory.CreateScanner(config);
-
-            var action = new Action(() => scanner.Scan());
-            Assert.ThrowsException<InvalidOperationException>(action);
-        }
-
-        [TestMethod]
-        [Timeout(30000)]
-        public void SingleWindowScanWithID_MultipleRootsEnabledThrows()
```
DaveTryon, line 140: Is this case covered elsewhere?

DaveTryon:
@codeofdusk, this looks awesome! There's a merge conflict in ExecutionWrapper.cs, but I think we can move this out of draft

DaveTryon (approved):

codeofdusk:
@msftbot merge this in two minutes

PR #757, opened 2022-10-07
feat(async-api): Provide separate contexts for each scan

DaveTryon:
#### Details

I apologize in advance for the size of this PR. I wasn't able to identify good ways to split it up into smaller chunks ☹️ 

The async work requires us to support concurrent scans. The existing code can't handle this, as all data lives in a big singleton that is accessed from multiple classes. This creates an `IActionContext` abstraction that is intended to provide a separate data space for each action. The `IActionContext` includes the following:
Field | Content
--- | ---
`DataManager` | The `DataManager` object that stores the UIA hierarchy and associated properties
`SelectAction` | A `SelectAction` object that is bound to the underlying `DataManager` object
`Registrar` | A `Registrar` object for managing custom UIA properties
`TreeWalkerDataContext` | A `TreeWalkerDataContext` that unites the `TreeWalkerForTest` with its associated `CancellationToken` and `Registrar`

It's worth noting that `IActionContext.Registrar` and `IActionControl.TreeWalkerDataContext.Registrar` refer to the same object. The overlap could be removed, but then Actions code that has no relationship to `TreeWalker` objects would be accessing the data through the `TreeViewDataContext`, which looked wrong. This is easy to change if we decide to.

There are 2 implementations of `IActionContext`
- `DefaultActionContext` connects everything to the singleton objects that the previous code used. This object triggers no cleanup on its `Dispose` method, just like the existing code.
- `ScopedActionContext` connects everything to scoped-lifetime objects. This object triggers cleanup when it's disposed to (hopefully) prevent memory leaks

Methodology:
1) Find all places where we use a class's `GetDefaultInstance` method
2) Plumb these through using members from `IActionContext`, working up the stack to Actions
3) At the Action level, split the methods into 2 methods--an internal method that takes an `IActionContext` parameter, and the old public method that calls the internal method, passing in the `DefaultActionContext`. The intended goal is that AIWin can call the existing API and keep its existing behavior
4) In the `Automation` namespace, create a `ScopedActionContext` for the duration of the scan, then use it to call the internal version of the existing API.

Hopefully, when all of the pieces are brought together, we'll end up with AIWin working as it currently does, while the Automation interface can maintain distinct sets of data for each scan. We also now have the plumbing to get a caller-provided `CancellationToken` down to the points in the code where we check for cancellation.

While working on this, I ran into problems with `SelectAction.SetCandidateElementFromProcessId` and `SelectActcion.SetCandidateElementFromHandle`, which have corresponding methods in the `A11yAutomation` class. These problems weren't insurmountable (we just needed to store the `Registrar` in the `SelectionAction`), but since AIWin never calls these methods, it was cleaner to remove them than to fix them.

##### Motivation

Part of async feature

#### Pull request checklist

- [x] Addresses an existing issue: [#1996306](https://mseng.visualstudio.com/1ES/_workitems/edit/1996306)


src/Actions/Actions/CaptureAction.cs
```diff
@@ -27,7 +26,6 @@ public static class CaptureAction
         // Test hooks; making this non-static and enabling constructor injection
         // would be nicer, but this was the minimal set of changes to remove the
         // need to use Fakes to test this class.
-        internal static Func<DataManager> GetDataManager = () => DataManager.GetDefaultInstance();
```
DaveTryon, line 30: This hook existed for testing. It's no longer needed since we can control it via IActionContext.DataManager

src/Actions/Actions/ScreenShotAction.cs
```diff
@@ -17,7 +19,6 @@ namespace Axe.Windows.Actions
     public static class ScreenShotAction
     {
         // unit test hooks
-        internal static Func<DataManager> GetDataManager = () => DataManager.GetDefaultInstance();
```
DaveTryon, line 20: This hook existed for testing. It's no longer needed since we can control it via IActionContext.DataManager

src/Actions/Actions/SelectAction.cs
```diff
@@ -214,36 +218,11 @@ public bool IsPaused
...
-        public void SetCandidateElementFromHandle(IntPtr handle)
-        {
-            lock (_elementContextLock)
-            {
-                var element = A11yAutomation.ElementFromHandle(handle);
-                this.CandidateEC?.Dispose();
-                this.CandidateEC = new ElementContext(element);
-            }
-        }
-
```
DaveTryon, line 246: These methods are never called by AIWin and added extra complexity to properly handle with separate data, so it seemed better to remove them than to add extra complexity for stuff that we don't use

src/Actions/Contexts/ScopedActionContext.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public DataManager DataManager { get; }
+
+        public SelectAction SelectAction { get; }
+
+        public Registrar Registrar => TreeWalkerDataContext.Registrar;
+
+        public TreeWalkerDataContext TreeWalkerDataContext { get; }
+
+        protected virtual void Dispose(bool disposing)
```
sfoslund, line 34: General question: What is the purpose of the `bool disposing` parameter, here and throughout the code base? In what instance is it false?

src/ActionsTests/Contexts/ScopedActionContextTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Threading;
+
+namespace Axe.Windows.ActionsTests.Contexts
+{
+    [TestClass]
+    public class ScopedActionContextTests
+    {
+        [TestMethod]
+        [Timeout(1000)]
+        public void DataManager_IsnotDefaultDataManager()
```
sfoslund, line 18: nit: Not isn't capitalized

src/ActionsTests/Contexts/ScopedActionContextTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using Axe.Windows.Actions.Contexts;
+using Axe.Windows.Desktop.UIAutomation.CustomObjects;
+using Axe.Windows.Desktop.UIAutomation.TreeWalkers;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Threading;
+
+namespace Axe.Windows.ActionsTests.Contexts
+{
+    [TestClass]
+    public class ScopedActionContextTests
```
sfoslund, line 14: Might be worth adding test coverage that the cancellation token passed in is the same as `TreeWalkerDataContext.CancellationToken`. Otherwise very thorough and great to see more test coverage being added!

sfoslund (approved):
A few nits and questions but otherwise looks good!

src/ActionsTests/Contexts/ScopedActionContextTests.cs
DaveTryon: Fixed

src/ActionsTests/Contexts/ScopedActionContextTests.cs
DaveTryon: Added

src/AutomationTests/SnapshotCommandTests.cs
```diff
@@ -396,15 +398,15 @@ public void Execute_NullOutputFilePath_ThrowsException()
             _scanToolsMock.Setup(x => x.ResultsAssembler).Returns(_resultsAssemblerMock.Object);
             _scanToolsMock.Setup(x => x.OutputFileHelper).Returns(_outputFileHelperMock.Object);
 
-            _targetElementLocatorMock.Setup(x => x.LocateRootElements(It.IsAny<int>())).Returns(CreateMockElementArray());
+            _targetElementLocatorMock.Setup(x => x.LocateRootElements(It.IsAny<int>(), It.IsAny<IActionContext>())).Returns(CreateMockElementArray());
```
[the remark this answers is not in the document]
DaveTryon: I'll make a separate PR to refactor some of these. There are a couple of other things that this test class does that vary from our normal practice

src/Actions/Contexts/ScopedActionContext.cs
DaveTryon: Most of this is VS boilerplate. VS had created this code blurb that I deleted but have since restored:

```
        // TODO: override a finalizer only if Dispose(bool disposing) above has code to free unmanaged resources.
        // ~ScopedActionContext() {
        //   // Do not change this code. Put cleanup code in Dispose(bool disposing) above.
        //   Dispose(false);
        // }
```

If this code blurb is uncommented, then the `disposing` parameter would be false if the object were garbage collected before its `Dispose` method was called. A derived class can also call the `Dispose` method with `disposing` = false. In this particular class, neither of those cases exist, so we could reasonably remove the parameter. Most of the time, though, we just leave the default boilerplate as it was originally created. The boilerplate covers the broadest general case, even if it's generally not fully leveraged. There's a fair write-up of this pattern at https://vkontech.com/the-dispose-pattern-step-by-step/, and I'm happy to chat about it if you have other questions 😃

PR #758, opened 2022-10-08
chore: Refactor unit tests

DaveTryon:
#### Details

A PR comment in #757 suggested some unit test refactoring, which we put into a separate PR for clarity. These refactorings include:
- 2 classes use the `MockRepository` pattern, while no others do. For consistency, don't use it here, either
- Use fixtures and the `[TestInitialize]` pattern in the same 2 classes
- Move all mock verification into a separate `VerifyAllMocks` method
- Try to move most of the repeated mock setup into helper methods to help focus what each unit test does uniquely. It's possible to put so much into the helper classes that it's hard to discern what the unit test is actually doing. Hopefully these changes stop short of that point.

##### Motivation

Make the unit tests more readable and more consistent

### DIFF
--- a/src/AutomationTests/ScanToolsBuilderTests.cs
+++ b/src/AutomationTests/ScanToolsBuilderTests.cs
@@ -11,34 +11,55 @@ namespace Axe.Windows.AutomationTests
     [TestClass]
     public class ScanToolsBuilderTests
     {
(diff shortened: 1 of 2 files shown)


sfoslund (approved):
So much cleaner, yay!!! 🎉

PR #762, opened 2022-10-14
feat(async-api): Update documentation for the Axe.Windows 2.0 API

codeofdusk:
#### Details
This PR updates documentation for the Axe.Windows 2.0 API and adds a migration guide for 1.x users.

##### Motivation
Part of feature [1986365](https://mseng.visualstudio.com/1ES/_workitems/edit/1986365) to add an asynchronous API (internal access required to view).

##### Context
One of a series of PRs.

#### Pull request checklist
- [n/a] Addresses an existing issue: #000

### DIFF
--- a/docs/AutomationReference.md
+++ b/docs/AutomationReference.md
@@ -245,6 +245,7 @@ example below):
 ```C#
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using Axe.Windows.Automation;
 
@@ -269,7 +270,7 @@ namespace AxeWindowsDemo
 
             try
             {
-                var output = await scanner.ScanAsync(null);
+                var output = await scanner.ScanAsync(null, CancellationToken.None);
                 foreach (var window in output.WindowScanOutputs)
                     Assert.AreEqual(0, window.ErrorCount);
             }
--- a/docs/AutomationReference.md
+++ b/docs/AutomationReference.md
@@ -305,7 +305,7 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
 `AxeWindowsAutomationException` is thrown for errors in `Axe.Windows.Automation`'s logic where meaningful error reporting can be generated. In some situations, the `Exception.InnerException` property may contain a corresponding system-level exception for errors encountered by Axe.Windows. Other exceptions may be thrown by the system from calls to this library.
 
 ### Migrating from Axe.Windows 1.x
-Migration from Axe.Windows 1x to Axe.Windows 2.0 should require minimal code changes for most projects.
+Migration from Axe.Windows 1.x to Axe.Windows 2.0 should require minimal code changes for most projects.
 
 #### `IScanner.ScanAll` removed
 The `ScanAll` method for scanning multiple top-level windows has been removed. Use `ScanAsync` or `Scan` instead.
--- a/docs/AutomationReference.md
+++ b/docs/AutomationReference.md
@@ -142,8 +142,6 @@ The `Scan` method synchronously runs AxeWindows automated tests using the config
 
 ###### Parameters
 `IScanner.Scan` accepts one parameter:
-###### Parameters
-`IScanner.ScanAsync` accepts two parameters:
 
 **Name** | **Type** | **Description**
 ---|---|---
@@ -156,7 +154,7 @@ The `Scan` method synchronously runs AxeWindows automated tests using the config
 The `ScanOptions` constructor accepts the following arguments:
 
 **Name** | **Type** | **Description** | **Default value**
----|---|---
+---|---|---|---
 scanId | `string` | A string identifier for the scan. If the scan produces output files based on the `Config` object used to create the scanner, the output files will be given the name of the scan id (e.g., MyScanId.a11ytest). | `null`
 
 #### ScanOutput
@@ -250,7 +248,7 @@ example below):
 -   Prerequisite: Your project *must* be able to use .NET Standard 2.0 libraries.
 -   If you’re using NuGet, add the appropriate feed to your project.
 -   Add **using Axe.Windows.Automation;** to your code.
--   Follow the steps in [How To Run An Automated Scan](#how-to-run-an-automated-scan).
+-   Invoke Axe.Windows similar to the [example](#example) below.
 
 #### Example
 ```C#
@@ -316,6 +314,8 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
 
 `AxeWindowsAutomationException` is thrown for errors in `Axe.Windows.Automation`'s logic where meaningful error reporting can be generated. In some situations, the `Exception.InnerException` property may contain a corresponding system-level exception for errors encountered by Axe.Windows. Other exceptions may be thrown by the system from calls to this library.
 
+If a `CancellationToken` is provided to the library which is later cancelled, the original task will throw an `AggregateException` containing a `TaskCancelledException`.
+
 ### Migrating from Axe.Windows 1.x
 Migration from Axe.Windows 1.x to Axe.Windows 2.0 should require minimal code changes for most projects.
 
@@ -330,7 +330,7 @@ In place of this | Do this
 `IScanner.Scan("scanId")` | `IScanner.Scan(new ScanOptions(scanId: "scanId"))`
 
 #### `ScanResults` replaced with `ScanOutput`
-The new return type of `IScanner`'s methods is `ScanOutput`. This object contains a `WindowScanOutputs` field, a `IReadOnlyCollection` of `WindowScanOutput` objects. These objects contain the same fields as the previous `ScanResults` class.
+The new return type of `IScanner`'s methods is [`ScanOutput`](#scanoutput). This object contains a `WindowScanOutputs` field, a `IReadOnlyCollection` of `WindowScanOutput` objects. These objects contain the same fields as the previous `ScanResults` class.
 
 In place of this | Do this
 ---|---
--- a/docs/AutomationReference.md
+++ b/docs/AutomationReference.md
@@ -127,7 +127,12 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
 The `ScanAsync` method asynchronously runs AxeWindows automated tests using the config provided at the time of creation of the scanner. This method should be `await`ed.
 
 ###### Parameters
-`IScanner.ScanAsync` accepts two parameters: an instance of `ScanOptions` containing custom settings for this scan (or `null` for default options), and a [`CancellationToken`](https://learn.microsoft.com/en-gb/dotnet/api/system.threading.cancellationtoken?view=netstandard-2.0).
+`IScanner.ScanAsync` accepts two parameters:
+
+**Name** | **Type** | **Description**
+---|---|---
+`scanOptions` | `ScanOptions` | An object describing custom settings for this scan. Pass `null` for default options.
+`cancellationToken` | `CancellationToken` | A [`CancellationToken`](https://learn.microsoft.com/en-gb/dotnet/api/system.threading.cancellationtoken?view=netstandard-2.0).
 
 ###### Return object
 `ScanAsync` returns a `ScanOutput` object.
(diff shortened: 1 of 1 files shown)


docs/AutomationReference.md
```diff
@@ -123,26 +123,37 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
...
 
+##### `ScanAsync`
+The `ScanAsync` method asynchronously runs AxeWindows automated tests using the config provided at the time of creation of the scanner. This method should be `await`ed.
+
 ##### `Scan`
-The Scan runs AxeWindows automated tests using the config provided at the time of creation of the scanner.
+The `Scan` method synchronously runs AxeWindows automated tests using the config provided at the time of creation of the scanner, and blocks until the scan is complete.
 
-###### Parameters
+##### Parameters
```
DaveTryon, line 132: Docs usually have an API, then information about the API as children. Links as children are normal and expected. As currently drafted, ScanAsync has no information and it's all in siblings. Can we put Parameters and Return object as H6 children of each of the H5 Scan and the H5 ScanAsync, then they can link to the types? it might make sense to have an H5 Objects and list the ScanOutput and ScanOptions--along with the other types--there. I

docs/AutomationReference.md
```diff
@@ -223,42 +234,43 @@ example below):
...
+                .Build();
+
+            var scanner = ScannerFactory.CreateScanner(config);
+
+            try
+            {
+                var output = await scanner.ScanAsync(null);
+                Assert.AreEqual(0, output.WindowScanOutputs.First().ErrorCount);
+            }
+            catch(AxeWindowsAutomationException e)
```
DaveTryon, line 266: Now that ExecutionWrapper is gone, the caller needs to handle a broader set of Exceptions. Maybe this?
```suggestion
            catch(Exception e)
```

docs/AutomationReference.md
```diff
@@ -223,42 +234,43 @@ example below):
...
+                .WithOutputFileFormat(OutputFileFormat.A11yTest)
+                .WithOutputDirectory(outputDir)
+                .Build();
+
+            var scanner = ScannerFactory.CreateScanner(config);
+
+            try
+            {
+                var output = await scanner.ScanAsync(null);
+                Assert.AreEqual(0, output.WindowScanOutputs.First().ErrorCount);
```
DaveTryon, line 264: Might be worth a comment that we're using `First()` because we know that we only have 1 top-level window

docs/AutomationReference.md
```diff
@@ -223,42 +234,43 @@ example below):
...
 #### Example
 ```C#
-    using System;
-    using System.Collections.Generic;
-    using Axe.Windows.Automation;
-
-    namespace AxeWindowsDemo
+using System;
+using System.Threading.Tasks;
```
DaveTryon, line 239: We should ensure that our sample comples. I believe the `First()` method requires System.Linq

docs/AutomationReference.md
```diff
@@ -280,12 +292,25 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
 
 #### Error handling
 
-`AxeWindowsAutomationException` is thrown for all unhandled errors in Axe.Windows.Automation. If an exception was thrown from code not owned by Axe.Windows.Automation, that exception will be wrapped in the `Exception.InnerException` property.
+`AxeWindowsAutomationException` is thrown for errors in `Axe.Windows.Automation`'s logic where meaningful error reporting can be generated. In some situations, the `Exception.InnerException` property may contain a corresponding system-level exception for errors encountered by Axe.Windows.
```
DaveTryon, line 295: Should we clarify that other Exception types can also be thrown here?

docs/AutomationReference.md
```diff
@@ -280,12 +292,25 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
...
+
+#### Migrating from Axe.Windows 1.x
```
DaveTryon, line 297: This doesn't feel like an H4. Maybe H3, so it's a peer of the Class details?

docs/AutomationReference.md
```diff
@@ -280,12 +292,25 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
...
+Migration to Axe.Windows 2.0 from 1.x should require minimal code changes for most projects.
```
DaveTryon, line 298: nit: Since people will be migrating *from* 1.x *to* 2.0, should we describe it that way?
```suggestion
Migration from Axe.Windows 1x to Axe.Windows 2.0 should require minimal code changes for most projects:
```

docs/AutomationReference.md
```diff
@@ -280,12 +292,25 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
...
+
+##### `IScanner.ScanAll` removed
+The `ScanAll` method for scanning multiple top-level windows has been removed. Use `ScanAsync` or `Scan` instead.
```
DaveTryon, line 301: This brings up an interesting point that I didn't consider before. Scanning multiple-level windows was a two-step process--the user had to enable multiple scans on the ConfigBuilder, then call the ScanAll API. That was for backward compatibility. Since 2.0 is a breaking change, maybe we can revisit that. What are your thoughts on enabling multiple top-level windows by default?

docs/AutomationReference.md
```diff
@@ -123,26 +123,46 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
 
 #### IScanner
 
+##### `ScanAsync`
+The `ScanAsync` method asynchronously runs AxeWindows automated tests using the config provided at the time of creation of the scanner. This method should be `await`ed.
+
+###### Parameters
+`IScanner.ScanAsync` accepts two parameters: an instance of `ScanOptions` containing custom settings for this scan (or `null` for default options), and a [`CancellationToken`](https://learn.microsoft.com/en-gb/dotnet/api/system.threading.cancellationtoken?view=netstandard-2.0).
```
DaveTryon, line 130: Can we put these in a table similar to the ScanOptions constructor? That's the norm for our docs

docs/AutomationReference.md
```diff
@@ -123,26 +123,46 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
...
+###### Return object
+`Scan` returns a `ScanOutput` object.
 
-The `Scan` method accepts 0 or 1 parameters.
+#### `ScanOptions`
+The `ScanOptions` constructor accepts the following arguments:
 
 **Name** | **Type** | **Description**
 ---|---|---
 scanId | `string` | A string identifier for the scan. If the scan produces output files based on the `Config` object used to create the scanner, the output files will be given the name of the scan id (e.g., MyScanId.a11ytest).
```
DaveTryon, line 149: For this table, should we have a column for the default parameter/behavior?

docs/AutomationReference.md
```diff
@@ -123,26 +123,46 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
...
+
+###### Return object
+`ScanAsync` returns a `ScanOutput` object.
+
 ##### `Scan`
-The Scan runs AxeWindows automated tests using the config provided at the time of creation of the scanner.
+The `Scan` method synchronously runs AxeWindows automated tests using the config provided at the time of creation of the scanner, and blocks until the scan is complete.
 
 ###### Parameters
+`IScanner.Scan` accepts one parameter: an instance of `ScanOptions` containing custom settings for this scan (or `null` for default options).
```
DaveTryon, line 139: Table?

DaveTryon, line 139: Where possible, let's use internal links. Instead of `ScanOptions`, it could be [`ScanOptions`](https://github.com/microsoft/axe-windows/blob/main/docs/AutomationReference.md#scanoptions)

docs/AutomationReference.md
```diff
@@ -123,26 +123,46 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
...
+
+###### Return object
+`Scan` returns a `ScanOutput` object.
```
DaveTryon, line 142: Link to ScanIOutput in doc
```suggestion
`Scan` returns a [`ScanOutput`](https://github.com/microsoft/axe-windows/blob/main/docs/AutomationReference.md#scanoutput) object.
```

DaveTryon (commented):
We should consider leveraging internal links in this document. For example, when we say that we have a collection of `ScanResult` objects, it would be ideal if clicking the `ScanResult` text would take us to where `ScanResult` is defined. This goes beyond the scope of the V2 API changes, though.

For example `ScanResult` would become [`ScanResult`](https://github.com/microsoft/axe-windows/blob/main/docs/AutomationReference.md#scanresult)

docs/AutomationReference.md
```diff
@@ -280,12 +303,25 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
...
-a solution where you can make those calls from separate processes.
+#### Calling `IScanner.Scan`
+
+In place of this | Do this
+---|---
+`IScanner.Scan()` | `IScanner.Scan(null)`
+`IScanner.Scan("scanId")` | `IScanner.Scan(new ScanOptions(scanId: "scanId"))`
+
+#### `ScanResults` replaced with `ScanOutput`
+The new return type of `IScanner`'s methods is `ScanOutput`. This object contains a `WindowScanOutputs` field, a `IReadOnlyCollection` of `WindowScanOutput` objects. These objects contain the same fields as the previous `ScanResults` class.
```
DaveTryon, line 322: This would be a prime example of internal reference, something like this:
```suggestion
The new return type of `IScanner`'s methods is [`ScanOutput`](Link to ScanIOutput in doc
```suggestion
`Scan` returns a [`ScanOutput`](https://github.com/microsoft/axe-windows/blob/main/docs/AutomationReference.md#scanoutput) 
```

The long-term goal is to have things described in detail in exactly one place to avoid the risk of things drifting out of sync.

codeofdusk: I'm worried about linking to one particular branch. What if someone forks?

docs/AutomationReference.md
```diff
@@ -280,12 +314,25 @@ These versions embed the symbols into the assemblies and support [SourceLink](ht
 
 #### Error handling
 
-`AxeWindowsAutomationException` is thrown for all unhandled errors in Axe.Windows.Automation. If an exception was thrown from code not owned by Axe.Windows.Automation, that exception will be wrapped in the `Exception.InnerException` property.
+`AxeWindowsAutomationException` is thrown for errors in `Axe.Windows.Automation`'s logic where meaningful error reporting can be generated. In some situations, the `Exception.InnerException` property may contain a corresponding system-level exception for errors encountered by Axe.Windows. Other exceptions may be thrown by the system from calls to this library.
```
sfoslund, line 317: Can we add a note here that if the cancellation token is cancelled, then we will also throw an aggregate exception which contains a `TaskCancelledException`?

docs/AutomationReference.md
```diff
@@ -123,26 +123,57 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
...
 
 ###### Return object
+`Scan` returns a `ScanOutput` object.
 
-The `Scan` method returns a `ScanResults` object and has the following properties:
+#### `ScanOptions`
+The `ScanOptions` constructor accepts the following arguments:
+
+**Name** | **Type** | **Description** | **Default value**
+---|---|---
```
DaveTryon, line 159: This table is malformed because we have more column headers than columns defined. I think this needs to be
```suggestion
---|---|---|---
```

docs/AutomationReference.md
```diff
@@ -123,26 +123,57 @@ The `ScannerFactory.CreateScanner` method returns an `IScanner` object.
...
+###### Return object
+`ScanAsync` returns a `ScanOutput` object.
+
+##### `Scan`
+The `Scan` method synchronously runs AxeWindows automated tests using the config provided at the time of creation of the scanner, and blocks until the scan is complete.
+
+###### Parameters
+`IScanner.Scan` accepts one parameter:
+###### Parameters
+`IScanner.ScanAsync` accepts two parameters:
```
DaveTryon, line 146: These have been moved up and should be removed
```suggestion
```

docs/AutomationReference.md
DaveTryon: The change to always enable multiple top-level windows has been merged

DaveTryon:
Not directly related to this PR, but I noticed that we a link labeled "How To Run An Automated Scan", but no corresponding section. I think this was broken in #102. It could probably just say something like "Follow a pattern similar to what is shown in the following example"

DaveTryon (approved):

PR #764, opened 2022-10-17
feat(async-api): Adding multithread e2e test coverage

sfoslund:
#### Details

Add e2e test coverage for new multithreading scenarios:

- Scans running in parallel and completing successfully
- Scans running in parallel and being cancelled at various points in their execution

##### Motivation

Add e2e test coverage

##### Context

These test cases cover what we believe to be the more realistic applications of the new `ScanAsync` API added in this feature. 

#### Pull request checklist

- [x] Contributes to [1986365](https://mseng.visualstudio.com/1ES/_workitems/edit/1986365)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -143,53 +145,118 @@ public void ScanAsync_WindowsFormsSampler_TaskIsCancelled_ThrowsCancellationExce
...
-            return ValidateOutput(output, expectedErrorCount, expectedWindowCount);
+                taskFuncs.Add(() => scanner.ScanAsync(null, token));
+            }
+
+            // Kick scan tasks off
+            var tasks = new List<Task<ScanOutput>>();
+            foreach (var func in taskFuncs)
+            {
+                tasks.Add(func());
+                Thread.Sleep(300); // TODO investigate why this is necessary (without this UIAutomation.GetRootElement() throws)
```
sfoslund, line 256: @codeofdusk @DaveTryon do either of you recognize the following error for UIAutomation? It's getting thrown if you start `ScanAsync` tasks too close together. I've added this thread.sleep for now while I wrote the tests but I strongly suspect that this is what's causing the error in CI. 

> Error HRESULT E_FAIL has been returned from a call to a COM component.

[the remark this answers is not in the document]
sfoslund: 300 seemed to be around the minimum where I stopped seeing the UIAutomation error, although I'm not sure why that is. I just tested FromTicks(1), and that seems to produce the error too

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -143,53 +145,118 @@ public void ScanAsync_WindowsFormsSampler_TaskIsCancelled_ThrowsCancellationExce
...
+                    .WithOutputFileFormat(OutputFileFormat.A11yTest);
 
-            var scanner = ScannerFactory.CreateScanner(config);
+                var config = builder.Build();
 
-            var output = ScanAsyncWithProvisionForBuildAgents(scanner);
+                var scanner = ScannerFactory.CreateScanner(config);
 
-            return ValidateOutput(output, expectedErrorCount, expectedWindowCount);
+                taskFuncs.Add(() => scanner.ScanAsync(null, token));
```
DaveTryon, line 248: Part of the goal here is to run the scans concurrently. On a fast machine, it might be possible for the first call to `scanner.ScanAsync` to complete before later calls to `scanner.ScanAsync` are even started. Is there a way that we can start all of the apps up front, possibly building their Scanners, then put the calls to `scanner.ScanAsync` into a tight loop after all of the potential delays are out of the way?

sfoslund: That's what I'm attempting to do here- On this line (248) I'm creating a func instead of actually kicking off the scan. This way, we prepare everything (starting the app, creating the scanners, etc) ahead of time and then actually kick off the scans in a tight loop on line 255. Of course, there's the thread.sleep there to work around that UIAutomation issue for now, which means it's not really a tight loop at the moment, but that's a separate problem

src/AutomationTests/AutomationIntegrationTests.cs
sfoslund: Looks like UIAutomation has some odd multithreaded behavior, but adding a lock around those calls fixed it for me locally (see [c148be8](https://github.com/microsoft/axe-windows/pull/764/commits/c148be8ffa8186be41c110f43574fb9fc04dd13c))

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -110,9 +113,12 @@ public static IEnumerable<DesktopElement> ElementsFromProcessId(int pid, Registr
                 // if not, it will throw an ArgumentException
                 using (var proc = Process.GetProcessById(pid))
                 {
-                    root = UIAutomation.GetRootElement();
-                    matchingElements = FindProcessMatchingChildrenOrGrandchildren(root, pid);
```
DaveTryon, line 114: `FindProcessMatchingChildrenOrGrandchildren` calls into `GetTreeWalker`, which then returns a shared single object. This may be a part of why this needs to be locked. The concerning thing is that the scanner will also find its way into `GetTreeWalker`, and those would be a lot harder to fix. We might do well to have a separate `IUIAutomation` object inside each `TreeWalkerDataContext`. I'll start working on that change

sfoslund: Sounds good, I'll revert the locking change and merge in your change once it's ready :)

DaveTryon (approved):

PR #767, opened 2022-10-18
feat(async-api): Provide instanced IUIAUtomation objects

DaveTryon:
#### Details

#757 set up separate data contexts, but it didn't go far enough. When testing, we realized that the `A11yAutomation` static class was still being shared across all of those data contexts, so the `IUIAutomation` object held within that class was also being shared between contextx. These changes break down as follows:
1. Update `A11yAutomation` to support multiple instances, with each instance having its own `IUIAutomation`.
2. Add an `A11yAutomation` object to `TreeWalkerDataContext`
3. Rename `TreeWalkerDataContext` to `DesktopDataContext`, move it one level up the directory structure
4. Add an `IActionContext` member in the `BaseTracker` class, so each tracker is bound to a context. In practice, these will always be the default contexts (they're not used while scanning), but it puts that decision into a single place in the code
5. Update the plumbing to make sure that we have the `DesktopDataContext` or `A11yAutomation`, as appropriate
6. In methods that are called from AIWin, add a shim that passes in the default context
7. Some methods in `A11yAutomation` also need the `DesktopDataContext`. Add a check to ensure data consistency.
8. Update unit tests where feasible
9. Remove some unused methods from the A11yAutomation` class

##### Motivation

Part of async feature

##### Context

There are some places where classes or methods could have more restricted visibility. I left those out for this PR

#### Pull request checklist

### DIFF
--- a/src/Actions/Trackers/FocusTracker.cs
+++ b/src/Actions/Trackers/FocusTracker.cs
@@ -1,5 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Axe.Windows.Actions.Contexts;
 using Axe.Windows.Core.Bases;
--- a/src/Actions/Trackers/FocusTracker.cs
+++ b/src/Actions/Trackers/FocusTracker.cs
@@ -1,5 +1,7 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information
+
+using Axe.Windows.Actions.Contexts;
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Types;
 using Axe.Windows.Desktop.Types;
@@ -22,7 +24,7 @@ namespace Axe.Windows.Actions.Trackers
         /// Constructor
         /// </summary>
         /// <param name="action"></param>
-        public FocusTracker(Action<A11yElement> action) : base(action)
+        public FocusTracker(Action<A11yElement> action) : base(action, DefaultActionContext.GetDefaultInstance())
         {
             this.EventHandler = new EventListenerFactory(null); // listen for all element. it works only for FocusChangedEvent
         }
--- a/src/Core/Bases/A11yProperty.cs
+++ b/src/Core/Bases/A11yProperty.cs
@@ -47,6 +47,7 @@ namespace Axe.Windows.Core.Bases
             }
         }
 
+        // TODO: This does not yet support different custom properties in concurrent scans
         static readonly Dictionary<int, ITypeConverter> TypeConverterMap = new Dictionary<int, ITypeConverter>();
 
         /// <summary>
--- a/src/Desktop/UIAutomation/A11yAutomation.cs
+++ b/src/Desktop/UIAutomation/A11yAutomation.cs
@@ -260,9 +260,8 @@ namespace Axe.Windows.Desktop.UIAutomation
         /// <returns></returns>
         internal A11yElement GetAppElement(A11yElement e, DesktopDataContext dataContext)
         {
-            if (!ReferenceEquals(dataContext.A11yAutomation, this)) throw new ArgumentException("Called on wrong context", nameof(dataContext));
+            EnsureContextConsistency(dataContext);
 
-            // TODOL Check for consistency
             var walker = GetTreeWalker(TreeViewMode.Control);
             var tree = new DesktopElementAncestry(TreeViewMode.Control, e, false, dataContext);
             Marshal.ReleaseComObject(walker);
@@ -431,7 +430,7 @@ namespace Axe.Windows.Desktop.UIAutomation
             if (dataContext == null) throw new ArgumentNullException(nameof(dataContext));
             if (!ReferenceEquals(dataContext.A11yAutomation, this))
             {
-                throw new ArgumentException("TODO: Resource string about consistency", nameof(dataContext));
+                throw new ArgumentException(Resources.ErrorMessages.InvalidContext, nameof(dataContext));
             }
         }
     }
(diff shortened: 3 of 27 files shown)


src/Actions/Trackers/FocusTracker.cs
```diff
@@ -1,5 +1,7 @@
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information
```
sfoslund, line 2: It looks like this is triggering the license check to fail... I think it might be the missing period at the end?

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -136,17 +150,17 @@ public static IEnumerable<DesktopElement> ElementsFromProcessId(int pid, Registr
         /// <summary>
         /// Get DesktopElement from UIAElement interface.
         /// </summary>
-        /// <param name="uia"></param>
-        /// <returns></returns>
-        private static DesktopElement ElementFromUIAElement(IUIAutomationElement uia, Registrar registrar)
+        private DesktopElement ElementFromUIAElement(IUIAutomationElement uia, DesktopDataContext dataContext)
         {
+            EnsureContextConsistency(dataContext);
```
sfoslund, line 155: nit: it looks like this is only called from ElementsFromProcessId, which already called EnsureContextConsitency, so this seems to be redundant

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -391,23 +393,13 @@ public static A11yElement NormalizedElementFromPoint(int xPos, int yPos, TreeVie
...
-        /// <summary>
-        /// Get DesktopElement from Desktop
-        /// </summary>
-        /// <returns></returns>
-        public static DesktopElement ElementFromDesktop()
+        private void EnsureContextConsistency(DesktopDataContext dataContext)
         {
-            return new DesktopElement(UIAutomation.GetRootElement());
+            if (dataContext == null) throw new ArgumentNullException(nameof(dataContext));
+            if (!ReferenceEquals(dataContext.A11yAutomation, this))
```
sfoslund, line 397: I don't love that we have to make this check, can we possibly keep the methods where we pass in the dataContext static and only use the UIAutomation instance kept in the data context?

src/Desktop/UIAutomation/TreeWalkers/DesktopElementAncestry.cs
```diff
@@ -52,15 +52,20 @@ public class DesktopElementAncestry
         /// </summary>
         /// <param name="walker"></param>
         /// <param name="e"></param>
-        public DesktopElementAncestry(TreeViewMode mode, A11yElement e, bool setMem = false, Registrar registrar = null)
+        public DesktopElementAncestry(TreeViewMode mode, A11yElement e)
+            : this (mode, e, false, DesktopDataContext.DefaultContext)
+        {
+        }
+
+        internal DesktopElementAncestry(TreeViewMode mode, A11yElement e, bool setMem, DesktopDataContext dataContext)
```
DaveTryon, line 60: setMem is never true, so I think I'll remove it

sfoslund (commented):
Good changes overall, sorry for the slow review, lots to get through but I am done for now! I did a quick merge with my e2e tests branch and still seem to be seeing the same error in the `ScanAsync_WildlifeManager_MultipleProcessesCancelled_ThrowsCancellationException` test, so I think we might still need some more changes here.

DaveTryon:
> Good changes overall, sorry for the slow review, lots to get through but I am done for now! I did a quick merge with my e2e tests branch and still seem to be seeing the same error in the `ScanAsync_WildlifeManager_MultipleProcessesCancelled_ThrowsCancellationException` test, so I think we might still need some more changes here.

@sfoslund, I've done more digging here and I've isolated the problem to UIAutomation.GetRootElement. Ultimately, y fix was basically the same as yours--add a lock around the call. I've also added a comment about the fact that we don't yet support different custom properties between concurrent scans in the same process. I suspect that we can add this as a later time (later PR or after this feature)

src/Desktop/UIAutomation/A11yAutomation.cs
DaveTryon: I tried reverting the classes that receive the `DesktopDataContext` object to static, and it wasn't pretty. We had static methods calling non-static ones and the code got really hard to read. I can investigate splitting this into 2 classes, but we may have the same problem. Given the size of the current PR, I'd be inclined to move that effort into a separate PR

sfoslund: I agree, following up in a different PR makes sense :)

src/Core/Bases/A11yProperty.cs
```diff
@@ -47,6 +47,7 @@ public string TextValue
             }
         }
 
+        // TODO: This does not yet support different custom properties in concurrent scans
```
sfoslund, line 50: nit: Is this a large enough item that we should file an issue? I'm concerned we'll forget about a TODO in the code if we don't act on it immedietly.

src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -162,10 +174,11 @@ private static DesktopElement ElementFromUIAElement(IUIAutomationElement uia, Re
         /// <summary>
         /// Get DesktopElements from UIAElements.
         /// </summary>
-        /// <param name="uia"></param>
         /// <returns>An IEnumerable of <see cref="DesktopElement"/></returns>
-        private static IEnumerable<DesktopElement> ElementsFromUIAElements(IList<IUIAutomationElement> elementList, Registrar registrar)
+        private IEnumerable<DesktopElement> ElementsFromUIAElements(IList<IUIAutomationElement> elementList, DesktopDataContext dataContext)
         {
+            EnsureContextConsistency(dataContext);
```
sfoslund, line 180: Looks like this is only called from `ElementsFromProcessId`, which already has a check, so this is redundant

sfoslund (approved):

src/Desktop/UIAutomation/A11yAutomation.cs
DaveTryon: Fixed

PR #768, opened 2022-10-18
feat(localization): Remove hardcoded strings

sfoslund:
#### Details

Audit/ clean up hardcoded strings throughout the code base in preparation for localizing in the future.

##### Motivation

Ensure all strings that need to be localized are in resource files, and additionally code cleanliness

##### Context

- There are quite a few exceptions that use CultureInfo.InvariantCulture but as we have decided not to localize exceptions at this time I have left those as is for now
- This PR also removes many hardcoded strings from exceptions, which is not strictly necessary as we are not localizing exceptions, but I included anyways for code cleanliness

#### Pull request checklist

- [x] Addresses an existing issue: [2000537](https://dev.azure.com/mseng/1ES/_workitems/edit/2000537)


DaveTryon:
Please also check the code for instances of `using static System.FormattableString;`. TSA flagged these as errors in places where we used interpolated strings for Exception messages. If we're no longer using interpolated strings for Exceptions, then we shouldn't need them.

src/Core/Bases/A11yElement.cs
```diff
@@ -417,7 +419,7 @@ public T GetPlatformPropertyValue<T>(int propertyId)
         {
             var property = this.PlatformProperties?.ById(propertyId);
             if (property == null) return default;
-            if (!(property.Value is T)) throw new AxeWindowsException(Invariant($"Expected property.Value, which is type {property.Value.GetType().Name}, to be type {typeof(T).Name}"));
+            if (!(property.Value is T)) throw new AxeWindowsException(String.Format(CultureInfo.InvariantCulture, ErrorMessages.PropertyValueTypeUnexpected, property.Value.GetType().Name, typeof(T).Name));
```
DaveTryon, line 422: A potentially localizable string + CultureInfo.InvariantCulture seems a bit strange. Part of me wants to use CultureInfo.CurrentCulture here, so that we're all set if/when we localize exception messages, although that might make the exception messages a little harder to process. Another part of me wants to have some sort of extension method that hides the formatting in a single place, so that this line might look something like this:
```
if (!(property.Value is T)) throw new AxeWindowsException(ErrorMessages.PropertyValueTypeUnexpected.With2Parameters(property.Value.GetType().Name, typeof(T).Name));
```

The With2Parameters method would own the formatting, the CultureInfo, and perhaps some way to call out errors if the string and the parameter count somehow get out of sync.

To be 100% clear, these are just musings--feel free to ignore them as appropriate

sfoslund:
> Please also check the code for instances of using static System.FormattableString;. TSA flagged these as errors in places where we used interpolated strings for Exception messages. If we're no longer using interpolated strings for Exceptions, then we shouldn't need them.

There are a lot of instances of FormattableString that touch a lot of files, so I went ahead and separated that piece into a different PR: https://github.com/microsoft/axe-windows/pull/771

src/Core/Bases/A11yElement.cs
sfoslund: The usage of CultureInfo.InvariantCulture all over the place was bothering me too, so I think the extension method is a good idea. I've implemented it in the latest commit, let me know what you think!

src/Automation/TargetElementLocator.cs
```diff
@@ -23,7 +24,7 @@ public IEnumerable<A11yElement> LocateRootElements(int processId, IActionContext
             }
             catch (Exception ex)
             {
-                throw new AxeWindowsAutomationException(string.Format(CultureInfo.InvariantCulture, ErrorMessages.ErrorFailToGetRootElementsOfProcess, processId, ex), ex);
+                throw new AxeWindowsAutomationException(ErrorMessages.ErrorFailToGetRootElementsOfProcess.WithParameters(processId, ex), ex);
```
DaveTryon, line 27: Looking at this now, maybe we don't need to include the Exception in the message, since it's already in the innerException?

src/Core/Bases/A11yElement.cs
```diff
@@ -417,7 +419,7 @@ public T GetPlatformPropertyValue<T>(int propertyId)
         {
             var property = this.PlatformProperties?.ById(propertyId);
             if (property == null) return default;
-            if (!(property.Value is T)) throw new AxeWindowsException(Invariant($"Expected property.Value, which is type {property.Value.GetType().Name}, to be type {typeof(T).Name}"));
+            if (!(property.Value is T)) throw new AxeWindowsException(ExtensionMethods.WithParameters(ErrorMessages.PropertyValueTypeUnexpected, property.Value.GetType().Name, typeof(T).Name));
```
DaveTryon, line 422: There are 7 places where you're using `ExtensionMethods.WithParameters(foo...` instead of `foo.WithParameters(...`

I'm not sure why the 7 instances are different...

src/Core/Bases/A11yProperty.cs
```diff
@@ -159,7 +159,7 @@ private string GetBoundingRectangleText()
             if ((double)arr[2] < 0 || (double)arr[3] < 0)
             {
                 // the 3rd and 4th values in array are negative value, we need to show value in different format like l,t,w,h
-                text = string.Format(CultureInfo.InvariantCulture, "[l={0},t={1},w={2},h={3}]", arr[0], arr[1], arr[2], arr[3]);
+                text = ExtensionMethods.WithParameters("[l={0},t={1},w={2},h={3}]", arr[0], arr[1], arr[2], arr[3]);
```
DaveTryon, line 162: This string gets included in what gets returned to the scan caller (you can test this by running the CLI in verbose mode). As such, this might be a string to localize

src/Core/Misc/ExtensionMethods.cs
```diff
@@ -692,5 +693,10 @@ public static T GetPropertyOrDefault<T>(this A11yElement e, int propertyId)
 
             return a != null;
         }
+
+        public static string WithParameters(this string formatString, params object[] args)
+        {
+            return string.Format(CultureInfo.InvariantCulture, formatString, args);
```
DaveTryon, line 699: We are likely to have some formats that will need InvariantCulture, while others will need CurrentCulture. That suggests 2 extension methods, plus an audit to select the appropriate method. Is this in scope for this PR or is that a future task?

src/Core/Bases/A11yPattern.cs
```diff
@@ -144,7 +144,7 @@ static bool IsNumber(TypeCode code)
 
         public override string ToString()
         {
-            return string.Format(CultureInfo.InvariantCulture, "{0}: {1}", this.Name, this.Properties.FirstOrDefault()?.Value);
+            return ExtensionMethods.WithParameters("{0}: {1}", Name, Properties.FirstOrDefault()?.Value);
```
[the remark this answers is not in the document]
sfoslund: That was my thought- that they're only formatting with no actual words so not worth moving

src/Core/Misc/ExtensionMethods.cs
sfoslund: I'd prefer to make that a separate PR, as this one is already so large. What do you think the determining factor is for if it needs to be invariant or not?

src/Core/Bases/A11yElement.cs
sfoslund: Those instances can't use the normal syntax because they're dynamic objects, see: https://riptutorial.com/csharp/example/3775/extension-methods-aren-t-supported-by-dynamic-code-

src/Core/Bases/A11yProperty.cs
sfoslund: I'll factor this out into a new resource then, and it sounds like we will have to add this to the list of projects to localize. There are several other short strings that I'm seeing in this project that I had assumed wouldn't be displayed to the user, so I hadn't prioritized but it sounds like we should reconsider that. Do you know if any of these are exposed to the user?

https://github.com/microsoft/axe-windows/blob/main/src/Core/Bases/A11yProperty.cs#L101
https://github.com/microsoft/axe-windows/blob/main/src/Core/CustomObjects/Converters/EnumTypeConverter.cs#L22
https://github.com/microsoft/axe-windows/blob/main/src/Core/CustomObjects/CustomProperty.cs#L28

src/Core/Misc/ExtensionMethods.cs
DaveTryon: Makes sense. As I understand it, our current bar is that it needs to be CurrentCulture if it's in the data that comes back from the `IScanner` interface. Messages for `AxeWindowsAutomationException` should presumably be localized and use CurrentCulture, since they're for consuming developers. Other Exception messages are for us and are probably out of scope for localization or CurrentCulture.

src/Core/Bases/A11yProperty.cs
DaveTryon: I would guess that these are all in scope for loc

DaveTryon (approved):
This looks like a good start. I think we'll need to tune it either way moving forward...

src/Core/Misc/ExtensionMethods.cs
sfoslund: Gotcha. I made the change in this PR by searching/replacing CurrentCulture.InvariantCulture, so this shouldn't touch any of the CurrentCulture instances at the moment and I think it makes sense to address that in a separate PR. Maybe it makes sense to have different extension methods for the different scenarios:

- An extension for exceptions, that will always use InvariantCulture
- An extension for AxeWindowsAutomatedExceptions, which may use either Current or Invariant culture depending on if we end up deciding to localize
- And an extension for message exposed to the user

What do you think?

src/Core/Bases/A11yProperty.cs
sfoslund: Okay, I'll make a separate PR to extract these out into a resx file then, and add the core project to our list of projects to be localized

PR #771, opened 2022-10-19
feat(localization): Remove instances of FormattableString

sfoslund:
#### Details

Splitting PR feedback into a separate PR, since these changes touch so many files: https://github.com/microsoft/axe-windows/pull/768#issuecomment-1284207447

##### Motivation

Localization feature

#### Pull request checklist

- [x] Addresses an existing issue: [2000537](https://dev.azure.com/mseng/1ES/_workitems/edit/2000537)

### DIFF
--- a/src/Rules/Extensions/ExtensionMethods.cs
+++ b/src/Rules/Extensions/ExtensionMethods.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using Axe.Windows.Core.Bases;
 using Axe.Windows.Core.Enums;
@@ -8,7 +8,6 @@ using Axe.Windows.Rules.PropertyConditions;
 using System.Diagnostics;
 using System.Drawing;
 using UIAutomationClient;
-using static System.FormattableString;
 
 namespace Axe.Windows.Rules.Extensions
 {
@@ -50,7 +49,6 @@ namespace Axe.Windows.Rules.Extensions
                 var stopwatch = Stopwatch.StartNew();
                 bool retVal = c.Matches(e);
                 var ms = stopwatch.ElapsedTicks / (float)10000;
-                Debug.WriteLine(Invariant($"{ms} ms {message}"));
                 return retVal;
             });
         }
--- a/src/Automation/ScanResultsAssembler.cs
+++ b/src/Automation/ScanResultsAssembler.cs
@@ -10,8 +10,6 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 
-using static System.FormattableString;
-
 namespace Axe.Windows.Automation
 {
     /// <summary>
--- a/src/Core/Bases/A11yElement.cs
+++ b/src/Core/Bases/A11yElement.cs
@@ -15,8 +15,6 @@ using System.IO;
 using System.Linq;
 using System.Text;
 
-using static System.FormattableString;
-
 namespace Axe.Windows.Core.Bases
 {
     /// <summary>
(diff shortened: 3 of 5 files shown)


DaveTryon (commented):
Interpolated strings still need the call to `Invariant`, or else we get errors in TSA. The interpolated strings should be replaced with calls to `String.Format`, either with local strings or with resource strings

src/Automation/OutputFileHelper.cs
```diff
@@ -100,7 +99,7 @@ private string GenerateScanId()
         {
             var now = _dateTime.Now;
 
-            var nowString = Invariant($"{now:yy-MM-dd_HH-mm-ss.fffffff}");
+            var nowString = $"{now:yy-MM-dd_HH-mm-ss.fffffff}";
```
DaveTryon, line 102: This string should not be localized since it's formatting in a modified ISO format

DaveTryon: We can keep `Invariant` or we can use `String.Format`. TSA gets unhappy if we use string interpolation without using `Invariant`. The root issue is that string intrerpolation returns a [FormattableString](https://learn.microsoft.com/en-us/dotnet/api/system.formattablestring?view=netstandard-2.0), which then gets converted to a `string` via an implicit call to `ToString`. This `ToString` call doesn't specify a Culture, so it's flagged as a loc problem. The `Invariant` method (see https://referencesource.microsoft.com/#mscorlib/system/FormattableString.cs,67) is literally just a wrapper that calls `ToString` with `CultureInfo.InvariantCulture` after a null check.

I guess my question might be more along the lines of whether this should be hardcoded in the code or moved to a non-localized resource string

sfoslund: As it looks like this is only used as a file name, my instinct would be not to localize, so I'll add this instance of Invariant back.

sfoslund:
> Interpolated strings still need the call to Invariant, or else we get errors in TSA. The interpolated strings should be replaced with calls to String.Format, either with local strings or with resource strings

https://github.com/microsoft/axe-windows/pull/774 will remove some of the interpolated strings, so it might be a good idea to merge that one first then come back and fix this one up after that

src/Automation/OutputFileHelper.cs
sfoslund: @DaveTryon I've updated this PR such that the instances of string interpolation that definitely don't need to be localized are left untouched, but any instances I wasn't sure about are removed and moved into resx files if necessary.

src/Rules/Conditions/TreeDescentCondition.cs
```diff
@@ -33,7 +33,7 @@ public override bool Matches(IA11yElement e)
 
         public override string ToString()
         {
-            return Invariant($"{this.ParentCondition} / {this.ChildCondition}");
+            return string.Format(CultureInfo.CurrentCulture, ConditionDescriptions.TreeDescentConditionFormat, this.ParentCondition, this.ChildCondition);
```
DaveTryon, line 36: @codeofdusk brought up an interesting point about the conditions. If we can successfully argue that they're used as a domain-specific language, then the would no longer be in scope for localization. A corollary of that might be that this becomes non-localizable and should use CultureInfo.InvariantCulture

src/Rules/Extensions/ExtensionMethods.cs
```diff
@@ -50,7 +51,7 @@ public static Condition WithTimer(this Condition c, string message)
                 var stopwatch = Stopwatch.StartNew();
                 bool retVal = c.Matches(e);
                 var ms = stopwatch.ElapsedTicks / (float)10000;
-                Debug.WriteLine(Invariant($"{ms} ms {message}"));
+                Debug.WriteLine(string.Format(CultureInfo.CurrentCulture, ConditionDescriptions.MillisecondsFormat, ms, message));
```
DaveTryon, line 54: This was originally added for debugging. We should probably just remove the line

DaveTryon (approved):

PR #772, opened 2022-10-19
feat(async-api): Refactor A11yAutomation class

DaveTryon:
#### Details

In #767, we agreed to revisit refactoring the `A11yAutomation` class to avoid the possibility of the `DesktopDataContext` referencing a different `A11yAutomation` object. This PR does the following:
1. Removes the unused `GetFocusedElement` method (I confirmed it's not used by AIWin)
2. Makes the `GetAppElement` and `NormalizedElementFromPoint` static to avoid the chance their `DesktopDataContext` parameters refer to a different `A11yAutomation` object
3. Adds necessary internal checks to guard against a null `DesktopDataContext`
4. Removes the now-superfluous `EnsureContextConsistency` method
5. Updates the calling code
6. Tightens function visibility on the `A11yAutomation` object.

##### Context

I realized that the `TargetElementLocator` in the Automation assembly has a direct dependency on `A11yAutomation.ElementsFromProcessId`. That sort of violates our intended encapsulation of Automation doing everything through Actions, but the fix there is to refactor the `TargetElementLocator` class, which feels like a separate PR

#### Pull request checklist


src/Desktop/UIAutomation/A11yAutomation.cs
```diff
@@ -110,11 +110,11 @@ private static void ReleaseElements(IUIAutomationTreeWalker walker, params IList
         /// Get DesktopElements based on Process Id.
         /// </summary>
         /// <param name="pid"></param>
-        /// <param name="dataContext">The context we're workiing with</param>
+        /// <param name="dataContext">The data context</param>
         /// <returns>return null if we fail to get elements by process Id</returns>
         public IEnumerable<DesktopElement> ElementsFromProcessId(int pid, DesktopDataContext dataContext)
```
sfoslund, line 115: Should this be static now? It feels weird to me that we would have a nonstatic method that could be dealing with UIAutomation objects that don't match

DaveTryon: Yes it should. I missed that one. Fixed now

sfoslund (approved):

PR #775, opened 2022-10-19
chore: Fix malformed rule description

DaveTryon:
#### Details

We noticed that the description for the `LocalizedLandmarkTypeExcludesPrivateUnicodeCharacters` was malformed. It read:

The [StringProperty not set] property must not contain any characters in the private Unicode range.

It should read:

The LocalizedControlType property must not contain any characters in the private Unicode range.

Easy fix, just add the description to the property and rebuild RulesDescription.md

##### Motivation

Docs should be correct

#### Pull request checklist

### DIFF
--- a/src/Rules/Resources/ConditionDescriptions.resx
+++ b/src/Rules/Resources/ConditionDescriptions.resx
@@ -245,4 +245,7 @@
   <data name="IsDialog" xml:space="preserve">
     <value>IsDialog</value>
   </data>
+  <data name="LocalizedLandmarkType" xml:space="preserve">
+    <value>LocalizedLandmarkType</value>
+  </data>
 </root>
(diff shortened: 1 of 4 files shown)


src/Rules/Resources/ConditionDescriptions.resx
```diff
@@ -245,4 +245,7 @@
   <data name="IsDialog" xml:space="preserve">
     <value>IsDialog</value>
   </data>
+  <data name="LocalizedLandmarkType" xml:space="preserve">
+    <value>LocalizedLandmarkType</value>
+  </data>
 </root>
```
codeofdusk, line 251: Should this file have a final newline?

codeofdusk (approved):

src/Rules/Resources/ConditionDescriptions.resx
DaveTryon: It was edited via the IDE, so I'm assuming that the file formatting is normalized

PR #782, opened 2022-10-24
feat(localization): Add translator comments for Core namespace

codeofdusk:
#### Details
This PR adds translator comments for each string in the resource files under the `Core` directory.

##### Motivation
Part of localization feature (internal access required to view).

##### Context
Other namespaces to follow in subsequent PRs.

#### Pull request checklist

### DIFF
--- a/src/Core/Resources/DisplayStrings.resx
+++ b/src/Core/Resources/DisplayStrings.resx
@@ -119,7 +119,7 @@
   </resheader>
   <data name="BoundingRectangleFormat" xml:space="preserve">
     <value>[l={0},t={1},w={2},h={3}]</value>
-    <comment>Do not translate: the coordinate specification of a bounding rectangle for an accessibility element.</comment>
+    <comment>The coordinate specification of a bounding rectangle for an accessibility element. l refers to the left position, t to the top position, w to the width, and h to the height. When translating, please keep these descriptions to one or two characters for display purposes.</comment>
   </data>
   <data name="HorizontalOrientation" xml:space="preserve">
     <value>Horizontal(1)</value>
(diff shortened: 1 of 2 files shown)


src/Core/Resources/DisplayStrings.resx
```diff
@@ -119,23 +119,30 @@
...
     <value>[l={0},t={1},w={2},h={3}]</value>
+    <comment>Do not translate: the coordinate specification of a bounding rectangle for an accessibility element.</comment>
   </data>
   <data name="HorizontalOrientation" xml:space="preserve">
     <value>Horizontal(1)</value>
+    <comment>A description of the visual orientation of an accessibility element.</comment>
   </data>
   <data name="LeftTopRightBottomFormat" xml:space="preserve">
     <value>[l={0},t={1},r={2},b={3}]</value>
+    <comment>Do not translate: the coordinate specification of the position of an accessibility element.</comment>
```
sfoslund, line 130: Would it make sense to use full words here (for example, `left` instead of `l`) so that they can be translated? Same comment for the BoundingRectangleFormat string.

codeofdusk: It might – NVDA uses full words in its `location` property (closest analog).

src/Core/Resources/DisplayStrings.resx
```diff
@@ -119,23 +119,30 @@
...
     <value>[l={0},t={1},r={2},b={3}]</value>
+    <comment>The coordinate specification of the position of an accessibility element. l refers to the left position, t to the top position, r to the right position, and b to the bottom position. When translating, please keep these descriptions to one or two characters for display purposes.</comment>
   </data>
   <data name="NoneOrientation" xml:space="preserve">
     <value>None(0)</value>
+    <comment>A description of the visual orientation of an accessibility element.</comment>
   </data>
   <data name="PointFormat" xml:space="preserve">
     <value>[x={0},y={1}]</value>
+    <comment>Do not translate: the coordinate specification of the position of an accessibility element.</comment>
```
sfoslund, line 138: Do we want to add a similar note here to what we have for `BoundingRectangleFormat` and `LeftTopRightBottomFormat`?

sfoslund (approved):
One more small comment but otherwise looks good!

src/Core/Resources/DisplayStrings.resx
codeofdusk: I feel like `x` and `y` are pretty universal?

sfoslund: That's probably true for languages that share an alphabet with english, but I'm wondering if there are different norms for languages what use different alphabets

Issue #780, opened 2022-10-11
Handle BoundingRectangleSizeReasonable rule for elements with area of less than 25 pixels


dilyantraykov:
Hello,

In our application, we have rectangles (displayed by a sparkline control) that have an area of fewer than 25 pixels (for example, 12x2 pixels).

![image](https://user-images.githubusercontent.com/5472516/195145577-69ec200a-f74f-4773-893b-16475bc28b17.png)

What would be the suggested approach for handling these elements so that they're not marked as problematic by the Accessibility Insights for Windows tool?

Thank you in advance for your guidance on the matter.

DaveTryon:
In today's triage discussion, we agreed to update the `Condition` of the `BoundingRectangleSizeReasonable` rule to omit WPF controls if their ItemStatus property includes the text `<Property Name="DataContext" Value="Telerik.Windows.Controls.Sparklines.SparklineColumnDataPoint" />`

chingucoding:
@DaveTryon @asksep Mind if I work in this issue?

RobGallo:
@chingucoding,
You're welcome to work on this issue :)

DaveTryon:
This is fixed but we need to keep the bug in the resolved state until we actually release the updated package

DaveTryon:
@dilyantraykov, this has been fixed in Axe.Windows 2.1.0, released today. Please close out at your convenience. Thanks!

dilyantraykov:
Hello team,

Please excuse me for the late response and thank you for the effort in creating a fix.

Although this works in our particular scenario, there are other elements that also exhibit such errors which have a different ItemStatus code. I assume that other vendors may also have such issues with similar elements but their ItemStatus values will differ.

Would it be possible to expose some property or other API to mark certain elements to be ignored for checks on a particular rule?

DaveTryon:
@dilyantraykov, I just saw your request. Adding an API is probably a non-starter. The best way to handle these would be to create a new issue for each class that would potentially fall into this category, then we'll need to evaluate them on a case-by-case basis.

PR #785, opened 2022-10-25
fix: Make Telerik Sparkline elements not fail the Bounding Rectangle rule anymore

chingucoding:
#### Details

Updating the BoundingRectangleReasonableSized rule to exclude the Telerik Sparkline elements.

##### Motivation

Addresses #780

##### Context

N/A

#### Pull request checklist

- [x] Addresses an existing issue: #780

### DIFF
--- a/src/Rules/Library/BoundingRectangleSizeReasonable.cs
+++ b/src/Rules/Library/BoundingRectangleSizeReasonable.cs
@@ -37,7 +37,7 @@ namespace Axe.Windows.Rules.Library
                 & BoundingRectangle.NotNull
                 & BoundingRectangle.CorrectDataFormat
                 & ~ignoreableText
-                & ~BoundingRectangle.NotTelerikSparklineItemstatusContext;
+                & ~BoundingRectangle.TelerikSparklineItemstatusContext;
         }
     } // class
 } // namespace
--- a/src/Rules/Library/BoundingRectangleSizeReasonable.cs
+++ b/src/Rules/Library/BoundingRectangleSizeReasonable.cs
@@ -26,7 +26,7 @@ namespace Axe.Windows.Rules.Library
 
         public override bool PassesTest(IA11yElement e)
         {
-            if(e.ItemStatus.Contains("<Property Name=\"DataContext\" Value=\"Telerik.Windows.Controls.Sparklines.SparklineColumnDataPoint\" />"))
+            if (e?.ItemStatus?.Contains("<Property Name=\"DataContext\" Value=\"Telerik.Windows.Controls.Sparklines.SparklineColumnDataPoint\" />") == true)
             {
                 return true;
             }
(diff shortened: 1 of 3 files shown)


src/Rules/Library/BoundingRectangleSizeReasonable.cs
```diff
@@ -26,6 +26,10 @@ public BoundingRectangleSizeReasonable()
 
         public override bool PassesTest(IA11yElement e)
         {
+            if (e?.ItemStatus?.Contains("<Property Name=\"DataContext\" Value=\"Telerik.Windows.Controls.Sparklines.SparklineColumnDataPoint\" />") == true)
```
DaveTryon, line 29: @chingucoding, is there a reason not to put this in the Condition as we called out as the plan in the issue?

chingucoding: Nope not at all, I somehow forgot that.

src/Rules/PropertyConditions/BoundingRectangle.cs
```diff
@@ -74,5 +75,11 @@ private static bool ElementCompletelyObscuresContainer(IA11yElement e)
 
             return e.BoundingRectangle.CompletelyObscures(container.BoundingRectangle);
         }
+        private static bool ElementItemStatusContainsTelerikSparklineContext(IA11yElement e)
```
DaveTryon, line 78: (nit): Please add a newline between the methods

src/Rules/Library/BoundingRectangleSizeReasonable.cs
```diff
@@ -36,7 +36,8 @@ protected override Condition CreateCondition()
             return IsNotOffScreen
                 & BoundingRectangle.NotNull
                 & BoundingRectangle.CorrectDataFormat
-                & ~ignoreableText;
+                & ~ignoreableText
+                & ~BoundingRectangle.TelerikSparklineItemstatusContext;
```
DaveTryon, line 40: Can we leverage `PropertyCOnditions.WPF` here, as opposed to burying it in the BoundingRectangle Concition? I _think_ this would be
```suggestion
                & ~(WPF & BoundingRectangle.TelerikSparklineItemstatusContext);
```

src/Rules/PropertyConditions/BoundingRectangle.cs
```diff
@@ -74,5 +75,11 @@ private static bool ElementCompletelyObscuresContainer(IA11yElement e)
...
+        {
+            if (e == null) throw new ArgumentNullException(nameof(e));
+
+            return e.Framework == "WPF" && e.ItemStatus?.Contains("<Property Name=\"DataContext\" Value=\"Telerik.Windows.Controls.Sparklines.SparklineColumnDataPoint\" />") == true;
```
DaveTryon, line 82: If we apply the WPF check at the rule, this would become
```suggestion
            return e.ItemStatus?.Contains("<Property Name=\"DataContext\" Value=\"Telerik.Windows.Controls.Sparklines.SparklineColumnDataPoint\" />") == true;
```

src/RulesTest/Library/BoundingRectangleSizeReasonableTests.cs
```diff
@@ -32,6 +32,18 @@ public void TestBoundingRectangleSizeReasonableEmptyFail()
                 Assert.IsFalse(Rule.PassesTest(e));
             } // using
         }
+        
+        [TestMethod]
```
DaveTryon, line 36: Maybe add a test where the ItemStatus check passes but the Framework check fails? Won't impact code coverage but just to pin the intent

chingucoding: Good idea, done

DaveTryon (approved):
Thanks, @chingucoding! Looks good

PR #787, opened 2022-10-26
feat(localization): Add script to check resource strings for comments

sfoslund:
#### Details

As the resource strings in the core, rules, and ruleselection projects will be localized soon, it is important that we include any necessary translator comments along with those resources. We have gone through and added comments in previous PRs, but this PR adds a script to our PR builds to verify that we continue to have translator notes for all resource strings in those projects.

##### Motivation

Localization feature

##### Context

The script does not include the resource files for projects that are not localized right now, since there are quite a lot of strings in those projects that do not have comments yet. If we ever decide to localize those projects, it will be simple to remove them from the exclude list in this script. 

#### Pull request checklist

- [x] Addresses an existing issue: [2002611](https://dev.azure.com/mseng/1ES/_workitems/edit/2002611)

### DIFF
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -36,7 +36,7 @@ jobs:
       arguments: '-target  $(Build.Repository.LocalPath) -licenseHeaderPath tools\scripts\verification.scripts\LicenseHeader.txt -extensions *.xaml,*.xml,*.cs,*.ps1 -addIfAbsent $false'
 
   - task: PowerShell@2
-    displayName: 'Resource String Comment Check'
+    displayName: 'Resource String Translator Comment Check'
     inputs:
       targetType: "filePath"
       filePath: tools\scripts\verification.scripts\ResourceStringCommentVerification.ps1
--- a/build/prbuild.yml
+++ b/build/prbuild.yml
@@ -35,6 +35,13 @@ jobs:
       filePath: tools\scripts\verification.scripts\LicenseHeaderVerification.ps1
       arguments: '-target  $(Build.Repository.LocalPath) -licenseHeaderPath tools\scripts\verification.scripts\LicenseHeader.txt -extensions *.xaml,*.xml,*.cs,*.ps1 -addIfAbsent $false'
 
+  - task: PowerShell@2
+    displayName: 'Resource String Comment Check'
+    inputs:
+      targetType: "filePath"
+      filePath: tools\scripts\verification.scripts\ResourceStringCommentVerification.ps1
+      arguments: '-target  $(Build.Repository.LocalPath)'
+
   - task: VSBuild@1
     displayName: 'Build Solution **\*.sln'
     inputs:
--- a/src/Rules/Resources/ErrorMessages.resx
+++ b/src/Rules/Resources/ErrorMessages.resx
@@ -155,7 +155,7 @@
   </data>
   <data name="MissingRuleInforAttribute" xml:space="preserve">
     <value>expected {0} to have the RuleInfo attribute set</value>
-    <comment>Do not translate: the text of an internal error. {0} is type name</comment>
+    <comment>Do not translate: the text of an internal error. Developers: {0} is type name</comment>
   </data>
   <data name="NoControlTypeEntryFound" xml:space="preserve">
     <value>No control type entry was found in dictionary. Key value = {0}</value>
@@ -171,7 +171,7 @@
   </data>
   <data name="NoRuleMatchingId" xml:space="preserve">
     <value>No rule matching the ID '{0}' was found.</value>
-    <comment>Do not translate: the text of an internal error. {0} is rule id</comment>
+    <comment>Do not translate: the text of an internal error. Developers: {0} is rule id</comment>
   </data>
   <data name="RuleInfoAttributeExpected" xml:space="preserve">
     <value>All rules are expected to have the RuleInfo attribute with the ID property set.</value>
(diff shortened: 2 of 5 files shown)


build/prbuild.yml
```diff
@@ -35,6 +35,13 @@ jobs:
       filePath: tools\scripts\verification.scripts\LicenseHeaderVerification.ps1
       arguments: '-target  $(Build.Repository.LocalPath) -licenseHeaderPath tools\scripts\verification.scripts\LicenseHeader.txt -extensions *.xaml,*.xml,*.cs,*.ps1 -addIfAbsent $false'
 
+  - task: PowerShell@2
+    displayName: 'Resource String Comment Check'
```
codeofdusk, line 39:
```suggestion
    displayName: 'Resource String Translator Comment Check'
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -155,7 +155,7 @@
   </data>
   <data name="MissingRuleInforAttribute" xml:space="preserve">
     <value>expected {0} to have the RuleInfo attribute set</value>
-    <comment>{0} is type name</comment>
+    <comment>Do not translate: the text of an internal error. {0} is type name</comment>
```
codeofdusk, line 158:
```suggestion
    <comment>Do not translate: the text of an internal error. Developers: {0} is type name</comment>
```

src/Rules/Resources/ErrorMessages.resx
```diff
@@ -171,10 +171,11 @@
   </data>
   <data name="NoRuleMatchingId" xml:space="preserve">
     <value>No rule matching the ID '{0}' was found.</value>
-    <comment>{0} is rule id</comment>
+    <comment>Do not translate: the text of an internal error. {0} is rule id</comment>
```
codeofdusk, line 174:
```suggestion
    <comment>Do not translate: the text of an internal error. Developers: {0} is rule id</comment>
```

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
```diff
@@ -0,0 +1,68 @@
...
+#>
+
+param(
+    $Target
+)
+
+$NewLine=([Environment]::NewLine)
+$FailedStrings = @()
+# Exclude the projects that are not yet 
+$excludeList="(Actions|Automation|CLI|Desktop|tools)"
```
codeofdusk, line 21: Might it be better to have a whitelist, explicitly **including** namespaces that we're localizing? Can we keep this list in one place that everything else can reference?

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
```diff
@@ -0,0 +1,68 @@
...
+
+    for($i=0; $i -lt $resourceStrings.Length; $i++) {
+        if ($resourceStrings[$i] -Match "<data name=") {
+            # Account for any multi-line resource values
+            $j = $i + 1;
+            while ($resourceStrings[$j] -NotMatch "</value>") {
+                $j = $j + 1
+            }
+
+            # There should be a comment directly after the resource value
```
codeofdusk, line 50: Might it also occur before?

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+
+    echo "Finished scanning $path"
+}
+
+echo "$NewLine"
+if($FailedStrings.Count -gt 0){
+    $message = "The following resources do not have comments $NewLine" + ($FailedStrings -join $NewLine) + "$NewLine"
```
codeofdusk, line 64:
```suggestion
    $message = "The following resources do not have translator comments $NewLine" + ($FailedStrings -join $NewLine) + "$NewLine"
```

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
```diff
@@ -0,0 +1,68 @@
...
+    throw ($message)
+} else {
+    echo "All string resources have comments"
```
codeofdusk, line 67:
```suggestion
    echo "All string resources have translator comments"
```

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
sfoslund: I haven't seen that before, and it definitely doesn't exist in our repo. It's probably not explicitly disallowed by the .resx  format, but IMO I think it's reasonable to assume that the comment will always come after as an invariant for our files.

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
sfoslund: My thought with having an exclude list is that, if/when we add new namespaces in the future, they will automatically be included in this check. This may or may not be the behavior we want for new namespaces, but it would force us to make an explicit decision, as opposed to just forgetting about translator comments for a new namespace altogether. 

> Can we keep this list in one place that everything else can reference?

Where else would we be referencing this?

codeofdusk: > My thought with having an exclude list is that, if/when we add new namespaces in the future, they will automatically be included in this check.

Makes sense to me.

> Where else would we be referencing this?

In whatever automatically builds the localization catalogues.

codeofdusk (approved):

tools/scripts/verification.scripts/ResourceStringCommentVerification.ps1
codeofdusk: I think PowerShell might have support for parsing XML? If so, you might be able to make this more robust.

PR #790, opened 2022-10-27
feat(localization): initial how to fix fixes

codeofdusk:
#### Details
This PR adds more meaningful "how to fix" text for some rules in place of just repeating the description.

##### Motivation
Part of localization feature (internal access required to view).

##### Context
Remaining rules to follow, these are just the ones that I felt confident fixing.

#### Pull request checklist

### DIFF
--- a/src/Rules/Resources/HowToFix.resx
+++ b/src/Rules/Resources/HowToFix.resx
@@ -177,7 +177,7 @@
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableForCustomShouldBeTrue" xml:space="preserve">
-    <value>Insure that this element can be manipulated via the keyboard, then set the IsKeyboardFocusable UI Automation property to true.</value>
+    <value>Ensure that this element can be manipulated via the keyboard, then set the IsKeyboardFocusable UI Automation property to true.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableDescendantTextPattern" xml:space="preserve">
--- a/src/Rules/Resources/HowToFix.resx
+++ b/src/Rules/Resources/HowToFix.resx
@@ -149,23 +149,23 @@
     <comment>Brief guidance on how to fix an accessibility problem. The term "HelpText" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsContentElementFalseOptional" xml:space="preserve">
-    <value>If this element does not contain data that is presented to the user (such as items in a list box or a button in a modal dialog box), set the IsContentElement UI Automation property to false.</value>
+    <value>Set the IsContentElement UI Automation property to false.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsContentElement" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsContentElementTrueOptional" xml:space="preserve">
-    <value>If this element contains data that is presented to the user (such as items in a list box or a button in a modal dialog box), set the IsContentElement UI Automation property to true.</value>
+    <value>Set the IsContentElement UI Automation property to true.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsContentElement" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsControlElementTrueOptional" xml:space="preserve">
-    <value>If this element represents an interactive UI control, set the IsControlElement UI Automation property to true.</value>
+    <value>Set the IsControlElement UI Automation property to true.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsControlElement" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableShouldBeTrue" xml:space="preserve">
-    <value>If this element represents a control that is reachable via the keyboard, set the IsKeyboardFocusable UI Automation property to true.</value>
+    <value>Set the IsKeyboardFocusable UI Automation property to true.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableFalseButDisabled" xml:space="preserve">
-    <value>If this element is still reachable via the keyboard even when it is disabled (dimmed), set the IsKeyboardFocusable UI Automation property to true.</value>
+    <value>If this element is still reachable via the keyboard even when it is disabled (unavailable), set the IsKeyboardFocusable UI Automation property to true.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableForListItemShouldBeTrue" xml:space="preserve">
@@ -185,8 +185,8 @@
     <comment>The description of a problem with an element currently under inspection. The terms "IsKeyboardFocusable" and "Text" should not be translated (i.e. they should be treated as loanwords).</comment>
   </data>
   <data name="IsKeyboardFocusableOnEmptyContainer" xml:space="preserve">
-    <value>If this container should be discoverable by keyboard users, set the IsKeyboardFocusable UI Automation property to true.</value>
-    <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>If this container should be discoverable by keyboard users, set the IsKeyboardFocusable UI Automation property to true. Otherwise, consider changing the ControlType.</value>
+    <comment>Brief guidance on how to fix an accessibility problem. The terms "IsKeyboardFocusable" and "ControlType" should not be translated (i.e. they should be treated as loanwords).</comment>
   </data>
   <data name="IsKeyboardFocusableShouldBeFalse" xml:space="preserve">
     <value>Set the IsKeyboardFocusable UI Automation property to false. If this element really should be reachable via the keyboard, verify its ControlType.</value>
--- a/src/Rules/Resources/HowToFix.resx
+++ b/src/Rules/Resources/HowToFix.resx
@@ -165,7 +165,7 @@
     <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableFalseButDisabled" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


sfoslund (commented):
These look good to me, but I think @ferBonnin should be the one to give final approval

src/Rules/Resources/HowToFix.resx
```diff
@@ -122,12 +122,15 @@
     <comment>The description of a problem with an element currently under inspection. The term "BoundingRectangle" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="ButtonWithSplitButtonParentPattern" xml:space="preserve">
-    <value>A button element should only support one of Invoke, Toggle, or ExpandCollapse patterns when a splitbutton is the parent. </value>
-    <comment>The description of a problem with an element currently under inspection. The terms "Invoke", "Toggle", and "ExpandCollapse" should not be translated (i.e. they should be treated as loanwords).</comment>
+    <value>Modify the button to support exactly one of the following patterns:
```
ferBonnin, line 125: the original value provided relevant information as to why the rule was failing. (e.g. A button element should only support one of Invoke, Toggle, or ExpandCollapse patterns when a splitbutton is the parent. ) Love the new how to fix information but wondering if we should keep the _why it failed_ information as well. Also, these new instructions are longer than the previous ones, how does this impact the UI in Accessibility Insights for Windows?

sfoslund: > wondering if we should keep the why it failed information as well.

Agree on this point!

> These new instructions are longer than the previous ones, how does this impact the UI in Accessibility Insights for Windows?

I don't think we need to worry about that, considering some of our existing how to fix strings are already quite long (for example, https://github.com/microsoft/axe-windows/blob/main/src/Rules/Resources/HowToFix.resx#L251) and they show up well in AI for windows:

[Screenshot of long how to fix string rendered in AIWin]
HowToFixScreenshot

sfoslund: > wondering if we should keep the why it failed information as well.

Actually, I think @codeofdusk might have mentioned that part of the motivation to replace these strings was that they are duplicated, and it does look like the original strings here do also exist in the rules descriptions. @codeofdusk please confirm, but if that is the case then I don't think we need to keep the why it failed information in these strings as well.

codeofdusk: Correct – these strings were duplicated from the rule descriptions.

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
-    <comment>The description of a problem with an element currently under inspection. The terms "IsKeyboardFocusable" and "IsOffscreen" should not be translated (i.e. they should be treated as loanwords).</comment>
+    <value>If this element is still reachable via the keyboard even when it is off-screen, set the IsKeyboardFocusable UI Automation property to true.</value>
+    <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableForCustomShouldBeTrue" xml:space="preserve">
-    <value>The IsKeyboardFocusable property for a custom element should be true when the element supports actionable patterns.</value>
-    <comment>The description of a problem with an element currently under inspection. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>Insure that this element can be manipulated via the keyboard, then set the IsKeyboardFocusable UI Automation property to true.</value>
```
sfoslund, line 180: nit: Should be 'Ensure'

src/Rules/Resources/HowToFix.resx
ferBonnin: got it! will the user still have access to the strings in all views (automated checks and inspect tree)?

sfoslund: @ferBonnin, here's the zoomed out picture of my screenshot above. You can see that the how to fix strings (the ones being changed in this PR) are at the bottom in the 'fix the following' section and the rule descriptions (which describe why the failure occurred) are still above and visible in the 'test results' section.

[Screenshot of full how to fix page in AIWin]
LargerHowToFix

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
+    <value>Write meaningful text for the element's UI Automation HelpText property. This text is typically similar to, or taken from, a tooltip or placeholder, such as "type here to search".</value>
+    <comment>Brief guidance on how to fix an accessibility problem. The term "HelpText" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsContentElementFalseOptional" xml:space="preserve">
-    <value>The recommended value of the IsContentElement property for the given control type is false. Please consider if this is an element that should be reported to an assistive technology user as content.</value>
-    <comment>The description of a problem with an element currently under inspection. The term "IsContentElement" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>If this element does not contain data that is presented to the user (such as items in a list box or a button in a modal dialog box), set the IsContentElement UI Automation property to false.</value>
```
ferBonnin, line 152: There is a bit of mismatch on how we write this sentences. For some of them, we are saying the action but for others we are saying _If_. Is there something different between one check and the other (for example an automated check vs a warning)?

DaveTryon: Looking at the rule, it might be simpler to just say something like "Set the IsContentElement UI Automation property for this element to false". The rule's condition should already filter out controls where the user doesn't need to change anything.

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
+    <comment>Brief guidance on how to fix an accessibility problem. The term "IsContentElement" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsContentElementTrueOptional" xml:space="preserve">
-    <value>The recommended value of the IsContentElement property for the given control type is true. Please consider if this is an element that should be reported to an assistive technology user as content.</value>
-    <comment>The description of a problem with an element currently under inspection. The term "IsContentElement" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>If this element contains data that is presented to the user (such as items in a list box or a button in a modal dialog box), set the IsContentElement UI Automation property to true.</value>
```
DaveTryon, line 156: Looking at the rule, it might be simpler to just say something like "Set the IsContentElement UI Automation property for this element to true". The rule's condition should already filter out controls where the user doesn't need to change anything.

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
+    <comment>Brief guidance on how to fix an accessibility problem. The term "IsContentElement" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsControlElementTrueOptional" xml:space="preserve">
-    <value>The recommended value of the IsControlElement property for the given control type is true. Please consider if this is an element that should be reported to an assistive technology user as a control. Note that almost all controls are required to have the IsControl Property set to true.</value>
-    <comment>The description of a problem with an element currently under inspection. The term "IsControlElement" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>If this element represents an interactive UI control, set the IsControlElement UI Automation property to true.</value>
```
DaveTryon, line 160: Looking at the rule, it might be simpler to just say something like "Set the IsControlElement UI Automation property for this element to true". The rule's condition should already filter out controls where the user doesn't need to change anything.

DaveTryon: Similar comment for IsControlElementFalseOptional, although it's not current in this PR

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
+    <comment>Brief guidance on how to fix an accessibility problem. The term "IsControlElement" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="IsKeyboardFocusableShouldBeTrue" xml:space="preserve">
-    <value>The IsKeyboardFocusable property for the given element should be true based on its control type.</value>
-    <comment>The description of a problem with an element currently under inspection. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>If this element represents a control that is reachable via the keyboard, set the IsKeyboardFocusable UI Automation property to true.</value>
+    <comment>Brief guidance on how to fix an accessibility problem. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
```
DaveTryon, line 165: Looking at the rule, it might be simpler to just say something like "Set the IsKeyboardFocusable UI Automation property for this element to true". The rule's condition should already filter out controls where the user doesn't need to change anything.

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
   </data>
   <data name="IsKeyboardFocusableFalseButDisabled" xml:space="preserve">
-    <value>The IsKeyboardFocusable property is false for an element where it would normally be true. However, the IsEnabled property on the element is also false, so the value of IsKeyboardFocusable may be acceptable.</value>
-    <comment>The description of a problem with an element currently under inspection. The terms "IsKeyboardFocusable" and "IsEnabled" should not be translated (i.e. they should be treated as loanwords).</comment>
+    <value>If this element is still reachable via the keyboard even when it is disabled (dimmed), set the IsKeyboardFocusable UI Automation property to true.</value>
```
DaveTryon, line 168: Can we omit "dimmed" ? It adds a term that isn't defined elsewhere and could potentially lead to confusion

src/Rules/Resources/HowToFix.resx
```diff
@@ -142,68 +145,68 @@
...
     <value>The IsKeyboardFocusable property may be false when the given element supports the Text pattern and is the descendant of an element that also supports the Text pattern. Please consider if the given element should or should not be focusable.</value>
     <comment>The description of a problem with an element currently under inspection. The terms "IsKeyboardFocusable" and "Text" should not be translated (i.e. they should be treated as loanwords).</comment>
   </data>
   <data name="IsKeyboardFocusableOnEmptyContainer" xml:space="preserve">
-    <value>The IsKeyboardFocusable property should be true when you want an empty container to be discoverable by assistive technology users. IsKeyboardFocusable may be false when you want an empty container not to be discoverable by assistive technology users.</value>
-    <comment>The description of a problem with an element currently under inspection. The term "IsKeyboardFocusable" should not be translated (i.e. it should be treated as a loanword).</comment>
+    <value>If this container should be discoverable by keyboard users, set the IsKeyboardFocusable UI Automation property to true.</value>
```
DaveTryon, line 188: My problem with starting these sentences with "If" is that we generally don't provide an "else" clause. In most cases, the user action would be to change the ControlType property, which would generally only be needed for a custom control. Custom controls are relatively uncommon, so optimizing for the majority case will result in more actionable messages.

src/Rules/Resources/HowToFix.resx
codeofdusk: I'll change to "unavailable" as that's how NVDA indicates it (and I think JAWS as well).

sfoslund (approved):

DaveTryon (approved):

codeofdusk:
@ferBonnin accepted these changes offline. Merging.

PR #792, opened 2022-10-28
Use CsWin32 to generate NativeMethods

mcooley:
#### Details

Uses the [CsWin32 library](https://github.com/microsoft/cswin32) to generate P/Invoke wrappers for windows APIs that Axe.Windows calls.

##### Motivation

Eliminate the need to maintain P/Invoke method definitions, since they will now be generated based on [official metadata from the Windows team](https://github.com/microsoft/win32metadata).

##### Context

Eventually, I would like CsWin32 to generate interop code not only for these methods but also for the whole UI automation API. Ideally the .NET ecosystem would not need to maintain libraries like [UIAutomation-Interop](https://github.com/FlaUI/UIAutomation-Interop) just to wrap UIA. I've been using the Axe.Windows codebase to prototype this, and as a result I've submitted some fixes and suggestions upstream.

I'm submitting this PR to check whether Axe.Windows maintainers would be interested in using CsWin32 in general, before I go further to prepare a change to use CsWin32 for UIA.


dbjorge:
I think we should probably get a few different folks' opinions on this (@DaveTryon / @RobGallo ?). My two cents:

* I think that the concept of using an officially supported standard interop library for these instead of writing it ourselves based on community-provided versions is righteous
* However, I'm concerned about adding a dependency on a beta version of a project which [self-proclaims](https://github.com/microsoft/CsWin32/issues/698) that "we still have active bugs for defective APIs generated".
* I'm doubly concerned about a beta-stability dependency in sections of code that are especially difficult to practically unit test
* I feel like I'd need to do some more research/analysis of the generated code to feel comfortable with some of these changes. For example, I don't know enough to feel comfortable that the switch from `Marshal.SizeOf` to `sizeof` in `src/Win32/HighContrast.cs` is safe.

Overall, I'm not opposed to using this project in concept, but I would feel more comfortable waiting to accept this dependency until the project declares a stable version and has demonstrated some stable bake time with it.

chingucoding:
If you don't mind me jumping in on this discussion as a consumer of Axe.Windows : 
I see the benefit of adding a library to get rid of code maintaining these interops from a maintainer standpoint, after all it's less code to maintain. However, reading that this library still has active bugs and still is shipping as pre-release only, as a consumer I'm a bit worried of having to try and figure out these issues when Axe is hitting these bugs. Especially since Axe is running against rendered UIs which can be flaky in CI, adding this additional possible layer of uncertainty is something I would prefer avoided. As @dbjorge pointed out, since those things are also not the easiest to cover properly with Unit Tests, this seems like a risk to stability of Axe.Windows.

mcooley:
I'd like to contextualize the "CsWin32 still has active bugs" thing that folks are reading: the scope of CsWin32 is to cover _all Windows APIs_, and because of that scope I think it will be a while before the project can meaningfully declare stability. Axe.Windows uses only a few Windows APIs, and the APIs it uses are not exotic, so I expect that the code generated by CsWin32 here will be quite stable in practice. "Stable in practice" and "officially stable" are different, and it is totally reasonable for the project maintainers to say they want "officially stable"--but I want to provide some reassurance that this is not "unstable in practice" ;)

DaveTryon:
@mcooley We'll triage this in our weekly triage meeting then get back to you.

DaveTryon:
@mcooley We discussed this in our triage meeting and we don't feel that this is a good change for us at this time. The NativeMethods class may be old-school, but it never changes, is a known quantity, and works in all cases, including on Win7, which is currently our minimum required platform. Moving to CsWin32 won't improve the user experience, won't save us engineering time, and in a worse case scenario could introduce a bug that we might then have trouble tracing back to the root cause. Closing the PR.

PR #804, opened 2022-11-01
chore: Remove unused Preconditions class

DaveTryon:
#### Details

The `Preconditions` class is not used. It was originally written as a proposal that didn't get adopted, but the code stuck around and has now propagated to 2 different repos. It's time for it to disappear.

##### Motivation

Code cleanup

#### Pull request checklist


src/Core/Resources/ErrorMessages.resx
```diff
@@ -141,10 +141,6 @@
     <value>Value cannot be null</value>
     <comment>Do not translate: the text of an internal error.</comment>
   </data>
-  <data name="ValueCannotBeEmptyOrWhiteSpace" xml:space="preserve">
```
DaveTryon, line 144: @sfoslund I'm assuming that this has no impact on the loc process, but can you confirm?

sfoslund: No, I don't believe so.

PR #824, opened 2022-11-08
fix: Add MSAA role name to LegacyIAccessiblePattern.Role

DaveTryon:
#### Details

AIWin [#1454](https://github.com/microsoft/accessibility-insights-windows/issues/1454) requests that we add the enum name for MSAA role types. This change provides the mapping, and it will light up in AIWin after this change gets published and picked up.

We add the `LegacyIAccessibleRoleType` class to map from numeric values to MSAA role types. We then add this to `A11yProperty` to capture this on the element. Finally, we need to expose this via the `A11yPatternProperty`, which leverages the `LegacyIAccessibleRoleType` class to get the name. In order to avoid confusing quotation marks around the mapped role, we add `A11yPatternProperty.OmitQuotesFromString`, which gets set only for `LegacyIAccessiblePattern.Role`. When writing out the JSON file, we set the flag to omit null values, so `A11yPatterProperty.OmitQuotesFromString` only gets written to disk when it's specified.

Validated by building locally and testing against the pentest site provided in the AIWin issue.

Two notes:

1. If an `.a11ytest` file is created with the new build, then opened in a previous version of AIWin, the properties will contain the LegacyIAccessiblePattern.Role value in quotes. This is because older versions don't know about the `A11yPatternProperty.OmitQuotesFromString` property and include the quotes
2. Since AIWin displays the data stored in the `.a11ytest` file, existing `.a11ytest` files will continue to display just the number without the enum name
 
##### Motivation

Set things up to address AIWin [#1454](https://github.com/microsoft/accessibility-insights-windows/issues/1454)

##### Screenshots

Scenario | Before PR | After PR
--- | --- | ---
Live mode (pen test) | image | image
Open old saved file | image | image
Open new saved file | image | image

#### Pull request checklist

- [x] Addresses an existing issue: AIWin [#1454](https://github.com/microsoft/accessibility-insights-windows/issues/1454)

### DIFF
--- a/src/Core/Types/LegacyIAccessibleRoleType.cs
+++ b/src/Core/Types/LegacyIAccessibleRoleType.cs
@@ -99,7 +99,7 @@ namespace Axe.Windows.Core.Types
 #pragma warning restore CA1024 // Use properties where appropriate
 
         /// <summary>
-        /// private constructor since it would be singleton model
+        /// private constructor since this uses a singleton model
         /// </summary>
         private LegacyIAccessibleRoleType() : base("ROLE_SYSTEM_") { }
 
--- a/src/Core/Types/LegacyIAccessibleRoleType.cs
+++ b/src/Core/Types/LegacyIAccessibleRoleType.cs
@@ -1,0 +1,120 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text;
+
+using static System.FormattableString;
+
+namespace Axe.Windows.Core.Types
+{
+    /// <summary>
+    /// Class for LegacyIAccessible.Role Types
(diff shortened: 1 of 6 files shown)


src/Core/Types/LegacyIAccessibleRoleType.cs
```diff
@@ -0,0 +1,120 @@
...
+            {
+                TheInstance = new LegacyIAccessibleRoleType();
+            }
+
+            return TheInstance;
+        }
+#pragma warning restore CA1024 // Use properties where appropriate
+
+        /// <summary>
+        /// private constructor since it would be singleton model
```
sfoslund, line 102: nit: can we make this comment match the changes you made in https://github.com/microsoft/axe-windows/pull/823?

sfoslund (approved):

PR #834, opened 2022-11-09
chore: Add wxs validation/fix gaps

DaveTryon:
#### Details

In https://github.com/microsoft/accessibility-insights-windows/pull/1388, we added a WXS validation check to the AIWin MSI definition to identify gaps that are introduced by package updates (mainly driven by dependabot). It seemed like a good idea to have a similar check for Axe.Windows. When running the test, it identified gaps that had been introduced by package updates, so those are includes in this PR.

Note that since the MSI build is limited release builds, so are the MsiFileTests.

##### Motivation

Ensure that the CLI always has the expected set of files.

#### Pull request checklist


src/MsiFileTests/MsiFileTests.csproj
```diff
@@ -0,0 +1,26 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
sfoslund, line 4: Looks like this got missed in the merge- it should be 6.0 now right?

DaveTryon: Updated to 6.0. Curiously, it insists on using `net6.0-windows`, which wasn't the case for other projects. I'm not 100% why this is needed here but not elsewhere ☹️

sfoslund: It's because this project has `UseWPF` set to true, I implemented that feature on my last team 😆

sfoslund (approved):

PR #849, opened 2022-11-23
chore: Ignore launchSettings.json

DaveTryon:
#### Details

Current versions of VS store user-specific settings in `launchSettings.json`. This file shouldn't be in the repo, so adding it to the `.gitignore` file seems appropriate.

##### Motivation

Enable local debugging without risk of contaminating the repo with these files.

#### Pull request checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.user
 *.userosscache
 *.sln.docstates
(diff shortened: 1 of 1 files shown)


codeofdusk (approved):
LGTM

PR #870, opened 2023-01-18
chore: add automation ids to failing controls in wildlife test app

sfoslund:
#### Details

Prior to this change none of the controls in the wildlife test app had automation IDs. This doesn't really matter since it's just a test app, but it means the automation ID column in the VS accessibility checker is not populated when testing this app. This is a simple improvement to make the wildlife test app better to demo with. 

##### Motivation

Improve demo app

#### Pull request checklist


codeofdusk (approved):
Seems fine to me

Issue #836, opened 2022-11-11
[Feature Request] Make scans against Chromium-based UIs more actionable


dbjorge:
## Is your feature request related to a problem? Please describe.

This issue is based on discussion with @RobGallo and @DaveTryon after several customer interactions where users attempting to scan apps that are Chromium-based get confused.

The most common case is a user trying to scan a browser (Edge, Chrome, etc) with their page loaded, but we've also seen cases where a user has their own Windows app that embeds Chromium (eg, via Electron or WebView2).

There are two related problems that happen when a user does this:

* Most ATs special case Chromium instances such that it's a lot more useful for practical accessibility purposes to scan web content with a web scanner like Accessibility Insights for Web, rather than with axe-windows/Accessibility Insights for Windows. However, axe-windows results don't make any mention of this, even though we could easily distinguish these cases programmatically (by looking for a FrameworkId UIA property of "Chrome").
* Even with an absolutely minimal WebView2 usage serving an empty page with no issues, there are a few low-severity issues that axe-windows flags that are essentially framework issues (~10 `BoundingRectangleNotNull` violations). Users can't do anything about these, but we don't currently treat them as framework issues.

## Describe the solution you'd like

Rob and I think we'd ideally handle this on the axe-windows side similarly to how we treat old edge - create a rule which specifically detects an embedded chromium instance (via FrameworkId "Chome") and present a specific error that directly instructs users to scan the contents of the browser with a web-based scanner. Such a rule should probably somehow suppress any further errors in descendants of the root "Chrome" element - this will require some new form of handling, we don't currently have any other rules that do this type of subtree suppression.

It's a little questionable to call this an "Error" since we don't really want to treat it as an error to use a WebView2 at all, but we don't really have a better means to present actionable feedback to a user about needing an extra step for scanning such a component. I think that a single actionable error would clearly be a lot better than the current case of "dozens or hundreds of inactionable errors", though.

## Describe alternatives you've considered

* The real gold standard in terms of customer ease-of-use might be to actually perform an axe-core scan inside the detected chromium instance somehow. But we think this would probably be technically infeasible to do automatically unless the Chromium instance were accepting CDP debugger commands, and it would be a huge chunk of work to present it reasonably.
[...]

PR #871, opened 2023-01-25
feat(chrome-scans): Add `ChromiumComponentsShouldUseWebScanner` rule

sfoslund:
#### Details

This PR is 1/2 of addressing https://github.com/microsoft/axe-windows/issues/836. It includes the following commits/changes, which can be reviewed separately:

- https://github.com/microsoft/axe-windows/commit/27295d6a0c22942f26d72efe62be886ce0da5460 adds a new sample project to the tools repo that uses a minimal webview2.
- https://github.com/microsoft/axe-windows/commit/b0076947294e73f061ac56ce349979296a57a8d7 adds the `ChromiumComponentsShouldUseWebScanner` rule and a new e2e test leveraging the new web view sample added in this PR. 

##### Motivation

Addresses issue https://github.com/microsoft/axe-windows/issues/836

##### Context

This PR only adds the new rule logic but does not do the work to suppress any further errors in descendants of the root "Chrome" element (hence why there are 35 expected errors in the end to end test instead of just 1). A future PR will add the suppression logic separately. 

#### Pull request checklist

- [x] Addresses an existing issue: [#0000](https://github.com/microsoft/axe-windows/issues/836)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -26,12 +26,14 @@ public class AutomationIntegrationTests
         const int WindowsFormsControlSamplerKnownErrorCount = 6;
         const int WpfControlSamplerKnownErrorCount = 7;
         const int WindowsFormsMultiWindowSamplerAppAllErrorCount = 12;
+        const int WebViewSampleKnownErrorCount = 35;
```
codeofdusk, line 29: Maybe add a note that this will eventually be reduced to 1 error?

codeofdusk (approved):
lgtm

src/Rules/Resources/Descriptions.resx
```diff
@@ -520,4 +520,8 @@ HowToFix:    {2}
 Condition:  {3}</value>
     <comment>{0} is id, {1} is description, {2} is how to fix, and {3} is condition</comment>
   </data>
+  <data name="ChromiumComponentsShouldUseWebScanner" xml:space="preserve">
+    <value>Chromium components should be scanned with a web-based scanner.</value>
```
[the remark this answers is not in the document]
codeofdusk: The "how to fix" guidance suggests K4W.

src/Rules/Library/ChromiumComponentsShouldUseWebScanner.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Axe.Windows.Rules.Library
+{
+    [RuleInfo(ID = RuleId.ChromiumComponentsShouldUseWebScanner)]
+    class ChromiumComponentsShouldUseWebScanner : Rule
+    {
+        public ChromiumComponentsShouldUseWebScanner()
+        {
+            Info.Description = Descriptions.ChromiumComponentsShouldUseWebScanner;
+            Info.HowToFix = HowToFix.ChromiumComponentsShouldUseWebScanner;
+            Info.ErrorCode = EvaluationCode.Error;
```
DaveTryon, line 18: At one point, someone mentioned treating this as a framework issue, which would potentially mean creating a web page and linking to it. Did we ultimately decide against that approach?

sfoslund: I wasn't aware of that discussion and Fer is OOF so we won't be able to get her input for a while. Is that something that had been settled previously?

src/RulesTest/Library/EdgeBrowserHasBeenDeprecatedTests.cs
```diff
@@ -29,7 +29,7 @@ public void FrameworkIssueLink_IsNotNull()
         [TestMethod]
         public void Condition_FrameworkIsNotEdge_ReturnsFalse()
         {
-            string[] nonWin32Values = { FrameworkId.DirectUI, FrameworkId.InternetExplorer, FrameworkId.WinForm, FrameworkId.WPF, FrameworkId.XAML, FrameworkId.Win32, "NotEdge" };
+            string[] nonWin32Values = { FrameworkId.DirectUI, FrameworkId.InternetExplorer, FrameworkId.WinForm, FrameworkId.WPF, FrameworkId.XAML, FrameworkId.Win32, FrameworkId.Chrome, "NotEdge" };
```
DaveTryon, line 32: I find 3 other test classes where we take a similar approach:
* ControlShouldSupportSetInfoWPFTests.cs
* ControlShouldSupportSetInfoXAMLTests.cs
* FrameworkDoesNotSupportUIAutomationTests.cs

Perhaps a future PR could add a utility method that would have a signature along the lines of
```
IEnumerable<string> GetFilteredFrameworks(string excludeThis, string addThis);
```
and we could use that method in all of them. Assuming that method were to use reflection to enumerate the values, adding a fields added to the `FrameworkId` class would automatically get picked up in the test classes.

tools/WebViewSample/src/WebViewSample.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>WinExe</OutputType>
+    <TargetFramework>net7.0-windows</TargetFramework>
```
DaveTryon, line 5: We're using net6.0 in all other tests. For consistency, should we also use it here? According to https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core, net6.0 is LTS, while net7.0 is not.

src/RulesTest/Library/EdgeBrowserHasBeenDeprecatedTests.cs
sfoslund: Good idea! Created a separate PR to add that utility: https://github.com/microsoft/axe-windows/pull/872

DaveTryon (approved):

PR #872, opened 2023-01-26
feat(chrome-scans): Add filtered framework ids test utility

sfoslund:
#### Details

Adding test utility suggested in https://github.com/microsoft/axe-windows/pull/871#discussion_r1088151744. Initially this PR only leverages the new utility in the 3 tests that are not overlapping with those changed in https://github.com/microsoft/axe-windows/pull/871 to avoid PR conflicts, but I'll either update this PR or that one depending on which gets merged first. 

##### Motivation

Address PR feedback from https://github.com/microsoft/axe-windows/pull/871

#### Pull request checklist


src/RulesTest/Extensions.cs
```diff
@@ -23,5 +25,17 @@ public static IEnumerable<int> Difference(this IEnumerable<int> allItems, IEnume
                    where !itemsToRemove.Contains(item)
                    select item;
         }
+
+        public static string[] GetFilteredFrameworkIds(string[] exclude, string[] additions)
```
karanbirsingh, line 29: I like the idea of a utility method - personally, I think client code is easier to follow if this method just returns the raw list of known frameworks. Then there are no parameters and the caller is in charge of excluding or appending additional IDs. But both approaches are good with me!

sfoslund: Sounds reasonable to me, I agree that's more readable so I've switched to that approach :)

src/RulesTest/Library/ControlShouldSupportSetInfoWPFTests.cs
```diff
@@ -77,7 +79,7 @@ public void TestControlShouldSupportSetInfoWPFExpectedControlTypes()
         public void TestControlShouldSupportSetInfoWPFExpectedPlatform()
         {
             string[] expectedFrameworks = { FrameworkId.WPF };
-            string[] unexpectedFrameworks = { FrameworkId.DirectUI, FrameworkId.Edge, FrameworkId.InternetExplorer, FrameworkId.Win32, FrameworkId.WinForm, FrameworkId.XAML };
+            var unexpectedFrameworks = Extensions.GetFrameworkIds().Except(new string[] { FrameworkId.WPF });
```
DaveTryon, line 82: Most of the code uses explicit types instead of `var`. Any chance we can maintain that pattern?

DaveTryon (approved):
Approved with a suggestion to use explicit types instead of `var`

PR #873, opened 2023-01-27
feat(chrome-scans): Add mechanism for excluding elements from scans

sfoslund:
#### Details

This PR is 2/2 of addressing https://github.com/microsoft/axe-windows/issues/836. https://github.com/microsoft/axe-windows/pull/871 added the `ChromiumComponentsShouldUseWebScanner` rule and this PR adds logic to exclude elements that fail this rule from further scanning. It also includes logic to ensure that only the top level chrome element reports a `ChromiumComponentsShouldUseWebScanner` rule violation. 

##### Motivation

Addresses issue #836

#### Pull request checklist

- [x] Addresses an existing issue: #836

### DIFF
--- a/src/DesktopTests/UIAutomation/TreeWalkers/TreeWalkerForTestTests.cs
+++ b/src/DesktopTests/UIAutomation/TreeWalkers/TreeWalkerForTestTests.cs
@@ -44,7 +44,7 @@ namespace Axe.Windows.DesktopTests.UIAutomation.TreeWalkers
         public void RefreshTreeData_RunsAllRules()
         {
             _elementMock.Setup(e => e.Framework).Returns(FrameworkId.WPF);
-            var dataContext = new DesktopDataContext(Registrar.GetDefaultInstance(), A11yAutomation.GetDefaultInstance(), CancellationToken.None);
+            var dataContext = DesktopDataContext.DefaultContext;
             _treeWalker.RefreshTreeData(TreeViewMode.Raw, dataContext);
             var scanResults = _elementMock.Object.ScanResults;
             Assert.AreEqual(2, scanResults.Items.Count);
@@ -57,7 +57,7 @@ namespace Axe.Windows.DesktopTests.UIAutomation.TreeWalkers
         public void RefreshTreeData_ExcludesElementsWhenTheyViolateExclusionRules()
         {
             _elementMock.Setup(e => e.Framework).Returns(FrameworkId.Chrome);
-            var dataContext = new DesktopDataContext(Registrar.GetDefaultInstance(), A11yAutomation.GetDefaultInstance(), CancellationToken.None);
+            var dataContext = DesktopDataContext.DefaultContext;
             _treeWalker.RefreshTreeData(TreeViewMode.Raw, dataContext);
             var scanResults = _elementMock.Object.ScanResults;
             Assert.AreEqual(1, scanResults.Items.Count);
--- a/src/DesktopTests/UIAutomation/TreeWalkers/TreeWalkerForTestTests.cs
+++ b/src/DesktopTests/UIAutomation/TreeWalkers/TreeWalkerForTestTests.cs
@@ -10,6 +10,7 @@ using Axe.Windows.Desktop.UIAutomation.TreeWalkers;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Moq;
 using System;
+using System.Linq;
 using System.Threading;
 
 namespace Axe.Windows.DesktopTests.UIAutomation.TreeWalkers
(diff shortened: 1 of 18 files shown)


src/RuleSelection/RuleRunner.cs
```diff
@@ -54,15 +54,72 @@ private static void Run(ScanResults results, A11yElement e, CancellationToken ca
...
+            try
+            {
+                return ExcludeFromRunUnsafe(e, cancellationToken);
+            }
+            catch (OperationCanceledException)
+            {
+                throw;
+            }
+//#pragma warning disable CA1031 // Do not catch general exception types
+//            catch (Exception ex)
```
codeofdusk, line 75: Debugging code (did you intend to check this in)?

codeofdusk (commented):
Verified that this works in MS Teams using the CLI and the code changes look good to me but given the size of the change I'd like a second reviewer.

src/DesktopTests/UIAutomation/TreeWalkers/TreeWalkerForTestTests.cs
```diff
@@ -31,13 +32,38 @@ public void BeforeEachTest()
...
             cancellationToken.Cancel();
             var dataContext = new DesktopDataContext(Registrar.GetDefaultInstance(), A11yAutomation.GetDefaultInstance(), cancellationToken.Token);
             Assert.ThrowsException<OperationCanceledException>(() => _treeWalker.RefreshTreeData(TreeViewMode.Raw, dataContext));
         }
+
+        [TestMethod]
+        public void RefreshTreeData_RunsAllRules()
+        {
+            _elementMock.Setup(e => e.Framework).Returns(FrameworkId.WPF);
+            var dataContext = new DesktopDataContext(Registrar.GetDefaultInstance(), A11yAutomation.GetDefaultInstance(), CancellationToken.None);
```
DaveTryon, line 47: If there a reason not to use the `DefaultContext` ?
```suggestion
            var dataContext = DesktopDataContext.DefaultContext;
```

src/DesktopTests/UIAutomation/TreeWalkers/TreeWalkerForTestTests.cs
```diff
@@ -31,13 +32,38 @@ public void BeforeEachTest()
...
+            var exclusionRule = scanResults.Items.Where(i => i.Items.FirstOrDefault().Rule.Equals(RuleId.ChromiumComponentsShouldUseWebScanner));
+            Assert.AreEqual(1, exclusionRule.Count());
+            Assert.AreEqual(Core.Results.ScanStatus.Pass, exclusionRule.FirstOrDefault().Status);
+        }
+
+        [TestMethod]
+        public void RefreshTreeData_ExcludesElementsWhenTheyViolateExclusionRules()
+        {
+            _elementMock.Setup(e => e.Framework).Returns(FrameworkId.Chrome);
+            var dataContext = new DesktopDataContext(Registrar.GetDefaultInstance(), A11yAutomation.GetDefaultInstance(), CancellationToken.None);
```
DaveTryon, line 60: If there a reason not to use the `DefaultContext` ?
```suggestion
            var dataContext = DesktopDataContext.DefaultContext;
```

src/RuleSelection/RuleRunner.cs
```diff
@@ -54,15 +54,72 @@ private static void Run(ScanResults results, A11yElement e, CancellationToken ca
...
+        }
+
         private static ScanResult ConvertRunResultToScanResult(RunResult runResult)
         {
             if (runResult == null) throw new ArgumentNullException(nameof(runResult));
             if (runResult.RuleInfo == null) throw new ArgumentException(ErrorMessages.RunResultRuleInfoNull, nameof(runResult));
 
+            if (!runResult.IncludeInResults)
+            {
+                return null;
```
DaveTryon, line 120: Putting the check here adds a second responsibility to this method and forces upstream handling. What if we left this method unchanged and added a new method, something like this:

```
void ConvertAndAddNonExcludedRunResult(ScanResults results, RunResult runResult)
{
  if (runResult.IncludeInResults)
  {
    ScanResult scanResult = ConvertRunResultToScanResult(r);
    results.AddScanResult(scanResult);
  }
}
```

Then the code in the `Run` method simplifies to
```
ConvertAndAddNonExcludedRunResult(results, r);
```

src/Rules/Rule.cs
```diff
@@ -13,12 +13,16 @@ interface IRule
         RuleInfo Info { get; }
         Condition Condition { get; }
         bool PassesTest(IA11yElement element);
+        bool IncludeInResults(IA11yElement element);
+        // True if this rule failing should cause the element to be excluded from the rest of the scan.
+        bool Exclusionary { get; }
```
DaveTryon, line 18: How would you feel about naming this field `IsExclusionRule` to align with how it's consumed?

src/Rules/RuleRunner.cs
```diff
@@ -39,11 +40,25 @@ public RunResult RunRuleByID(RuleId id, IA11yElement element)
...
+            {
+                cancellationToken.ThrowIfCancellationRequested();
+                var result = RunRule(rule, element);
+                results.Add(result);
+            } // for all rules
+
+            return results;
+        }
+
         public IEnumerable<RunResult> RunAll(IA11yElement element, CancellationToken cancellationToken)
```
DaveTryon, line 57: The method name no longer accurately reflects what this method does

src/Rules/Library/ChromiumComponentsShouldUseWebScanner.cs
```diff
@@ -27,5 +29,11 @@ protected override Condition CreateCondition()
         {
             return Chrome;
         }
+
+        public override bool IncludeInResults(IA11yElement element)
```
DaveTryon, line 33: Taking a big step back, would it make sense to include this check in `ChronmiumComponentsShouldUseWebScanner.Condition` ? I think that would avoid the whole concept of included and excluded rules, leading to simpler code

DaveTryon (changes requested):
I'm fairly sure that we can do this more easily, just by updating the condition for the rule that we're trying to exclude. Please see [this comment](https://github.com/microsoft/axe-windows/pull/873/files#r1090854493) for more info

src/Rules/Library/ChromiumComponentsShouldUseWebScanner.cs
sfoslund: We could definitely do that, but I originally went with this approach because if we just updated the condition then we will continue to scan the chrome elements for all the other rules, which results in a lot of failures that aren't really reliable. One alternative is we could do the normal scan with the updated condition logic you're suggesting and then find some way to exclude all the failures from all the elements that fail the chrome rule after the fact, but that would mean running tests on many elements just to exclude them later, which felt wasteful to me.

sfoslund:
Per windows triage: Closing this for now as we're still undecided as to how best to approach this change. We plan to treat this as a full feature going forward.

PR #878, opened 2023-03-01
Fix: Clarify "how to fix" guidance for NameIsInformative rule violations

### LINKED ISSUE #874: [BUG] Should NameIsInformative rule filter out more frameworks?
## Describe the bug

The `NameIsInformative` rule exists for some very old framework bugs and is mostly noise for current frameworks. At the moment, the rule applies to _all_ non-Win32 frameworks. Should we modify the `Condition` to apply it only to the frameworks that are known to have this problem in their old versions?

codeofdusk:
#### Details
This PR adds additional guidance for when the `NameIsInformative` rule can be ignored safely.

##### Motivation
Related to #874.

#### Pull request checklist

- [ ] Addresses an existing issue: This PR should not close #874.

### DIFF
--- a/src/Rules/Resources/HowToFix.resx
+++ b/src/Rules/Resources/HowToFix.resx
@@ -587,8 +587,7 @@ If your application targets a version of .NET Framework before 4.7.1, but is run
     <comment>Brief guidance on how to fix an accessibility problem. The term "Name" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="NameNotEmpty" xml:space="preserve">
-    <value>Provide a UI Automation Name property that concisely identifies the element.
-</value>
+    <value>Provide a UI Automation Name property that concisely identifies the element.</value>
     <comment>Brief guidance on how to fix an accessibility problem. The term "Name" should not be translated (i.e. it should be treated as a loanword).</comment>
   </data>
   <data name="NameNotNull" xml:space="preserve">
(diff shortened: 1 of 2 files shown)


src/Rules/Resources/HowToFix.resx
```diff
@@ -581,9 +581,9 @@ If your application targets a version of .NET Framework before 4.7.1, but is run
     <comment>Brief guidance on how to fix an accessibility problem. The terms "name" and "LocalizedControlType" should not be translated (i.e. they should be treated as loanwords).</comment>
   </data>
   <data name="NameIsInformative" xml:space="preserve">
-    <value>Provide a UI Automation Name property for the element that:
- · Concisely identifies the element, AND
- · Does not include the element's class name (such as Microsoft.*.* or Windows.*.*).</value>
+    <value>Examine the UI Automation Name property of this element:
+ · If the Name property contains a class name (of the form Microsoft.*.* or Windows.*.*) that provides no meaningful information to end-users, update the Name to a concise, user-facing label.
+ · If the Name property contains "Microsoft" or "Windows" in user-facing content, such as a reference to microsoft.com in on-screen text, you may safely ignore this issue.</value>
```
DaveTryon, line 586: Suggested rewording:
```suggestion
 · If the Name property contains "Microsoft" or "Windows" and is closely related to the on-screen text (for example, a reference to "Microsoft.com"), you may safely ignore this issue.</value>
 · If the Name property is not closely related to the on-screen text (and provides no meaningful information to end-users), update the Name to a concise, user-facing label that is closely related to the on-screen text.
```

DaveTryon (approved):

PR #879, opened 2023-03-02
fix: Simplify NameIsInformative and whitelist common top-level domains

codeofdusk:
#### Details
This PR combines the `Microsoft.*.*` and `Windows.*.*` regexes into one expression, as they're fairly similar, and adds an additional check to pass elements whose names end in common TLDs (i.e. `microsoft.com` no longer throws an error).

##### Context
The check remains imperfect, as many of these violations don't start with `Microsoft` or `Windows` at all: for instance, this rule would not be able to detect microsoft/accessibility-insights-windows#1438.

Maybe we should change the regex to look for names with, say, at least three or four dots (and not specifically look for `Microsoft` or `Windows` at all)?

#### Pull request checklist


codeofdusk:
Closing in favour of a different approach.

PR #880, opened 2023-03-03
fix: Exclude Group and Text controls from NameIsInformative due to high false-positive rate

codeofdusk:
#### Details
This PR excludes the `Group` and `Text` UIA control types from the `NameIsInformative` rule in an attempt to reduce false positives.

##### Context
Re #874.

#### Pull request checklist

### DIFF
--- a/src/Rules/Library/NameIsInformative.cs
+++ b/src/Rules/Library/NameIsInformative.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using Axe.Windows.Core.Bases;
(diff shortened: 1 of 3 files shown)


src/Rules/Library/NameIsInformative.cs
```diff
@@ -45,6 +46,8 @@ protected override Condition CreateCondition()
         {
             return Name.NotNullOrEmpty & Name.NotWhiteSpace & BoundingRectangle.Valid
                 & ~Win32Framework
+                & ~PropertyConditions.ControlType.Group
```
codeofdusk, line 49: I was running into namespace collisions.

DaveTryon, line 49: Please also add a unit test for each of the new exclusion cases. While you're touching the tests, one of the existing tests (FrameworkWind32_DoesNotMatch) is misnamed (should be "Win32" instead of "Wind32"). Would you mind fixing the typo?

DaveTryon (approved):

PR #886, opened 2023-03-09
chore: Apply/suppress new analyzer suggestions

DaveTryon:
#### Details

The latest version of Visual Studio analyzers are flagging some things that they didn't flag before. This PR represents a single pass through the project to update the appropriate fields. The commits are non-overlapping and it's possibly simplest to review on a per-commit basis.

##### Motivation

Make analyzer suggestions more obvious in the IDE and less dangerous (i.e., don't use suggestions that require C# versions that we aren't guaranteed to have).

##### Context

According to the C# language versioning [docs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version), both .NET Standard 2.0 and .NET Framework default to C# 7.3, which is why we max out the suggestions there. The fields added to `.editorconfig` are based on the actual analyzer messages that occurred. I made no attempt to create an exhaustive list (potentially based on https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/overview?tabs=net-7), since it would quickly get out of date and it's easy to check new analyzer suggests as they occur.

The `.editorconfig` changes should probably also occur in AIWin, but that's far from urgent...

#### Pull request checklist

### DIFF
--- a/src/Actions/Actions/DataManager.cs
+++ b/src/Actions/Actions/DataManager.cs
@@ -65,7 +65,11 @@ namespace Axe.Windows.Actions
         /// <returns></returns>
         internal ElementContext GetElementContext(Guid ecId)
         {
-            return _elementContexts.ContainsKey(ecId) ? _elementContexts[ecId] : null;
+            if (_elementContexts.TryGetValue(ecId, out ElementContext ec))
+            {
+                return ec;
+            }
+            return null;
         }
 
         /// <summary>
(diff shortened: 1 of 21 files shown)


src/Actions/Actions/DataManager.cs
```diff
@@ -91,20 +94,17 @@ internal void RemoveElementContext(Guid ecId)
         internal void RemoveDataContext(Guid ecId, bool keepMainElement = true)
         {
             // check whether key exists. if not, just silently ignore.
-            if (_elementContexts.ContainsKey(ecId))
+            ElementContext ec = GetElementContext(ecId);
```
DaveTryon, line 97: Quick note for this class. The old code repeated this pattern several times:
```
if (_elementContexts.ContainsKey(ecId))
{
  var ec = _elementContexts[ecId];
  if (ec != null)
  {
    // Do stuff with ec
  }
}
```

Since there's already a `GetElementContext` method, the new pattern is as follows:

```
ElementContext ec = GetElementContext(ecId);
if (ec != null)
{
  // Do stuff with ec
}
```

It's a tiny bit shorter, although it hides the fact that `GetElementContext` is connected to `_elementContexts`. I'm happy to entertain other approaches if you have any suggestions

codeofdusk (approved):
LGTM

PR #889, opened 2023-03-09
chore: Prevent dependabot updates of pinned package versions

DaveTryon:
#### Details

The `CurrentFileVersionCompatibilityTests` project is intentionally pinned to old versions of `Axe.Windows` and `Newtonsoft.Json`. This is for testing only and not something that we ship. Dependabot knows to leave `Axe.Windows` alone, but we want updates to `Newtonsoft.Json` everywhere but in this project.

This PR creates a separate set of rules for the `CurrentFileVersionCompatibilityTests` project. This should work, based on the dependabot.yml [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file), but we'll ultimately find out empirically.

##### Motivation

Reduce overhead from unwanted dependabot changes

#### Pull request checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,9 +6,6 @@ updates:
     interval: daily
     time: "13:00"
   ignore:
-  - dependency-name: Axe.Windows
-    versions:
-    - "> 0.3.1-prerelease"
   - dependency-name: Microsoft.CodeAnalysis.BinSkim
     versions:
     - "> 1.7.0, < 1.8"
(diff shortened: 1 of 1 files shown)


codeofdusk (approved):
Seems reasonable to me...

.github/dependabot.yml
```diff
@@ -42,3 +39,16 @@ updates:
   - dependency-name: MSTest.TestAdapter
     versions:
     - 2.2.2
+
+- package-ecosystem: nuget
+  directory: "/src/CurrentFileVersionCompatibilityTests"
+  schedule:
+    interval: daily
+    time: "13:00"
```
chingucoding, line 47: I'm curious, do we need this in here given we also have this in the top level scope?

chingucoding (approved):

.github/dependabot.yml
DaveTryon: I honestly don't know. I couldn't find any dependabot.yml files that try to ignore within a subtree, so this whole thing is sort of experimental. Let's try it this way for now and see if it works at all.

PR #899, opened 2023-03-29
chore: Make integration test timeouts more tolerant

DaveTryon:
#### Details

Automation integration tests frequently have timeouts in our pipelines, which is bad. This PR replaces the VSTest-controlled timeout with a timeout that we control, so that we can return Inconclusive if a test times out in the pipeline. It also slightly increases the timeouts on a couple of tests and makes the timeout directly proportional to the number of processes being spun up.

This will probably be easier to review if you enable the "Hide whitespace" setting

##### Motivation

Flaky tests make life hard.

#### Pull request checklist

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -137,7 +137,6 @@ namespace Axe.Windows.AutomationTests
             });
         }
 
-
         [DataTestMethod]
         [DataRow(true)]
         [DataRow(false)]
@@ -242,6 +241,10 @@ namespace Axe.Windows.AutomationTests
 
             if (wrapper.CaughtException != null)
             {
+                if (wrapper.CaughtException.GetType() == typeof(AssertInconclusiveException))
+                {
+                    Assert.Inconclusive($"AssertInconclusiveException caught - See details below.\n{wrapper.CaughtException.Message}\n{wrapper.CaughtException.StackTrace}");
+                }
                 Assert.Fail($"Exception caught - See details below.\n{wrapper.CaughtException.Message}\n{wrapper.CaughtException.StackTrace}");
             }
 
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -77,139 +77,182 @@ namespace Axe.Windows.AutomationTests
         public void Cleanup()
         {
             StopTestApp();
-
             CleanupTestOutput();
         }
 
         [DataTestMethod]
         [DataRow(true)]
         [DataRow(false)]
-        [Timeout(30000)]
         public void Scan_Integration_WildlifeManager(bool sync)
         {
-            var processId = LaunchTestApp(_wildlifeManagerAppPath);
-            WindowScanOutput results = ScanIntegrationCore(sync, _wildlifeManagerAppPath, WildlifeManagerKnownErrorCount, processId: processId);
-            EnsureGeneratedFileIsReadableByOldVersionsOfAxeWindows(results, processId);
+            RunWithTimedExecutionWrapper(TimeSpan.FromSeconds(30), () =>
+            {
+                var processId = LaunchTestApp(_wildlifeManagerAppPath);
+                WindowScanOutput results = ScanIntegrationCore(sync, _wildlifeManagerAppPath, WildlifeManagerKnownErrorCount, processId: processId);
+                EnsureGeneratedFileIsReadableByOldVersionsOfAxeWindows(results, processId);
+            });
         }
 
         // [DataTestMethod]
         // [DataRow(true)]
         // [DataRow(false)]
-        [Timeout(30000)]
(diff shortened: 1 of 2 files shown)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -77,139 +77,182 @@ public AutomationIntegrationTests()
...
         [DataTestMethod]
         [DataRow(true)]
         [DataRow(false)]
-        [Timeout(30000)]
         public void Scan_Integration_WildlifeManager(bool sync)
         {
-            var processId = LaunchTestApp(_wildlifeManagerAppPath);
-            WindowScanOutput results = ScanIntegrationCore(sync, _wildlifeManagerAppPath, WildlifeManagerKnownErrorCount, processId: processId);
-            EnsureGeneratedFileIsReadableByOldVersionsOfAxeWindows(results, processId);
+            RunWithTimedExecutionWrapper(TimeSpan.FromSeconds(30), () =>
```
DaveTryon, line 88: In most cases, the `TimeSpan` and the old `Timeout` value are kept consistent

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -77,139 +77,182 @@ public AutomationIntegrationTests()
...
-            var cancelledCount = 2;
-            // The first 2 tasks will be cancelled in ~50ms and ~550ms
-            var cancelledTasks = tasks.Take(cancelledCount);
-            // The final 3 tasks will be cancelled after more than 5 seconds, so we expect them to finish before the cancellation is recognized
-            var finishedTasks = tasks.Skip(cancelledCount);
-
-            var timeout = 50;
-            for (int i = 0; i < cancelledCount; i++)
+            const int instanceCount = 5;
+            RunWithTimedExecutionWrapper(TimeSpan.FromSeconds(20 * instanceCount), () =>
```
DaveTryon, line 196: This intentionally increases the timeout from 60 seconds to 100 seconds

src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -77,139 +77,182 @@ public AutomationIntegrationTests()
...
-
-            foreach (var result in results)
+            const int instanceCount = 3;
+            RunWithTimedExecutionWrapper(TimeSpan.FromSeconds(20 * instanceCount), () =>
             {
-                ValidateOutput(result.WindowScanOutputs, WildlifeManagerKnownErrorCount);
-            }
+                var cancellationTokens = Enumerable.Range(0, instanceCount).Select(_ => new CancellationTokenSource().Token).ToList();
+                var tasks = GetAsyncScanTasks(_wildlifeManagerAppPath, cancellationTokens);
+                var results = Task.WhenAll(tasks).Result;
```
DaveTryon, line 183: Using `.Result` here (and in the next method) allowed me to keep `TimedExecutionWrapper` simple

src/AutomationTests/TimedExecutionWrapper.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private void WaitForThreadToStart()
+        {
+            while (_stopwatch.Elapsed < _allowedTime && !_thread.IsAlive)
+            {
+                Thread.Sleep(SleepTime);
+            }
+        }
+
+        private void WaitForThreadToComplete()
```
codeofdusk, line 67: It looks like `WaitForThreadToStart` and `WaitForThreadToComplete` are basically identical except for the state of the `IsAlive` boolean. Could you factor this out into a helper method, passing the expected liveness as a parameter?

src/AutomationTests/TimedExecutionWrapper.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    /// <summary>
+    /// A simple wrapper to bring our timeout behavior under our own control. Given a max allowed time
+    /// and an Action, it spins up a thread, runs the Action on that thread, and reports the outcome
+    /// via the Completed and CaughtException properties. The same object can be used to run multiple
+    /// Actions (or the same Action multiple times), as long as they all share the same timeout.
+    /// </summary>
+    internal class TimedExecutionWrapper
+    {
+        private static readonly TimeSpan SleepTime = TimeSpan.FromMilliseconds(250);
```
codeofdusk, line 18: Guessing this was chosen somewhat arbitrarily?

codeofdusk (commented):
Approach looks reasonable to me!

src/AutomationTests/TimedExecutionWrapper.cs
DaveTryon: Potentially yes, but it's a 3 line function, so it feels like breaking it out would reduce clarity.

src/AutomationTests/TimedExecutionWrapper.cs
DaveTryon: Yes. It's not special

codeofdusk (approved):

PR #915, opened 2023-05-10
fix: Skip English-specific rules on non-English systems

DaveTryon:
#### Details

https://github.com/microsoft/accessibility-insights-windows/issues/1572 called out the fact that the `LocalizedControlTypeReasonable` rule uses English language strings to validate localized names. The chosen solution was to change the `Condition` so that this rule only applies on English language systems. This PR adds a new `Condition` called `IsEnglish` that takes advantage of .NET's built-in support of 3 letter language codes as defined in the ISO 639-2 spec. For our purposes, we disable the rule for anything other than "eng". The check is case-insensitive, which probably isn't strictly necessary, but seemed like the safest bet.

Tested by setting my Windows language to French and verifying that `SystemPropertiesTests.OverriddenCultureName_ResetsToDefault` reports as inconclusive. It also revealed the need for adjustments to `MonsterTest` when run on non-English systems.

##### Motivation

Address https://github.com/microsoft/accessibility-insights-windows/issues/1572, avoid noise for Visual Customers running on non-English systems.

##### Context

My first iteration used the `CultureInfo.Name` property, which follows a format like **en-us**. This worked reasonably well, but using the three-letter language standard made the code a little it simpler.
 
#### Pull request checklist

- [x] Addresses an existing issue: https://github.com/microsoft/accessibility-insights-windows/issues/1572

### DIFF
--- a/src/Rules/PropertyConditions/SystemProperties.cs
+++ b/src/Rules/PropertyConditions/SystemProperties.cs
@@ -1,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Axe.Windows.Rules.PropertyConditions
+{
+    static class SystemProperties
+    {
+        private static string SystemCultureName = System.Globalization.CultureInfo.CurrentCulture.Name;
+
+        internal static string OverriddenCultureName { get; set; }
+
+        public static Condition IsEnglish = CreateEnglishConditionWithTestOverride();
+
+        private static Condition CreateEnglishConditionWithTestOverride()
+        {
+            StringProperty cultureName = new StringProperty(_ => (OverriddenCultureName ?? SystemCultureName).ToLowerInvariant());
+
+            return new OrCondition(cultureName.IsNoCase("en"), cultureName.MatchesRegEx("^en-.*"));
+        }
+    } // class
+} // namespace
--- a/src/Rules/PropertyConditions/SystemProperties.cs
+++ b/src/Rules/PropertyConditions/SystemProperties.cs
@@ -5,17 +5,17 @@ namespace Axe.Windows.Rules.PropertyConditions
 {
     static class SystemProperties
     {
-        private static string SystemCultureName = System.Globalization.CultureInfo.CurrentCulture.Name;
+        private static readonly string SystemISOLanguageName = System.Globalization.CultureInfo.CurrentCulture.ThreeLetterISOLanguageName;
 
-        internal static string OverriddenCultureName { get; set; }
+        internal static string OverriddenISOLanguageName { get; set; }
 
         public static Condition IsEnglish = CreateEnglishConditionWithTestOverride();
 
         private static Condition CreateEnglishConditionWithTestOverride()
         {
-            StringProperty cultureName = new StringProperty(_ => (OverriddenCultureName ?? SystemCultureName).ToLowerInvariant());
+            StringProperty cultureName = new StringProperty(_ => (OverriddenISOLanguageName ?? SystemISOLanguageName));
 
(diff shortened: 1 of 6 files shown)


src/Rules/PropertyConditions/SystemProperties.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public static Condition IsEnglish = CreateEnglishConditionWithTestOverride();
+
+        private static Condition CreateEnglishConditionWithTestOverride()
+        {
+            StringProperty cultureName = new StringProperty(_ => (OverriddenISOLanguageName ?? SystemISOLanguageName));
+
+            return cultureName.IsNoCase("eng");
+        }
+    } // class
```
codeofdusk, line 20: What is the purpose of these comments? Should they also be placed at the end of functions (or are classes/namespaces a special case)? Why is indentation insufficient to determine what the braces close?

src/RulesTest/MonsterTest.cs
```diff
@@ -236,6 +239,17 @@ public void MonsterUserControl()
             CheckForExtraRules(expectedResults);
         }
 
+        private static void RemoveEnglishSpecificRulesIfNeeded(Dictionary<RuleId, EvaluationCode> expectedResults)
+        {
+            if (!string.Equals(CultureInfo.CurrentCulture.ThreeLetterISOLanguageName, "eng", System.StringComparison.OrdinalIgnoreCase))
+            {
+                if (expectedResults.ContainsKey(RuleId.LocalizedControlTypeReasonable))
```
codeofdusk, line 246: Is there a common place where we can keep a list of English-specific rules/a function to remove them, usable in all of our system tests? Does this matter?

src/RulesTest/PropertyConditions/SystemProperitesTests.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
codeofdusk, line 1: Rename file to `SystemPropertiesTests`

src/RulesTest/PropertyConditions/SystemProperitesTests.cs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Axe.Windows.Rules.PropertyConditions;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Globalization;
+
+namespace Axe.Windows.RulesTests.PropertyConditions
+{
+    [TestClass]
+    public class SystemProperitesTests
```
codeofdusk, line 11: And the class as well

DaveTryon: Done

src/RulesTest/PropertyConditions/SystemProperitesTests.cs
DaveTryon: Done

src/RulesTest/MonsterTest.cs
DaveTryon: So far this is the only place that we use them. I've refactored the single value into an `IEnumerable` to streamline the code a little bit. If a future change requires sharing of the `IEnumerable`, it'll be a trivial change to point to the new location.

src/Rules/PropertyConditions/SystemProperties.cs
DaveTryon: This is @RobGallo's pattern. His text editor isn't particularly good about indents, so he uses this technique to help clarify where classes end. It's probably more valuable in cases where we have multiple classes in a source file, which is pretty rare in our code. I started the class by copying and pasting another class, so this came along for the ride.

I find 349 instances of `// namespace` and 331 instances of `// class`. If we want to clean those up, it feels like a holistic change to consider

codeofdusk (approved):

PR #919, opened 2023-05-23
Chore: improve usage of {Locked} keyword and minor string cleanup

codeofdusk:
#### Details
This PR:
* Removes some strings from `Locked` comments where they aren't used in the original message.
* Sets more appropriate scoping on some `Locked` usage.
* Adds some `Locked` keywords that were missed previously.
* General cleanup of strings (opportunistic typo fixing, etc.)

#### Pull request checklist
- [n/a] Addresses an existing issue: #0000

PR #927, opened 2023-05-25
chore: Further changes to translator comments

codeofdusk:
#### Details
This PR:
* Adds the C# keywords `true`, `false`, and `null` to `Locked` comments where used.
* Adds guidance to translators to alphabetize the list of standard landmark types, as done in English.

##### Motivation
Follow-up of #919 and esn updates in #926.

### DIFF
--- a/src/Rules/Resources/Descriptions.resx
+++ b/src/Rules/Resources/Descriptions.resx
@@ -507,7 +507,7 @@
   </data>
   <data name="FrameworkDoesNotSupportUIAutomation" xml:space="preserve">
     <value>The framework used to build this application does not support UI Automation.</value>
-    <comment>The description of an accessibility problem with an application currently under inspection.</comment>
+    <comment>The description of an accessibility problem with an application currently under inspection. {Locked="UI Automation"}</comment>
   </data>
   <data name="EdgeBrowserHasBeenDeprecated" xml:space="preserve">
     <value>The non-Chromium version of Microsoft Edge has been deprecated.</value>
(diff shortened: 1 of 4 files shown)


codeofdusk:
FYI @csigs.

src/Rules/Resources/Descriptions.resx
```diff
@@ -147,51 +147,51 @@
   </data>
   <data name="IsContentElementFalseOptional" xml:space="preserve">
     <value>The recommended value of the IsContentElement property for the given control type is false. Please consider if this is an element that should be reported to an assistive technology user as content.</value>
-    <comment>The description of an accessibility problem with an element currently under inspection. {Locked="IsContentElement"}</comment>
+    <comment>The description of an accessibility problem with an element currently under inspection. {Locked="false", "IsContentElement"}</comment>
```
DaveTryon, line 150: @csigs, what's the policy on localizing boolean values? I know that `false.ToString()` in French returns `faux` instead of `false`. What's the guidance for boolean values in docs?

src/Rules/Resources/HowToFix.resx
```diff
@@ -150,51 +150,51 @@
   </data>
   <data name="IsContentElementFalseOptional" xml:space="preserve">
     <value>Set the IsContentElement UI Automation property to false.</value>
```
DaveTryon, line 152: @codeofdusk, should "UI Automation" be locked in these strings? It's currently localized

DaveTryon (approved):

PR #931, opened 2023-06-01
fix: Reduce noise in ChromiumComponentsShouldUseWebScanner rule

DaveTryon:
#### Details

https://github.com/microsoft/accessibility-insights-windows/issues/1625 points out that the `ChromiumComponentsShouldUseWebScanner` rule is flagging items that are outside the context of a web page, which goes against the rule's primary purpose. This change limits the `ChromiumComponentsShouldUseWebScanner` rule to elements within a HTML document (including the document itself).

While working on this, we discovered a bug in WebView2, which is tracked at https://github.com/MicrosoftEdge/WebView2Feedback/issues/3530.

##### Motivation

Address https://github.com/microsoft/accessibility-insights-windows/issues/1625

###### Text results using Edge

As expected, only the `ChromiumComponentsShouldUseWebScanner` rule is impacted
Rule | Count Before | Count After
--- | --- | ---
An element of the given ControlType must not support the invoke pattern | 1 | 1
An on-screen element must not have a null BoundingRectangle property | 7 | 7
Chromium components should be scanned with a web-based scanner | 211 | 135
The Name must not include the same text as the LocalizedControlType | 5 | 5
The Name property must not contain only whitespace | 3 | 3
The Name property must not include the element's control type | 1 | 1
The Name property of a focusable element must not be null | 10 | 10

###### Screenshots using Edge

These screenshots are limited to just the `ChromiumComponentsShouldUseWebScanner` rule:

State | Screenshot
--- | ---
Before change | image
After change | image

##### Context

This includes an opportunistic optimization to some of the code in `Relationships`, because the unit test code looked odd with arbitrary multipliers to compensate for hoe `Relationships` is coded. This could be a separate PR if that's preferred.

#### Pull request checklist

- [x] Addresses an existing issue: https://github.com/microsoft/accessibility-insights-windows/issues/1625

### DIFF
--- a/src/Rules/PropertyConditions/Relationships.cs
+++ b/src/Rules/PropertyConditions/Relationships.cs
@@ -43,7 +43,7 @@ namespace Axe.Windows.Rules.PropertyConditions
             if (siblings == null) return false;
             if (index < 1 || index > siblings.Count()) return false;
 
-            return e.Parent.Children.ElementAt(index - 1).RuntimeId == e.RuntimeId;
+            return siblings.ElementAt(index - 1).RuntimeId == e.RuntimeId;
         }
 
         private static bool HasSiblingsOfSameType(IA11yElement e)
--- a/src/Rules/PropertyConditions/Relationships.cs
+++ b/src/Rules/PropertyConditions/Relationships.cs
@@ -6,6 +6,7 @@ using Axe.Windows.Core.Exceptions;
 using Axe.Windows.Core.Misc;
 using Axe.Windows.Rules.Resources;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace Axe.Windows.Rules.PropertyConditions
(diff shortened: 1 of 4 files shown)


src/Rules/PropertyConditions/Relationships.cs
```diff
@@ -38,17 +39,19 @@ private static Condition CreateSecondChildCondition()
         private static bool MatchOrderInSiblings(IA11yElement e, int index)
         {
             if (e == null) throw new ArgumentNullException(nameof(e));
-            if (e.Parent == null || e.Parent.Children == null) return false;
```
DaveTryon, line 41: Old code evaluated e.Parent up to 4 times, and e.Parent.Children up to 3 times

src/Rules/PropertyConditions/Relationships.cs
```diff
@@ -38,17 +39,19 @@ private static Condition CreateSecondChildCondition()
...
+            return siblings.ElementAt(index - 1).RuntimeId == e.RuntimeId;
         }
 
         private static bool HasSiblingsOfSameType(IA11yElement e)
         {
-            if (e?.Parent == null) return false;
+            IA11yElement parent = e?.Parent;
+            if (parent == null) return false;
 
-            var count = (from c in e.Parent.Children
```
DaveTryon, line 51: Old code evaluated e.Parent twice

src/Rules/PropertyConditions/Relationships.cs
```diff
@@ -66,9 +69,10 @@ private static bool MatchParent(Condition c, IA11yElement e)
         {
             if (c == null) throw new ArgumentNullException(nameof(c));
             if (e == null) throw new ArgumentNullException(nameof(e));
-            if (e.Parent == null) return false;
```
DaveTryon, line 69: Old code evaluated e.Parent up to twice. New code maxes out at once.

codeofdusk (approved):

PR #951, opened 2023-07-20
fix: Reduce chromium noise

DaveTryon:
#### Details

As called out in #836, the `ChromiumContentShouldUseWebScanner` has caused a lot of confusion. This PR aims to reduce that confusion. It does the following:

- The `ChromiumContentShouldUseWebScanner` rule will now trigger _only_ on Chromium-generated `Document` elements
- Other rules are disabled on elements that are Chromium-generated `Document` elements or any of their descendants
- Future rules default to being disabled on elements that are Chromium-generated `Documents` or any of their descendants. The default behavior can easily be overridden
- Where needed for unit tests with strict `mock`s, provide test constructors that allow unit tests to isolate the class under test from the Chromium suppression

##### Motivation

Address #836, make the `ChromiumContentShouldUseWebScanner` rule less intrusive

##### Attachments
[BeforeAndAfter.zip](https://github.com/microsoft/axe-windows/files/12101240/BeforeAndAfter.zip) contains 2 `a11ytest` files
- `Before.a11ytest` contains scan results from Edge before this change
- `After.a11ytest` contains scan results from Edge after this change

Here are how the numbers line up

Message | Count before | Count after | Count of Chromium elements
--- | --- | --- | ---
An element must not have the same Name and LocalizedControlType as its parent | 7 | 0 | 7
An element of the given ControlType must not support the Invoke pattern | 1 | 1 | 0
An element of the given ControlType must support the Text pattern | 1 | 0 | 1
An on-screen element must not have a null BoundingRectangle property | 5 | 5 | 0
Chromium components should be scanned with a web-based scanner | 63 | 2 | 63
The Name must not include the same text as the LocalizedControlType | 2 | 2 | 0
The Name property must not include the element's control type | 2 | 2 | 0

Things to note on this table:
- There are 2 Chromium documents on the page--one for the HTML content and one for the search button. This is an implementation detail of the search button
- Only the 2 Chromium documents are counted in "Count after". The 61 omitted elements are children of one of the two Chromium documents
- For all other columns. "Count before" - "Count after" = "Count of Chromium elements". This is as expected

If you open `After.a11ytest` and inspect the children of the Chromium documents, the UIA properties are reported but no tests were run. That's exactly what this change was intended to do.

##### Screenshots

Here are visual representations of the test results with all test results displaying. Notice the difference in the HTML document space:

File | Snapshot (click to enlarge) | Test Results (click to enlarge)
--- | --- | ---
Before | image | image
After | image | image

#### Pull request checklist

- [x] Addresses an existing issue: #836


src/Rules/Library/ChromiumComponentsShouldUseWebScanner.cs
```diff
@@ -27,7 +26,7 @@ public override bool PassesTest(IA11yElement e)
 
         protected override Condition CreateCondition()
         {
-            return Chrome & (Document | AnyAncestor(Document));
+            return Chrome & Document;
```
DaveTryon, line 29: This is the change that reduces the scope of `ChromiumComponentsShouldUseWebScanner` from _all_ Chromium elements to just Chromium _Documents_

src/Rules/Rule.cs
```diff
@@ -19,17 +20,28 @@ abstract class Rule : IRule
...
-            // keep these two calls in the following order or the RuleInfo.Condition string won't get populated
+            // keep these calls in the following order or the RuleInfo.Condition string won't get populated
 #pragma warning disable CA2214
             Condition = CreateCondition();
 #pragma warning restore CA2214
+            Condition = CreateCondition();
+
+            if (excludedCondition != null)
+            {
+                Condition = Condition - excludedCondition;
```
DaveTryon, line 35: This is the change that excludes the passed-in condition

src/Rules/Rule.cs
```diff
@@ -19,17 +20,28 @@ abstract class Rule : IRule
     {
         public RuleInfo Info { get; private set; }
         public Condition Condition { get; }
+        protected static Condition DefaultExcludedCondition => IsChromiumContent;
```
DaveTryon, line 23: Exposing this to derived classes gives us a single place to modify the default, then the production constructors of the derived classes just pick up the new value without requiring any additional code change

src/Rules/Rule.cs
```diff
@@ -19,17 +20,28 @@ abstract class Rule : IRule
...
+
+            if (excludedCondition != null)
+            {
+                Condition = Condition - excludedCondition;
+            }
 
             InitRuleInfo();
         }
 
+        protected Rule() : this (excludedCondition: DefaultExcludedCondition)
```
DaveTryon, line 41: By default, exclude `DefaultExcludedCondition`

src/RulesTest/Conditions/IsChromiumContentConditionTests.cs
```diff
@@ -0,0 +1,153 @@
...
+            _elementMock.Setup(m => m.ControlTypeId)
+                .Returns(ControlType.Document)
+                .Verifiable();
+
+            Assert.IsTrue(Condition.Matches(_elementMock.Object));
+
+            _elementMock.VerifyAll();
+        }
+
+        public void Matches_FrameworkIsChrome_IsNotDocument_ParentIsChromeDocument_ReturnsTrue()
```
DaveTryon, line 81: This test overlaps with the one after it, so could potentially be removed

src/Rules/Library/ChromiumComponentsShouldUseWebScanner.cs
```diff
@@ -6,14 +6,13 @@
...
 using static Axe.Windows.Rules.PropertyConditions.ControlType;
 using static Axe.Windows.Rules.PropertyConditions.Framework;
-using static Axe.Windows.Rules.PropertyConditions.Relationships;
 
 namespace Axe.Windows.Rules.Library
 {
     [RuleInfo(ID = RuleId.ChromiumComponentsShouldUseWebScanner)]
     class ChromiumComponentsShouldUseWebScanner : Rule
     {
-        public ChromiumComponentsShouldUseWebScanner()
```
DaveTryon, line 16: I could build an argument for renaming this to `ChromiumDocumentsShouldUseWebScanner`, but that would make the PR harder to review. Happy to put this into a subsequent PR if needed.

src/Rules/Library/ChromiumComponentsShouldUseWebScanner.cs
```diff
@@ -27,7 +25,7 @@ public override bool PassesTest(IA11yElement e)
 
         protected override Condition CreateCondition()
         {
-            return Chrome & (Document | AnyAncestor(Document));
+            return IsChromiumDocument;
```
DaveTryon, line 28: This is the change that limits `ChromiumComponentsShouldUseWebScanner` to just scan Chromium Documents

codeofdusk (approved):
Verified that:

* In a WebView2 application, the "chromium content should use  web scanner" rule was reported to fail. No spurious Chromium-related errors were reported.
* In an Electron application, the "chromium content should use  web scanner" rule was reported to fail. No spurious Chromium-related errors were reported.
* In a non-web scenario, no change is observed.

src/Rules/Rule.cs
```diff
@@ -19,17 +20,28 @@ abstract class Rule : IRule
...
 
-        protected Rule()
+        protected Rule(Condition excludedCondition)
         {
-            // keep these two calls in the following order or the RuleInfo.Condition string won't get populated
+            // keep these calls in the following order or the RuleInfo.Condition string won't get populated
 #pragma warning disable CA2214
             Condition = CreateCondition();
 #pragma warning restore CA2214
+            Condition = CreateCondition();
```
RobGallo, line 31: Why is this called a second time?

RobGallo (changes requested):

src/Rules/Conditions/IsChromiumContentCondition.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using static Axe.Windows.Rules.PropertyConditions.ElementGroups;
+using static Axe.Windows.Rules.PropertyConditions.Relationships;
+
+namespace Axe.Windows.Rules
+{
+    /// <summary>
+    /// Determines is the element is an element within the context of a Chromium document
+    /// </summary>
+    class IsChromiumContentCondition : Condition
```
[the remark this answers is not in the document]
DaveTryon: Updated

src/Rules/Rule.cs
DaveTryon: Typo -- fixed

RobGallo (approved):

PR #967, opened 2023-08-30
feat: Enable scanning of chromium content

DaveTryon:
#### Details

In #951, we intentionally excluded all elements inside a Chromium document from scan results. This makes sense for most developers, since the conversion from HTML to UIA is convoluted and outside the control of the HTML developer. However, the Edge team relies on this functionality to help improve their HTML-to-UIA conversion code. This PR defaults to excluding the elements inside a Chromium document, but adds a way to request that scan results include elements within a Chromium document. This option is exposed in 3 places:
1. In the `Automation` layer
2. In the CLI
3. In the `SelectAction` class (for future use from Accessibility Insights for Windows)

##### Motivation

Requested by the Edge team

##### Results
[Comparison.zip](https://github.com/microsoft/axe-windows/files/12480720/Comparison.zip) contains `Default.a11ytest` and `WithChromiumContent.a11ytest`, which were captured by running the CLI on the same instance of Edge. The rule comparison is as follows:

Message | Default Count | WithChromiumContent count
--- | --- | ---
An on-screen element must not have a null BoundingRectangle property | 5 | 5
Chromium components should be scanned with an web-based scanner | 2 | 2
The Name must not include the same text as the LocalizedControlType | 2 | 2
The Name property must not include the element's control type | 2 | 2
An element must not have the same Name and LocalizedControlType as its parent | 0 | 4
An element of the given ControlType must support the Text pattern | 0 | 1

The elements that are unique to the WithChromiumContent file are all within the Chromium document

##### Screenshots 

Case | Snapshot (click to enlarge) | Results (click to enlarge)
--- | --- | ---
Default | image |image
WithChromiumContent | image | image
Delta | image | image

##### Context

- The `Actions` project didn't previously reference the `Rules` project directly, so the plumbing may be non-ideal. If we were to use the existing dependencies, `Actions` would call into `Desktop`, which would call into `RuleSelection`, which would call into `Rules`. We can add the 2 missing layers if we think there's value there.
- The flag to enable Chromium content is basically global. This will work as expected in the CLI, in AI-Win, and in AccChecker. It will fail if someone creates a scenario where concurrent scans want to have different Chromium behavior. I've taken a YAGNI approach for now, and we can always make it more complex at a future time, should that prove necessary.

#### Pull request checklist


docs/AutomationReference.md
```diff
@@ -100,6 +100,14 @@ Cause each scan to save a test file, even if no errors are reported. The default
 
 The `WithAlwaysSaveTestFile` method returns the  `Config.Builder` configured to always save a test file.
 
+##### `WithTestAllChromiumContent`
+
+Cause all Chromium HTML content to be included in scan results. The default is to omit Chromium HTML content from the scan results. This is intended only for use by browser development teams and should be omitted in other scenarios.
+
+###### Return object
+
+The `WithTestAllChromiumContent` method returns the  `Config.Builder` configured to always save a test file.
```
codeofdusk, line 109: Did you mean for this to mention changed Chromium behaviour?

src/Automation/Data/Config.cs
```diff
@@ -42,6 +42,9 @@ public class Config
         /// <summary>Override the default behavior of only saving a11ytest files if errors are found.</summary>
         public bool AlwaysSaveTestFile { get; private set; }
 
+        /// <summary>Override the default behavior or not testing Chromium content.</summary>
```
codeofdusk, line 45:
```suggestion
        /// <summary>Override the default behavior of not testing Chromium content.</summary>
```

src/Automation/Interfaces/IAxeWindowsActions.cs
```diff
@@ -49,5 +49,12 @@ internal interface IAxeWindowsActions
         /// </summary>
         /// <param name="path"></param>
         void RegisterCustomUIAPropertiesFromConfig(string path);
+
+        /// <summary>
+        /// Specify how Chromium content should be evaluated
+        /// </summary>
+        /// <param name="shouldTestAllChromiumContent">This can be set to true allow browser teams to test and debug code that converts from HTML to UIA.
```
codeofdusk, line 56:
```suggestion
        /// <param name="shouldTestAllChromiumContent">This can be set to true to allow browser teams to test and debug code that converts from HTML to UIA.
```

codeofdusk (approved):

docs/AutomationReference.md
DaveTryon: Fixed

src/Automation/Data/Config.cs
DaveTryon: Fixed

src/Automation/Interfaces/IAxeWindowsActions.cs
DaveTryon: Fixed

PR #994, opened 2024-01-25
feat(Pipeline Migration to 1ES Template): migrated signed build to 1ES micro build template

v-viyada:
#### Details

This PR is for migrating Microsoft.axe-windows(signed) pipeline to 1ES MicroBuild Template.

Old run link: https://dev.azure.com/mseng/1ES/_build/results?buildId=25829758&view=results
New run link: https://dev.azure.com/mseng/1ES/_build/results?buildId=25904428&view=results

Below are details:
1. Test cases are less in the new pipeline as we are filtering UI test cases in debug and release compliance job. These test cases are already filtered in Signed Release job as the agent pool used for micro build do not support it. After migration debug and release compliance jobs will also run on the same agent pool instead of default ADO agent pool. These test cases are running successfully in PR and CI Pipelines. Filter used is --filter TestCategory!=Integration.
2. We have to keep running compliance jobs as we are running TSA upload task. Currently 1ES template do not support dynamic TSA config file.
3. Binskim task is running twice in Signed Release job. One is auto injected by 1ES template and another one we added manually for TSA Upload. 1ES template do not support TSA upload for specific job in the pipeline. If we enable TSA in the template than it will inject TSA task in all the jobs including source SDL job. We cannot enable TSA for sourceSDL job because it require to hardcore tsa config in the repo.
4. Compared the output artifacts between old and new run. Below are the differences. 
   - _manifest folder in all the artifacts. This contains SBOM Manifest for the artifact. For Official template (used for production pipeline), this task is enabled by default. This folder is also available in Nuget artifact which is used in  "Deploy Axe-Windows to NuGet" release pipeline. I think it will not impact the release pipeline as per pattern specified to get the nuget files.
   - DebugDrop has extra files as there are more SDL task running due to auto injection from the template.
   - MicroBuild folder missing from drop artifacts. This is because MicroBuild plugins are now installed using auto injected task which is define to download the plugins in Temporary directory instead of source directory. I tried copying them into artifacts but credscan task for artifact folders fails as this contains some secret. Also we are not using this into our release pipeline so it is not a required output. Pipeline run [link](https://dev.azure.com/mseng/1ES/_build/results?buildId=25454449&view=results) for credscan task failure.
5. ##[warning]1ES PT Warning: Failed to fetch default repository branch.
This is an existing issue. Please refer
[Bug 2119901](https://dev.azure.com/mseng/1ES/_workitems/edit/2119901): Address issue "Failed to fetch default branch"

##### Motivation

[User Story 2136518](https://dev.azure.com/mseng/1ES/_workitems/edit/2136518): Migrate  Microsoft.axe-windows(signed) to 1ES template

#### Pull request checklist

- [na] Addresses an existing issue: #0000

### DIFF
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -32,6 +32,7 @@ extends:
       name: MSEngSS-MicroBuild2022-1ES
     sdl:
       binskim:
+        #If you modify this list, you also need to modify the list in the Run BinSkim task in Signed Release Job or vice-versa to keep them in sync
         analyzeTargetGlob: "$(System.DefaultWorkingDirectory)\\src\\CI\\bin\\**\\*.dll;\
                             $(System.DefaultWorkingDirectory)\\src\\Actions\\bin\\**\\*.dll;\
                             $(System.DefaultWorkingDirectory)\\src\\Automation\\bin\\**\\*.dll;\
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -42,10 +42,10 @@ extends:
                             $(System.DefaultWorkingDirectory)\\src\\SystemAbstractions\\bin\\**\\*.dll;\
                             $(System.DefaultWorkingDirectory)\\src\\Telemetry\\bin\\**\\*.dll;\
                             $(System.DefaultWorkingDirectory)\\src\\Win32\\bin\\**\\*.dll;"
-
-                
+         
     stages:
     - stage: Stage
+
       jobs:
       - job: ComplianceRelease
         templateContext:
@@ -59,30 +59,36 @@ extends:
           displayName: 'Use NuGet 5.x'
           inputs:
             versionSpec: '5.x'
+
         - task: NuGetCommand@2
           displayName: 'NuGet restore'
+
         - task: DotNetCoreCLI@2
           displayName: 'dotnet restore'
           inputs:
             command: restore
             projects: |
               **\*.csproj
+
         - task: PowerShell@2
           displayName: 'License Header Check'
           inputs:
             targetType: "filePath"
             filePath: tools\scripts\verification.scripts\LicenseHeaderVerification.ps1
             arguments: '-target  $(Build.Repository.LocalPath) -licenseHeaderPath tools\scripts\verification.scripts\LicenseHeader.txt -extensions *.xaml,*.xml,*.cs,*.ps1 -addIfAbsent $false'
+
         - task: VSBuild@1
           displayName: 'Build Solution **\*.sln'
           inputs:
             vsVersion: 17.0
             platform: '$(BuildPlatform)'
             configuration: release
+
         - task: ms.build-release-task.custom-build-release-task.wpf-static-analysis@0
           displayName: 'WPF Static Analysis'
           inputs:
             input: 'src\Axe.Windows\bin\Release'
+
         - task: DotNetCoreCLI@2
           displayName: 'Test Assemblies (.NET Core) **\*test*.csproj'
           inputs:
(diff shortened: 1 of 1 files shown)


build/signedbuild.yml
```diff
@@ -17,347 +17,351 @@ variables:
...
+                            $(System.DefaultWorkingDirectory)\\src\\RuleSelection\\bin\\**\\*.dll;\
+                            $(System.DefaultWorkingDirectory)\\src\\SystemAbstractions\\bin\\**\\*.dll;\
+                            $(System.DefaultWorkingDirectory)\\src\\Telemetry\\bin\\**\\*.dll;\
+                            $(System.DefaultWorkingDirectory)\\src\\Win32\\bin\\**\\*.dll;"
+
+                
+    stages:
+    - stage: Stage
+      jobs:
+      - job: ComplianceRelease
```
DaveTryon, line 50: General comment: The old file used newlines to try to break it up into chunks. the new file has removed the newlines, which makes it a lot harder to identify the various blocks. Can we please restore newlines between pieces?

build/signedbuild.yml
```diff
@@ -17,347 +17,376 @@ variables:
...
+extends:
+  template: azure-pipelines/MicroBuild.1ES.Official.yml@MicroBuildTemplate
+  parameters:
+    pool: 
+      name: MSEngSS-MicroBuild2022-1ES
+    sdl:
+      binskim:
+        analyzeTargetGlob: "$(System.DefaultWorkingDirectory)\\src\\CI\\bin\\**\\*.dll;\
+                            $(System.DefaultWorkingDirectory)\\src\\Actions\\bin\\**\\*.dll;\
+                            $(System.DefaultWorkingDirectory)\\src\\Automation\\bin\\**\\*.dll;\
```
DaveTryon, line 37: I understand why we're duplicating the glob list, but it opens the door to the two lists getting out of sync. I'd love to see this combined into a single statement, but a reasonable alternative would be to add a comment before each `analyzeTargetGlob` block to be sure that all changes gets reflected in both places. Maybe something like "If you modify this list, you also need to modify the list in the Run BinSkim task"

build/signedbuild.yml
```diff
@@ -17,347 +17,376 @@ variables:
...
+    ref: refs/tags/release
+
+extends:
+  template: azure-pipelines/MicroBuild.1ES.Official.yml@MicroBuildTemplate
+  parameters:
+    pool: 
+      name: MSEngSS-MicroBuild2022-1ES
+    sdl:
+      binskim:
+        analyzeTargetGlob: "$(System.DefaultWorkingDirectory)\\src\\CI\\bin\\**\\*.dll;\
```
DaveTryon, line 35: Looks like 1ESPT uses camelCase for this, while BinSkim uses PascalCase. Not our problem to fix, I guess :(

DaveTryon (approved):
@v-viyada, I made one suggestion about adding a couple of comments to help keep the BinSkim stuff in sync. No need to re-review it after making that change.

build/signedbuild.yml
v-viyada: Added the comment

build/signedbuild.yml
v-viyada: reverted the new lines

build/signedbuild.yml
v-viyada: yes, 1ES template uses camelCase for all sdl settings.

DaveTryon:
@v-viyada, the presence of the `_manifest` folder in the `msi` artifact will break the Canary release script because it only expects a single folder to exist under the `msi` folder. To fix this, edit https://github.com/microsoft/accessibility-insights-windows/blob/main/tools/scripts/pipeline/deploy/Canary/set_variables.ps1#L18 from

```
$a11yInsightsVersion = Get-ChildItem -Name -Directory $($msiFolder)
```

to

```
$a11yInsightsVersion = Get-ChildItem -Name -Directory $($msiFolder) -Exclude '_manifest'
```

PR #996, opened 2024-01-29
fix: Add license headers to .yml and .targets files

DaveTryon:
#### Details

While updating to 1ES pipeline templates, we realized that several of our files were missing copyright headers. This adds them and updates the tooling to ensure that we don't miss them in these files types in the future. Copyright headers were all added by running the script.

The policy service file can't have comments without breaking the bot, so an exception is made for any files under `.github\policies`.

##### Motivation

Address #991

#### Pull request checklist

- [x] Addresses an existing issue: #991

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,3 +1,5 @@
+﻿# Copyright (c) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE file in the project root for full license information.
 version: 2
 updates:
 - package-ecosystem: nuget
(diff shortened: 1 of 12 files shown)


.github/dependabot.yml
```diff
@@ -1,3 +1,5 @@
+﻿# Copyright (c) Microsoft. All rights reserved.
```
DaveTryon, line 1: There's a potential that this will break the dependabot agent, but we don't have a way to check it without merging it. If it breaks, we'll back out the change and expand the exceptions.

DaveTryon: It didn't break dependabot in accessibility-insights-windows, so we should be ok here

peterdur (approved):
LGTM!

PR #1020, opened 2024-04-13
feat(logical-size): Implement the LogicalSize property in Axe.Windows

codeofdusk:
#### Details
This PR adds a "pseudo-property", `LogicalSize`, a normalized size based on an element's `BoundingRectangle` to aid in touch size calculations. 

##### Motivation
Part of [Feature 2152841](https://mseng.visualstudio.com/1ES/_workitems/edit/2152841) (internal access required to view).

#### Pull request checklist

### DIFF
--- a/src/Core/Misc/ExtensionMethods.cs
+++ b/src/Core/Misc/ExtensionMethods.cs
@@ -599,6 +599,17 @@ namespace Axe.Windows.Core.Misc
             return retVal;
         }
 
+        /// <summary>
+        /// Convert Element Property value to Size
+        /// </summary>
+        /// <param name="property"></param>
+        /// <returns>A Size object</returns>
+        public static Size ToSize(this A11yProperty property)
+        {
+            if (property?.Value is null) return Size.Empty;
+            return new Size(property.Value[0], property.Value[1]);
+        }
+
         public static string ToLeftTopRightBottomString(this Rectangle r)
         {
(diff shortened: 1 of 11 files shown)


src/Core/Misc/ExtensionMethods.cs
```diff
@@ -599,6 +599,17 @@ public static T[] ToArray<T>(this Newtonsoft.Json.Linq.JArray j)
...
         }
 
+        /// <summary>
+        /// Convert Element Property value to Size
+        /// </summary>
+        /// <param name="property"></param>
+        /// <returns>A Size object</returns>
+        public static Size ToSize(this A11yProperty property)
+        {
+            if (property.Value is null) return Size.Empty;
```
DaveTryon, line 609: This throws a `NullReferenceException` if `property` is null:
```suggestion
            if (property?.Value is null) return Size.Empty;
```

brocktaylor7 (approved):
LGTM, but it might make sense to add some tests for this new logic.  One obvious example would be the ToStringTest for the A11yProperty.

codeofdusk:
I've added a unit test for the `ToString` logic. At some point, it might be worth:

* Adding a unit test for the `AddLogicalSizePseudoProperty` method in `DesktopElementExtensionMethods`, but that's a private method and test coverage in that file is virtually nonexistent.
* A system test of some kind to verify the new logic in a running app, but that's complicated buy different DPI scaling settings on different machines.

PR #1038, opened 2024-07-23
fix: Exclude the win32 IP address control from the SiblingUniqueAndFocusable rule

codeofdusk:
#### Details
The Win32 IP address control consists of a pane of edit fields, one per octet of an IPV4 address. As such, the `SiblingUniqueAndFocusable` rule is not appropriate for this control.

#### Pull request checklist
- [x] Addresses an existing issue: microsoft/accessibility-insights-windows#1838

### DIFF
--- a/src/Rules/Library/SiblingUniqueAndFocusable.cs
+++ b/src/Rules/Library/SiblingUniqueAndFocusable.cs
@@ -63,7 +63,7 @@ namespace Axe.Windows.Rules.Library
                 & WPF
                 & NoChild(Custom | Name.NullOrEmpty);
 
-            var ipAddressControl = ClassName.Is("SysIPAddress32");
+            var ipAddressControl = Win32Framework & ClassName.Is("SysIPAddress32");
 
             return EligibleChild
                 & NotParent(wpfDataItem)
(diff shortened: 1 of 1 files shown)


src/Rules/Library/SiblingUniqueAndFocusable.cs
```diff
@@ -63,7 +63,12 @@ protected override Condition CreateCondition()
                 & WPF
                 & NoChild(Custom | Name.NullOrEmpty);
 
-            return EligibleChild & NotParent(wpfDataItem);
+            var ipAddressControl = ClassName.Is("SysIPAddress32");
```
DaveTryon, line 66: You may or may not want to also restrict this to the Win32 control:
```suggestion
            var ipAddressControl = Win32 & ClassName.Is("SysIPAddress32");
```

codeofdusk: I tried this, but got an error to the effect of "`Win32` is a namespace but used like a variable".

DaveTryon: Use `Win32Framework` instead of `Win32` (refer to comment in Framework.cs)

JGibson2019 (approved):

Issue #1046, opened 2024-08-26
[BUG] False positive with Wizards (property sheet)


v-chojas:
## Describe the bug

Wizards are implemented as property sheets with one tab per page and that appears to cause a false positive.

**To Reproduce**

1. Open Device Manager
2. Select Computer and choose Action -> Add legacy hardware
3. Click Next
4. Hover over this window with Accessibility Insights and click Test icon

## Expected behavior

No false positives.

## Actual behavior

"Focusable sibling elements must not have the same Name and LocalizedControlType" error on each page of the wizard, highlighting the nonexistent "tabs" which are by no means focusable.

## Screenshots or .GIF

![addhw](https://github.com/user-attachments/assets/4f623c5e-a841-46de-98f9-032d957726ec)

## Desktop (please complete the following information):

 - OS: Windows Server 2022 (update version 10.20348.2655), Windows 11 24H2 (update version 10.26100.1457)
 - Accessibility Insights for Windows Version: 1.1.2703.1
 - Target Application: Add Hardware wizard
 - Target Application Version: see Windows version

## Additional context

I have a minimal repro app I can provide if necessary, but since I could also reproduce this on Windows' own Add Hardware wizard, that one can be used for an easy test.

codeofdusk:
Resolved in latest release (this rule has been downgraded to `NeedsReview`) in Win32 applications.

PR #1047, opened 2024-09-03
fix: Downgrade SiblingUniqueAndFocusable to NeedsReview for Win32 apps

codeofdusk:
#### Details
Due to a large number of false positives (and given the fact that Win32 applications don't support UIA natively), downgrade `SiblingUniqueAndFocusable` to `NeedsReview` in Win32 applications.

##### Motivation
Addresses #1046. Supersedes #1038. Addresses microsoft/accessibility-insights-windows#1838.

#### Pull request checklist
- [x] Addresses an existing issue: #1046, microsoft/accessibility-insights-windows#1838.


madalynrose (approved):
One formatting question but looks pretty straightforward!

Issue #1074, opened 2024-11-27
Switch all `[DataTestMethod]` to `[TestMethod]`


Youssef1313:
See https://github.com/microsoft/testfx/issues/4166

codeofdusk:
[Sounds reasonable](https://stackoverflow.com/questions/46904953/datatestmethod-vs-testmethod). Are you willing to submit a PR?

marcelwgn:
I think this issue has been addressed right?

Youssef1313:
Yes. And I'm avoiding now the commented out occurrence in https://github.com/microsoft/axe-windows/pull/1093.

codeofdusk:
We've typically kept issues open until the associated PR is included in a release. If and when we release next, this issue will be closed out.

PR #1075, opened 2024-11-27
Switch DataTestMethod to TestMethod

Youssef1313:
Addresses #1074

### DIFF
--- a/src/AutomationTests/AutomationIntegrationTests.cs
+++ b/src/AutomationTests/AutomationIntegrationTests.cs
@@ -88,7 +88,7 @@ namespace Axe.Windows.AutomationTests
             ScanIntegrationCore(sync: true, testAppPath: null, expectedErrorCount: 0, processId: BogusProcessId);
         }
 
-        [DataTestMethod]
+        [TestMethod]
         [DataRow(true)]
         [DataRow(false)]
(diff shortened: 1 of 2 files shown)


src/AutomationTests/AutomationIntegrationTests.cs
```diff
@@ -132,7 +132,7 @@ public void Scan_Integration_WildlifeManager_InvalidRoot(bool sync)
             });
         }
 
-        // [DataTestMethod]
+        // [TestMethod]
```
codeofdusk, line 135:
```suggestion
        // [DataTestMethod]
```

Youssef1313: @codeofdusk Why? Ideally when the test is enabled back, you will not want it to use DataTestMethod still.

Youssef1313: Note that once MSTest analyzers are enabled in the repo, you should actually start getting a warning here because it's a public method missing the attribute and you'll need to uncomment it soon and use `[Ignore]`

v-sharmachir:
As azure pipelines are not running for this PR, raised similar PR for this https://github.com/microsoft/axe-windows/pull/1083, hence closing this one

PR #1077, opened 2024-11-29
chore: update binskim to optout 4.3.1 version

v-viyada:
#### Details

This is new pre release version used in 1ES pipeline templates. But our TSA upload task still uses the older version (1.9.5) so this conflict is resulting into pipeline failure.

For now updating pipeline to optout for new version. 

##### Context

There will be separate work item to update binskim to 4.3.1 version as new rules in 4.3.1 might discover issues which also need to be resolved so that pipeline does not break.

#### Pull request checklist

### DIFF
--- a/build/signedbuild.yml
+++ b/build/signedbuild.yml
@@ -34,6 +34,7 @@ extends:
       name: MSEngSS-MicroBuild2022-1ES
     sdl:
       binskim:
+        preReleaseVersion: ''
         # If you modify this list, you also need to modify the list in the Run BinSkim task in Signed Release Job or vice-versa to keep them in sync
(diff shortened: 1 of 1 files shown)


build/signedbuild.yml
```diff
@@ -34,6 +34,7 @@ extends:
       name: MSEngSS-MicroBuild2022-1ES
     sdl:
       binskim:
+        preReleaseVersion: ''
```
codeofdusk, line 37:
```suggestion
        # #1077: Hold BinSkim back at 4.3.1 to resolve a pipeline incompatibility
        preReleaseVersion: ''
```

codeofdusk (approved):

PR #1091, opened 2025-01-31
Update README.md - fix Typos in sample code

mrlacey:
#### Details

Fixes typos in sample code so don't get warnings if copy and paste into a real project (as intended from the sample).

##### Motivation

Save someone else having to correct a few warnings.

##### Context

Only corrected this file and haven't checked others. Just addressing something I noticed while updating a project from an older version of the package that used different code to scan.

#### Pull request checklist


codeofdusk (approved):
Ah, the joys of text-to-speech and not carefully checking my capitalization. Thanks @mrlacey!

